PR #43, opened 2015-01-15
add buildout infrastructure, add uml2.py

thet:
- add buildout infrastructure to build/develop gaphor
- add uml2.py from the latest egg release. ok, saw this can be generated from uml2.gaphor (awesome!). but why not including it? makes it easier to bootstrap without the need of building that file.


danyeaw:
@thet Thanks for the contribution! Could you add some docs for how this new buildout infrastructure would be utilized?

requirements.txt
```diff
@@ -0,0 +1,3 @@
+setuptools
+zc.buildout
+etk.docking
```
amolenaar: The `etk.docking` dependency can be dropped.

It would be nice to have a build step that does an actual buildout build, so we can validate it remains working.

buildout.cfg
```diff
@@ -0,0 +1,24 @@
...
+
+eggs-directory = ${buildout:directory}/eggs
+sources-dir = ${buildout:directory}/devsrc
+parts += gaphor
+develop = .
+auto-checkout +=
+    gaphas
+
+[sources]
+gaphas = git git@github.com:amolenaar/gaphas.git rev=f11b3e31081c1557d86e858dec2a54464d140914
```
amolenaar: Change to `git@github.com:gaphor/gaphas.git`?

PR #51, opened 2018-09-22
Make syntax Python 3 compatible

danyeaw:
Builds on https://github.com/gaphor/gaphor/pull/50. Used futurize to make the code Python 3 compatible.

- [x] Fix errors with using future builtins
- [x] Fix errors with using future standard libraries
- [x] Fix errors with using Py3 object
- [x] Fix unicode errors `unicode object has no attribute 'value'` when building UML
- [x] Test with Python 3.7.0


danyeaw:
Gaphor is now running in Python 2.7 with most syntax forward ported to Python 3, and the syntax should be fully compatible with Python 3.7.0. :+1:  I conducted testing with Python 2 and 3. The next step to getting a Python 3 version of Gaphor will be to port PyGTK to PyGObject.

amolenaar:
Wow! That's one big change. I'll try to find some time to do some testing again.

PR #56, opened 2018-10-04
Create ISSUE_TEMPLATE.md

tonytheleg:
created a minimal issue template for bug reports, feature requests, and document change requests.


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,48 @@
...
+<pre><code>
+As a (user, developer, contributor, etc):
+I want:
+So that:
+<!-- Describe the motivation or the concrete use case. -->
+</code></pre>
+
+## Environment
+<pre><code>
+Browser:
```
danyeaw: Since this is a Gtk program, I don't think that the web browser information is needed.

danyeaw:
@tonytheleg Thanks, this looks really great. I just have one update that I would like to see made. :+1:

tonytheleg:
Ok removed the browser information (makes sense if its GTK haha). Changed the title and left OS information, I am assuming that would still be important to know. Left Windows in case someone is able to assist with your Windows support effort.

Issue #94, opened 2019-03-12
Make element editor more prominent


Alexander-Wilms:
### I'm submitting a...
- [ ] Bug report  
- [x] Feature request
- [ ] Documentation issue or request

### Current behavior

Initially, I thought something was broken, since I didn't see any way of editing elements like classes etc.

### Expected behavior
When right-clicking on an UML element, it should offer to open the element editor, or it should reside in a (semi-)permanent side pane.

### Minimal reproduction of the problem with instructions
Create new diagram, add class element, try to edit it.

### If this is a feature request please fill out the following:
<pre><code>
As a (user, developer, contributor, etc): user
I want: A more obvious way of editing UML elements
So that: I can start working faster

</code></pre>

### System Information
OS:
- [x] Linux (Please put in notes the specific distro)
- [ ] MacOS
- [ ] Windows 10
- [ ] Windows 7

NOTES:
Fedora 29 KDE

amolenaar:
In the GTK+2 version of Gaphor one could double-click on a grey box and edit it in place.

It would make sense to restore this behaviour, I think. Having the editor in a side pane would take up quite a bit of screen real estate.

danyeaw:
Should we fix this before 1.0.0 release?

amolenaar:
I need to allocate some time to dig into this. It should be worthwhile fixing.

amolenaar:
Fixed by #108. You can now double click on an element and a small edit popup appears.

Issue #96, opened 2019-03-12
Re-opening editor window makes Gaphor crash


amolenaar:
### I'm submitting a...
- [X] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

### Current behavior

When the editor window is opened for a second time, after it's closed, it will crash the application.

### Minimal reproduction of the problem with instructions

* Start Gaphor,
* Open the editor (Command-e)
* close the editor window
* open it again
* crash (SIGSEGV)

### System Information
OS:
- [ ] Linux (Please put in notes the specific distro)
- [X] MacOS
- [ ] Windows 10
- [ ] Windows 7

NOTES:

The Console window does not make Gaphor crash, instead the window will not show when opened for a second time.

danyeaw:
Closed by #109. :tada:

PR #108, opened 2019-03-23
Fix in-diagram editor to something workable

amolenaar:
It should still submit text when clicked outside the editor window.

### PR Type

- [X] Bugfix
- [ ] Feature

Text can only be edited via the Editor window

Issue Number: #94 

### What is the new behavior?

Double click on a diagram item will open a editor box.

### DIFF
--- a/gaphor/ui/diagramtools.py
+++ b/gaphor/ui/diagramtools.py
@@ -148,7 +148,6 @@ class TextEditTool(Tool):
         window = Gtk.Window.new(Gtk.WindowType.TOPLEVEL)
         window.set_property("decorated", False)
         window.set_property("skip-taskbar-hint", True)
-        window.set_resize_mode(Gtk.RESIZE_IMMEDIATE)
         window.set_modal(True)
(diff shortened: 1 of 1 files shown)


danyeaw:
@amolenaar Thanks for working on this! I am getting a seg fault when I load the editor in Linux. I will see if I can narrow it down to one of the changes you made. Do you know how to run Gaphor through gdb to get GTK debug info? I have never had success every time I try this.

danyeaw:
@amolenaar Removing the resize mode did fix the initial seg fault. I did find another one, but it is there with the master branch as well, so this is definitely an improvement. :+1:

I can submit a separate issue about that one.

amolenaar:
@danyeaw I did some old fashioned comment-out-the-line debugging in order to reveal the issue. I'm happy to track down the other segfault as well, if you can point me in the right direction.

danyeaw:
@amolenaar Nice :smile:. It looks like you already found the issue as well with #96.

PR #113, opened 2019-03-28
Diagram specific toolbox items

### LINKED ISSUE #70: ToolPalette Buttons Don't Click
### I'm submitting a...
- [X] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

### Current behavior

Although drag and drop currently works from the ToolPalette, clicking on a button to drag and drop it, does not result in the button actually being clicked. The cause of this is because each button has `Gtk.ToolButton.set_use_drag_window(True)`. This causes all events (even button clicks) to be intercepted as a drag event.

The docs say that the easiest way to use drag and drop with Gtk.ToolPalette is to call `Gtk.ToolPalette.add_drag_dest()` with the desired drag source palette and the desired drag target widget. This is kind of the opposite of the technique we are using, since we need to setup the toolbox and then add a callback to the drag source after the DiagramPage is created.

### Expected behavior
You should be able to click and drag and see the button pressed. You should also be able to click on the button and then click on the diagram page. This is used for connecting lines between two items.

amolenaar:
### PR Type

- [X] Bugfix
- [ ] Feature

Fixed the toolbox. You can select the items and use DND like a boss.

Fixes  #70


gaphor/ui/toolbox.py
```diff
@@ -31,60 +31,140 @@ class Toolbox(Gtk.ToolPalette):
...
+            self.properties.get("reset-tool-after-create", True)
+        )
         self.buttons = []
         self.shortcuts = {}
-        self._construct(toolboxdef)
-
-    def toolbox_button(self, action_name, stock_id):
-        button = Gtk.ToolButton.new_from_stock(stock_id)
-        button.action_name = action_name
-        button.set_use_drag_window(True)
```
danyeaw: For my understanding, how did you get this to work without using `set_use_drag_window`?

danyeaw:
Just :open_mouth:. Things are working so nice now, this is awesome! I would like to spend some time tomorrow reviewing this in more detail (mostly for my understanding).

gaphor/ui/toolbox.py
```diff
@@ -31,60 +31,140 @@ class Toolbox(Gtk.ToolPalette):
...
+    def _on_toolbox_destroyed(self, widget):
+        self._toolbox = None
+
+    @toggle_action(name="reset-tool-after-create", label=_("_Reset tool"), active=False)
+    def reset_tool_after_create(self, active):
+        self.properties.set("reset-tool-after-create", active)
+
+    @component.adapter(IDiagramPageChange)
+    def _on_diagram_page_change(self, event):
+        print("Diagram page changed", event.diagram_page)
```
danyeaw: Do we still need this print statement?

gaphor/ui/toolbox.py
```diff
@@ -31,60 +31,140 @@ class Toolbox(Gtk.ToolPalette):
...
+    def construct(self):
+        def toolbox_button(action_name, stock_id, label, shortcut):
+            button = Gtk.ToggleToolButton.new()
+            button.set_icon_name(stock_id)
+            button.action_name = action_name
+            if label:
+                button.set_tooltip_text("%s (%s)" % (label, shortcut))
+
+            # Enable Drag and Drop
+            inner_button = button.get_children()[0]
```
danyeaw: This looks like the magic that made this work: why / how does the ToggleToolButton have a child button?

gaphor/ui/toolbox.py
amolenaar: The ToggleToolButton is using an internal button to do the heavy lifting. I put the DND functionality of that button, instead of the tool button. I assume the tool button has no (gdk) window so it's not able to receive any event.

See https://github.com/gaphor/gaphor/blob/diagram-specific-menu-items/gaphor/ui/toolbox.py#L86

gaphor/ui/toolbox.py
amolenaar: It took me some digging into the Gtk code 😜

amolenaar:
@danyeaw I made some improvements based on your comments.

PR #121, opened 2019-03-31
Create a Working macOS App

danyeaw:
Fixes the Gaphor.app built so that it doesn't crash on startup.

- Allow gaphor to be run as a Python Module
- Add services as plugins to the pyproject.toml so they get installed correctly by pip when building the package

### PR Checklist

### PR Type

- [X] Bugfix
- [ ] Feature

Gaphor.app immediately closes

Issue Number: N/A

### What is the new behavior?
Gaphor.app launches successfully in macOS

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/__main__.py
+++ b/gaphor/__main__.py
@@ -1,0 +1,9 @@
+from gaphor import main
+
+
+def run_main():
+    main()
+
+
+if __name__ == "__main__":
+    run_main()
--- a/gaphor/__main__.py
+++ b/gaphor/__main__.py
@@ -1,9 +1,5 @@
 from gaphor import main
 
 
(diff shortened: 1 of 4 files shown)


gaphor/__main__.py
```diff
@@ -0,0 +1,9 @@
+from gaphor import main
+
+
+def run_main():
+    main()
+
+
+if __name__ == "__main__":
+    run_main()
```
amolenaar: No need for the run_main function. We can just call main directly.

danyeaw: @amolenaar Good point, I'll clean that up! 👍

PR #133, opened 2019-04-10
Remove UIManager service

amolenaar:
It was a weird one, inheriting from Gtk.UIManager and working in close
collaboration with the ActionManager. Now everything is in ActionManager.

Note: this change will require a `poetry install`.

### PR Type

- [ ] Bugfix
- [ ] Feature


danyeaw:
This is some great refactoring to simplify things. :+1:  Any thoughts on if we should eventually move to `Gtk.Builder` in the future since Gtk.UIManager is deprecated?

amolenaar:
I don’t know if we should switch directly. Think not. But now the services talk on a higher level than plain gtk calls, which I think is good.

PR #139, opened 2019-04-13
Self contained uml package

amolenaar:
Reduce external dependencies for the UML module.

### PR Type

- [ ] Bugfix
- [ ] Feature

The `gaphor.UML` module is dependent on global state, namely the adapter registry of a global "site-manager" inside the zope.component module.

This is bad. First of all since global state in general is a bad idea, but it also means that our own `component_registry` has to do some monkey-patching to make sure the application works as expected.

Issue Number: N/A

### What is the new behavior?

The eventing mechanism for derived properties is baked in to the properties classes. The event handling itself is externalized, and delegated to the ElementFactory, who knows about components and stuff.

- [ ] Yes
- [X] No

Unless your plugin is using `component.handle`, that is.

### Other information

For now, elements should keep their reference to the element factory. It's needed in case properties are parsed and some delegated properties performa query on the factory.

### DIFF
--- a/gaphor/UML/collection.py
+++ b/gaphor/UML/collection.py
@@ -237,10 +237,7 @@ class collection(object):
             i2 = self.items.index(item2)
             self.items[i1], self.items[i2] = self.items[i2], self.items[i1]
 
-            # send a notification that this list has changed
-            factory = self.object.factory
-            if factory:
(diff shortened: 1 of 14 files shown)


danyeaw:
These changes look go to me!

PR #152, opened 2019-05-02
Convert Zope interfaces to Abstract base classes and generic functions

amolenaar:
NB. This PR builds on #151.

In order to simplify Gaphor, and get rid of some complex dependencies, I want to get rid of zope.component's `getAdapter` and `queryMultiAdapter`.

Multi-adapters are replaced by multi-dispatch functions (classes) and the eventing system by a similar one, but dependent only on a small library that I made part of Gaphor.

For this to work I assimilated the `generic` project from Andrey Popp: https://github.com/andreypopp/generic. This lib was referenced in PEP 433, Single-dispatch generic functions. Although it's written for Python 2, it does pretty much all we need. It may need an update when it comes to using the `typing` library, but that's no requirement for now.

### PR Type

- [ ] Bugfix
- [ ] Feature

Heavy dependency on zope.interface, zope.component, etc.

Issue Number: N/A

### What is the new behavior?

We should be able to get rid of zope as a dependency and solely rely on our own code.

- [X] Yes - on the API level
- [ ] No

Plugins or dependent projects should stop using the zope stuff.


LICENSE.txt
```diff
@@ -1,207 +0,0 @@
-Apache License
```
danyeaw: Did you mean to delete the LICENSE.txt?

PR #154, opened 2019-05-08
Create Event manager service for event handling

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Restructuring

The ComponentRegistry service both maintained a registry of services and application wide objects, as well as the event management logic.

### What is the new behavior?

The event management code has been split out into a separate service: `EventManager`.

In addition, the API for event subscription for both `EventManager` and `ElementDispatcher` has been changed from using `(un)register_handler` to `(un)subscribe`. I think this is clear and concise.

- [X] Yes
- [ ] No

### DIFF
--- a/gaphor/transaction.py
+++ b/gaphor/transaction.py
@@ -127,7 +127,7 @@ class Transaction:
         try:
             event_manager = self.event_manager
         except (application.NotInitializedError, application.ComponentLookupError):
-            log.warning("Could not lookup component_registry. Not emitting events.")
+            log.warning("Could not lookup event_manager. Not emitting events.")
         else:
             event_manager.handle(event)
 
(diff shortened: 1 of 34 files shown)


gaphor/transaction.py
```diff
@@ -125,11 +125,11 @@ def _close(self):
 
     def _handle(self, event):
         try:
-            component_registry = self.component_registry
+            event_manager = self.event_manager
         except (application.NotInitializedError, application.ComponentLookupError):
             log.warning("Could not lookup component_registry. Not emitting events.")
```
danyeaw: Should this log also be renamed to event_manager?

amolenaar: Ah, nice catch 😄

PR #155, opened 2019-05-09
Add a bit of styling to the toolbox window

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

### What is the new behavior?

The tool item groups were big. Now they're much smaller and look nicer IMHO.

I also added a scrolled window so that if all tool item boxes are expanded, one can still access them all.

### DIFF
--- a/docs/generic/event_system.rst
+++ b/docs/generic/event_system.rst
@@ -3,7 +3,7 @@ Event system
 
 Generic library provides ``generic.event`` module which helps you implement
 event systems in your application. By event system I mean an API for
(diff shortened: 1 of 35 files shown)


danyeaw:
On my computer, this actually looks very similar to what it looked like before:
Screenshot from 2019-05-10 22-33-52

The toolbox sections are no longer raised, but it doesn't show up as more compact I don't think.

amolenaar:
The margins are slightly smaller.  I think it looks nicer when the section headers are flat.

danyeaw:
OK with me 👍

PR #157, opened 2019-05-11
Modules per diagram

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, ~no api changes~)

### What is the new behavior?

Moved code that belongs to the same diagram in the same module. This means that diagram items, in-diagram editors and the property editor (property pages) for all elements that belong to a single diagram type now lives together in one package.

The "adapters" package was basically shadowing the structure of the "diagram" package, where "diagram" contained only diagram items and "adapters" contains the connection logic and editor related stuff.

This new structure should make it easier to find elements and bits of code.

The `gaphor.adapters` module is gone.

- [X] Yes
- [ ] No

### Notes:

This PR is a continuation of #154.


gaphor/diagram/items.py
```diff
@@ -4,70 +4,18 @@
...
-from gaphor.diagram.extension import ExtensionItem
-from gaphor.diagram.profiles.metaclass import MetaclassItem
+from .diagramitem import DiagramItem
+from .diagramline import DiagramLine, NamedLine
+from .elementitem import ElementItem
+from .nameditem import NamedItem
+from .compartment import CompartmentItem, FeatureItem
+from .classifier import ClassifierItem
+
+from .actions import *
```
danyeaw: I know these imports were very verbose, but PEP8 discourages use of wildcard imports.

amolenaar: The specific import is already done in the diagram-specific package. Therefore it seemed to me a bit redundant to also do the specific imports here.

danyeaw: @amolenaar Ok, that's a good point. :+1:

gaphor/diagram/__init__.py
```diff
@@ -6,10 +6,16 @@
 
 import uuid
 
-from gaphor.diagram.style import Style
-
-# Map UML elements to their (default) representation.
-_uml_to_item_map = {}
+from . import (
```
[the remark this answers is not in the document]
amolenaar: I think you're right. Let me find some time for an update

danyeaw:
:tada: :+1:

PR #163, opened 2019-05-20
Do not set up DND for toolbox-pointer

amolenaar:
This is done explicitly. It does the trick, although it may not be the most clean way.

Fixes 158.

### PR Type

- [X] Bugfix
- [ ] Feature

### What is the new behavior?

Selection tool is no longer draggable.

- [ ] Yes
- [X] No


danyeaw:
That works :+1:

PR #173, opened 2019-05-25
Model maintenance

amolenaar:
Some maintenance and clean up on the model side of things.

### PR Type

- [ ] Bugfix
- [ ] Feature

### What is the new behavior?

Element factory is passed to diagram items as well as model elements. This is more consistent now.

I tried getting rid of `ElementFactory` in model elements altogether, but that's not possible since some UML (model) queries require a scan on the model elements. Also, at least one diagram item requires a factory to create a new model element. I would have liked to have no reference to a model factory from the model itself, but that seems not possible, so I'll leave that as is.

Diagram items have a (somewhat implicit) dependency on `ElementDispatcher`. `ElementDispatcher` is injected, hence an "application" should have been initialized. This is a pitty, since we can not load a model without setting up an application. However, this is food for a future change.

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/UML/diagram.py
+++ b/gaphor/UML/diagram.py
@@ -90,12 +90,15 @@ class Diagram(Namespace, PackageableElement):
         parameter is the element class to create.  The new element also has an
         optional parent and subject."""
 
+        return self.create_as(type, str(uuid.uuid1()), parent, subject)
+
+    def create_as(self, type, id, parent=None, subject=None):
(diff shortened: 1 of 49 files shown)


amolenaar:
I removed the commit "UML.parse requires an explicit element factory now": element needs to have a reference to a factory anyway, so it would be safest to use that reference to create new elements, just like we do in diagram items.

PR #179, opened 2019-06-01
Constructor DI

amolenaar:
Do Dependency injection through the constructor instead of via the `inject` function.

This way we do not depend on a single `Application` instance to govern all services.

### PR Type

- [ ] Bugfix
- [X] Feature


danyeaw:
This looks great!

PR #183, opened 2019-06-07
Boyscouting

amolenaar:
Some cleanup, all over the place.

### PR Type

- [ ] Bugfix
- [ ] Feature


gaphor/application.py
```diff
@@ -57,7 +57,7 @@ def load_services(self, services=None):
 
         Service should provide an interface gaphor.abc.Service.
         """
-        for ep in pkg_resources.iter_entry_points("gaphor.services"):
+        for ep in importlib_metadata.entry_points()["gaphor.services"]:
```
danyeaw: Nice, I didn't know about that package :+1:

danyeaw:
:1st_place_medal:

gaphor/application.py
amolenaar: I ran into it by accident. It looks like it’s used a lot already. It was part of our transient dependencies.

amolenaar:
I need to change one more thing: delete the inject() function.

PR #184, opened 2019-06-10
More boyscouting

amolenaar:
Some more cleanup and low hanging fruit.

* Use `ast.literal_eval()` instead of plain `eval()`
* Remove the `inject()` function altogether
* Fix small issue where rubberband box was not drawn
* Add an example model with all element


danyeaw:
Yay no more injecting dependencies :tada:

PR #188, opened 2019-06-23
New diagram item structure

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Working with diagram items is hard. There are deep inheritance structures and it's not obvious what happens where.

Examples include the `EditableTextSupport` and `StereotypeSupport` classes. The way `DiagramItem` is entangled with how Gaphas elements are working, etc.

### What is the new behavior?

The `gaphor.UML.Presentation` has been  customized, just like `gaphor.UML.Diagram`. I'm building a library of graphical components (shapes) that can be used to construct a diagram item. This should lead to more composition, less inheritance and a more descriptive way of describing what a diagram item looks like.

NB. I changed the style structure to use fields more in line with CSS, e.g. "min-width" and "min-height" instead of "min-size".

There are still things to figure out:

- [x] How to deal with events? Should event handling be part of the shape?
- [x] If so, how to register/unregister events?
- [x] How to deal with line-like items, such as `Association`?
- [x] How to do "editable text"?
- [x] Build a set of components to use for classes, interfaces and other classifier-like elements.

- [X] Yes
- [ ] No

NB. This PR should not contain the conversion of all diagram items. Instead it should lay the foundation on which other diagram items can be migrated to a simpler structure.

NB. Have a look at `gaphor/diagram/action/action.py`. This file contains the first items I migrated.


gaphor/diagram/tests/test_text.py
```diff
@@ -0,0 +1,52 @@
...
+
+TEXT_SIZE = (60, 15)
+
+
+@pytest.fixture(autouse=True)
+def mock_pango_cairo(monkeypatch):
+    def text_size(*args):
+        return TEXT_SIZE
+
+    monkeypatch.setattr("gaphor.diagram.text.text_size", text_size)
```
danyeaw: Cool, I haven't made use of pytest's monkeypatching yet :+1:

danyeaw:
Things look great so far!

gaphor/UML/presentation.py
```diff
@@ -0,0 +1,103 @@
...
+import gaphas
+
+from gaphor.UML.uml2 import Element
+
+
+class Presentation(Element):
+    """
+    This presentation is used to link the behaviors of `gaphor.UML.Element` and `gaphas.Item`.
+
+    This class should be inherited before the
```
danyeaw: Looks like your thought trailed off here

amolenaar: Ehhmm... Yeah. 🙈

amolenaar:
@danyeaw I'm not convinced that ElementPresentation and LinePresentation is the way to go. I dislike the multiple inheritance bit. It feels a bit like mixed concerns. I'd like to see if things work better if an item *contains* a Presentation element, instead of inheriting from it.

For now I'm focusing on abstracting away the drawing shapes to some extend. The hardest part being text placement -- doing it in an obvious way.

gaphor/diagram/shapes.py
```diff
@@ -26,8 +35,19 @@ def draw_default_box(box, context, bounding_box):
...
 
 class Box:
-    def __init__(self, *children, style={}, draw=draw_default_box):
+    """
+    A box like shape.
+
+    CSS properties:
+    - min-height
+    - min-width
+    - padding: a tuple (top, right, bottom, left)
```
danyeaw: It may be more readable here if we use an [Enum](https://docs.python.org/3/library/enum.html). Then below we could reference the padding more explicitly like:
```
x = bounding_box.x + padding[Padding.LEFT]
```

gaphor/UML/presentation.py
```diff
@@ -134,6 +143,11 @@ def post_update(self, context):
     def draw(self, context):
         from gaphor.diagram.text import text_point_at_line, TextAlign
```
danyeaw: Can we move this import statement to the top of the module?

gaphor/diagram/shapes.py
amolenaar: That's a good idea.

amolenaar:
Thanks.

I'm really happy with how expressive the lines (edges) look now. I left the multiple inheritance intact, otherwise the inheritance would be replaced by composition with some extra orchestration.

I'm not sure if I should move the code for stereotypes and name+stereotype to a separate unit (function or mixin class).

The big challenge is still up: classifiers.

gaphor/diagram/shapes.py
```diff
@@ -57,6 +64,7 @@ def __init__(self, *children, style={}, draw=None):
         self._draw_border = draw
 
     def size(self, cr):
+        global Padding
```
danyeaw: If you inherit your Padding class from `Enum`, would this avoid declaring this a global?

amolenaar:
Fixes:

- Fixed issue with reopening editor window (re-registration of events)
- abstract and static property rendering (italic and underline) works again

Regressions:
- [ ] attributes and operations can not be directly edited from the diagram (requires editor now)
- [ ] interface folding is not consistent, when connecting a dependency
- [ ] arrows are missing in communication diagram messages. This item should be redesigned.
- [ ] "from package" line is missing for classifiers (classes/interfaces)

amolenaar:
I created issues for the regression items mentioned above.

PR #198, opened 2019-08-11
Add type hints to the model

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

No type hints are available.

### What is the new behavior?
Type hinting gives some advantages when working with an IDE (e.g. VSCode + Python + MyPy).

... at least, that's the premise.

- [ ] Yes
- [X] No


danyeaw:
@amolenaar Would you like me to review this or do you think it is all set?

amolenaar:
I think it's okay-ish. Gradually the typing should improve. It's not a big-bang thing. I tried to type the things that MyPy is capable of tracking.

Unfortunately, `Presentation.subject` is still referred to as a type `Any` in my VS Code, but it's a start.

I would like to use a single generic parameter for `umlproperty`, but that would mean creating separate types for the `0..1` and `0..*` properties. I found this to be too much of a change for now.

But MyPy is not complaining for now, so there's a start :)

PR #199, opened 2019-08-30
Convert Windows Install to PyInstaller

danyeaw:
The current method of building Windows packages takes about 1 hour. It requires redownloading and installing all of the needed packages to a build root, compiling everything, removing unnecessary files, creating an exe, and then packaging it into an installer. It is prone to many issues, and troubleshooting it takes a very long time because the iterations are so long.

This PR attempts to fix this by using PyInstaller instead. It takes a different approach to analyze the code and look for dependencies. This is very fast, and I can package the app in to an installer in 6 minutes - with most of the time spent building the installer and portable installer files themselves.

There are issues with PyInstaller and GTK/msys2. I have been working through these, but this is the current status:

- [X] Testing in Linux results in an error about png be an unsupported type: https://github.com/pyinstaller/pyinstaller/issues/4485
Found workaround using `env PATH="$PATH:/usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0" pyinstaller -y gaphor.spec`
- [X] Create issue #216 to capture Windows only works with console enabled.

### PR Checklist

### PR Type

- [ ] Bugfix
- [ ] Feature

Very slow windows build times.

Issue Number: N/A

### What is the new behavior?
Build times are a couple of orders of magnitude faster.

- [ ] Yes
- [X] No


amolenaar:
The code looks good. It's far less code, which is always good. Is there an installer .exe file I can try on my vanilla windows machine?

danyeaw:
@amolenaar I have uploaded an installer for testing to a [draft Release](https://github.com/gaphor/gaphor/releases/download/untagged-0bcd8a97c9f6ce3a895f/gaphor-1.0.2-installer.exe).

amolenaar:
I finally ran the installer on my (vanilla) Windows machine.

It works :)

Screenshot (1)

danyeaw:
@amolenaar Ya, now if we can only get rid of that pesky command prompt that pops up!

amolenaar:
Isn’t the ‘pythonw’ executable supposed to take care of that?

danyeaw:
We need to build the app using the windowed option, which only fails when debug mode is disabled :dizzy_face:

Issue #204, opened 2019-09-28
Support dark mode


amolenaar:
GNOME has the concept of symbolic icons. Those icons change color depending on the theme. We should concert our icons (pixmaps) to these symbolic icons, so Gaphor is at least usable in dark mode. 

Also changes to Gaphas are required to support a dark background.

lightonflux:
I couldn't find an issue for a dark background/canvas. Is this already tracked in another issue?

amolenaar:
@lightonflux You can add styling to the model. This styling will be saved as part of a model.

See https://gaphor.readthedocs.io/en/latest/style_sheets.html

![image](https://user-images.githubusercontent.com/96249/106012374-a21ea400-60bb-11eb-9335-17e158c32d30.png)

PR #206, opened 2019-09-30
Dark theme support

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Icons are hardly visible when shown on a dark theme.

Issue Number: #204 

### What is the new behavior?

SVG icons that change color depending on the theme.

- [ ] Yes
- [X] No

image

image


amolenaar:
Hmm. I think we can get rid of those. I assume SVG rendering will work on all platforms.

Issue #202, opened 2019-09-18
Get rid of deprecated GTK+ features


amolenaar:
Gaphor is still using some deprecated features from GTK+ 3. The application should be updated so that no deprecated features are used.

Deprecated features include:

* stock icons
* `Gtk.UIManager` and friends (Gtk.Action) 

This issue is a placeholder to collect information on how to perform this update.

* [Stock item deprecation](https://docs.google.com/document/d/1KCVPoYQBqMbDP11tHPpjW6uaEHrvLUmcDPqKAppCY8o/pub)
* [Freedesktop icon naming](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html#names)
* [Stock item migration sheet](https://docs.google.com/spreadsheets/d/1HavJQRPpMuq-N0GoN1wJR-9KEGXpKy3-NEPpZZkUGJY/pub?output=html)

There seems to be no real replacement for `UIManager`, esp. the ability to add and remove partial menus. We either have to come up with our own solution or stick with a more dynamic menu.

* [Using the GAction framework](http://gexperts.com/wp/using-the-gaction-framework/)

Menu generation can be done via `Gtk.Builder` (=xml), but we might just as well code something in Python directly.

(My) plan is that once we're no longer using deprecated features, we can use the `HeaderBar` and `ActionBar`, as described in the [GNOME HIG](https://developer.gnome.org/hig/stable/).

* [Introducing the CSD Initiative](https://blogs.gnome.org/tbernard/2018/01/26/csd-initiative/)

amolenaar:
Allan Day made me this lovely mockup for a new menu layout:

![new-gaphor-mockup-aday](https://user-images.githubusercontent.com/96249/65424298-7de9ed80-de0b-11e9-9206-b8a039105c79.png)

danyeaw:
That looks great!

amolenaar:
Fixed by #207.

PR #207, opened 2019-09-30
Menu upgrade

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Menus are managed by `Gtk.UIManager`, which is a deprecated API. This API will be removed in GTK4.

Issue Number: #202 

### What is the new behavior?

The menu uses the CSD structure (icons in header bar), which makes for a clean design.

- [ ] Yes
- [X] No

* All plugins should publish their menu entries in a not yet defined way. The menu is less flexible, so I probably will provide a few extension points.

### Other information

* #202 contains some references.
* A new service (RecentFiles) is introduced, so you'll have to run a `poetry install` before Gaphor runs from this branch.

## To do

- [ ] Keyboard shortcuts for toolbox items


amolenaar:
Work in progress:

image

danyeaw:
@amolenaar I still have a file menu when I tried it locally, but I don't see that in your screenshot. Are you planning on removing that as well? This is coming together, the new header bar looks great!

amolenaar:
I temporarily removed the menu for the screenshot ;), so you would have a picture of what it will look like in the end. I still want the menu around, so I know which items still need to be migrated.

amolenaar:
I integrated the editor with the main window.

image

danyeaw:
@amolenaar I just gave this a spin, this looks really good! I like the new element editor on the sidebar, and how it hides everything not essential doing some basic modeling.

amolenaar:
Thanks. Almost all functionality is in. Some plugins (python import, alignment) need fixing anyway, so I can leave them out for now and add them once they work again.

danyeaw:
@amolenaar I am getting an AttributeError when creating a block following your latest commits:
```
  File "/home/dan/Projects/gaphor/gaphor/ui/diagrampage.py", line 284, in _on_view_selection_changed
    DiagramSelectionChange(view, view.focused_item, view.selected_items)
  File "/home/dan/Projects/gaphor/gaphor/services/eventmanager.py", line 60, in handle
    self._events.handle(e)
  File "/home/dan/Projects/gaphor/gaphor/misc/generic/event.py", line 62, in handle
    handler(event)
  File "/home/dan/Projects/gaphor/gaphor/services/copyservice.py", line 48, in _update
    diagram_view.get_action_group("win").lookup_action("edit-copy").set_enabled(
AttributeError: 'NoneType' object has no attribute 'lookup_action'
```

gaphor/plugins/diagramexport/tests/test_diagramexport.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.fixture
+def canvas():
+    c = gaphas.Canvas()
+    c.add(gaphas.examples.Box())
+    return c
+
+
+def test_export_to_svg(diagram_export, canvas, tmp_path):
```
danyeaw: Nice, I didn't know that pytest had a temp path fixture builtin!

amolenaar:
> ```
> AttributeError: 'NoneType' object has no attribute 'lookup_action'
> ```

Thanks. I'll check it out. I find it somewhat hard to build unit tests around this code, since it hooks deeply into Gtk. At some points I'm dependent on running the Gtk main loop. (which is possible, but I'd rather not do)

danyeaw:
@amolenaar Ya, GUI testing is hard. We could use something PyAutoGUI, but I'm not sure it would be worth the effort.

amolenaar:
I took the liberty of merging this branch. There has been quite some change and a couple of fixes.

I would like to do the GUI automation just below the GUI. With actions this is possible and out application startup times are pretty good, so  we can start a "complete" application just for a single test. 

BTW. I was not able to reproduce the error you mentioned before.

PR #208, opened 2019-10-12
Fix AttributeError when win action group is None

danyeaw:
This fixes two errors caused during creating a new element on a diagram where it tries to look up an action on the win action group that doesn't yet exist, and when copying an item caused by the same issue. This fix adds a guard to check if the win action group is None before looking
up the action.

I am not sure if there is a more elegant fix than adding a guard to check for the None type.

### PR Checklist

### PR Type

- [X] Bugfix
- [ ] Feature

Create a block and then click on it:
```
  File "/home/dan/Projects/gaphor/gaphor/ui/diagrampage.py", line 308, in _on_view_selection_changed
    DiagramSelectionChange(view, view.focused_item, view.selected_items)
  File "/home/dan/Projects/gaphor/gaphor/services/eventmanager.py", line 60, in handle
    self._events.handle(e)
  File "/home/dan/Projects/gaphor/gaphor/misc/generic/event.py", line 62, in handle
    handler(event)
  File "/home/dan/Projects/gaphor/gaphor/services/copyservice.py", line 48, in _update
    diagram_view.get_action_group("win").lookup_action("edit-copy").set_enabled(
AttributeError: 'NoneType' object has no attribute 'lookup_action'
```

Issue Number: N/A

### What is the new behavior?
No attribute error.

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/services/copyservice.py
+++ b/gaphor/services/copyservice.py
@@ -10,7 +10,6 @@ from gaphor.UML import Element
 from gaphor.UML.collection import collection
 from gaphor.core import _, event_handler, action, transactional
 from gaphor.abc import Service, ActionProvider
-from gaphor.ui.actiongroup import set_action_state
 from gaphor.ui.event import DiagramSelectionChange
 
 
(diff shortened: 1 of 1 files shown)


gaphor/services/copyservice.py
```diff
@@ -131,9 +132,11 @@ def copy_action(self):
             for i in items:
                 copy_items.append(i)
             self.copy(copy_items)
-            view.get_action_group("win").lookup_action("edit-paste").set_enabled(
-                bool(self.copy_buffer)
-            )
+            win_action_group = view.get_action_group("win")
+            if win_action_group is not None:
```
amolenaar: Funny you’re seeing this issue and not me 🤔. I think the ‘is not None’ part is not needed. It’ll do a none check anyway.

Issue #195, opened 2019-08-04
Arrows are missing in communication diagram messages


amolenaar:
Communication messages, connecting between "Object" or lifelines contain arrows. The way this was designed is awkward. Each `UML.Message` should have an associating representation on that line, but instead the first `UML.Message` is the `subject` of that item and there is a list of additional messages, which do not have a `presentation`.

### I'm submitting a...
- [X] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

amolenaar:
Fixed by #209

PR #209, opened 2019-10-12
Cleanup message item

amolenaar:
MessageItem was a weird one. Where each item (or part of it) references a model element (via `Presentation.subject` and the model element references the presentation item (via `Element.subject`), this was no the case for messages.

With the new drawing model it became clear that it's quite hard to draw a decent message with text. The solution is to make the MessageItem a *lot* simpler.

### PR Type

- [X] Bugfix
- [X] Feature

It was possible to have multiple messages connected to a single MessageItem.

As a result, actions that concern the Message, had to be executed on a MessageItem class.

Issue Number: #195

### What is the new behavior?

The message has become simpler. Only a single message can be presented by one MessageItem.

A migration is made that creates multiple message items. One for each (additional) message previously linked to the message item.

Technically: no, functionally: yes.

### DIFF
--- a/gaphor/diagram/interactions/interactionspropertypages.py
+++ b/gaphor/diagram/interactions/interactionspropertypages.py
@@ -13,46 +13,6 @@ from gaphor.diagram.propertypages import (
 from gaphor.diagram.interactions import MessageItem
 
 
-class CommunicationMessageModel(EditableTreeModel):
-    """GTK tree model for list of messages on communication diagram."""
-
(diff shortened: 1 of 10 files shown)


amolenaar:
It was well.. unconventional, how the messages were handled. Normally we want a model element and a number of presentation items to visualize it. Now there was no relation from the model element to the presentation, in this case `MessageItem`. As a result, model-like operations had to be performed through the `MessageItem`, which is IMHO a bad idea.
I'd rather keep the model (architecture if you like) clean.

PR #213, opened 2019-10-17
Interface folding

### LINKED ISSUE #193: Interface folding is not consistent, when connecting a dependency
### I'm submitting a...
- [X] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

### Current behavior
Regression from new diagram structure it seems. 
The interaction between a component line and Interface is a bit wonky. Same with Implementation (should become a solid line when the interface is folded).

amolenaar:
Simplify the way interfaces and (mainly) components interact.

This fixes the regression after the new item drawing model was implemented.

Fixes #193 

TODO:
- [x] Interface should not display it's name in folded mode when a Connector is attached. The name is displayed on the Connector line. At least the interface name should not be bold.
image

### DIFF
--- a/gaphor/diagram/classes/classconnect.py
+++ b/gaphor/diagram/classes/classconnect.py
@@ -30,10 +30,10 @@ class DependencyConnect(RelationshipConnect):
     def reconnect(self, handle, port):
         line = self.line
         dep = line.subject
-        if handle is line.head:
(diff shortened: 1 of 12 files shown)


danyeaw:
How did you create that diagram?

amolenaar:
I copied it from the specs :)

Issue #130, opened 2019-04-05
Fix CD Pipeline for Windows and Linux


danyeaw:
Currently the CD pipeline has issues with Windows and Linux:

- [ ] Windows - each stage is only taking seconds, and no exe files are actually uploaded.
- [X] Linux - we need to find a way to upload to PyPI on a tagged version. I would like to be able to use Poetry to accomplish this, but we may have to use twine since that is what seems to be supported by azure pipelines. Also we could create a linux package using Briefcase and upload it to the release as well.

### I'm submitting a...
- [X] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

amolenaar:
This is fixed, right?

danyeaw:
Windows scripts are now working, however pytest is failing in MSYS2. This is the next thing that needs to be resolved.

danyeaw:
Pytest is working now in Windows by setting the PY_IGNORE_IMPORTMISMATCH environmental variable. The final step is to get the PyInstaller team to merge my PRs, or I need to patch the changes downstream in MINGW packages.

amolenaar:
Can I upvote the patches?

danyeaw:
Your upvote must have worked, because my PR is now merged :smile:

PR #222, opened 2019-10-27
Fix Windows Tests in CI

danyeaw:
This PR makes updates so that the tests are running in Windows on Azure Pipelines.

### PR Checklist

### PR Type

- [X] Bugfix
- [ ] Feature

Issue Number: #130 
This will partially fix this issue. However https://github.com/pyinstaller/pyinstaller/pull/4417 needs to merged to allow PyInstaller to build a package.

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/diagram/classes/tests/test_class.py
+++ b/gaphor/diagram/classes/tests/test_class.py
@@ -100,4 +100,4 @@ class ClassTestCase(TestCase):
         diagram.canvas.update()
 
         width = klass.width
-        assert width > 170.0
+        assert width >= 170.0
--- a/.azure/msys2-test.sh
+++ b/.azure/msys2-test.sh
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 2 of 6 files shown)


danyeaw:
@amolenaar Can you back me up, the change to gaphor/diagram/classes/tests/test_class.py to make it slightly more permissive OK? I'm not exactly sure why this was only failing on Windows but not other platforms, but I did check that the class was expanding as more x's were added.

gaphor/diagram/classes/tests/test_class.py
```diff
@@ -100,4 +100,4 @@ def test_compartment_resizing(self):
         diagram.canvas.update()
 
         width = klass.width
-        assert width > 170.0
+        assert width >= 170.0
```
amolenaar: This looks okay to be. Testing the width doesn't make for a really stron test anyhow, since we depend on fonts and all.

PR #224, opened 2019-10-30
Reset tool after creation by default

amolenaar:
### PR Type

- [X] Bugfix

Tool is not reset. This makes things complex for newcomers when the pointer tool has scrolled out of the toolbox view.

Issue Number: N/A

### What is the new behavior?

Reset tool after creation.

I think we should release 1.1.1 to fix the issue.

### DIFF
--- a/gaphor/ui/diagrampage.py
+++ b/gaphor/ui/diagrampage.py
@@ -215,7 +215,7 @@ class DiagramPage:
 
     @event_handler(DiagramItemCreated)
     def _on_diagram_item_created(self, event):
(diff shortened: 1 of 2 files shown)


danyeaw:
@amolenaar I am running version 1.1.0, and the tool is resetting to the pointer for me I think. Am I missing something?
Peek 2019-10-30 19-39

amolenaar:
I think you have a small config file: `~/.config/gaphor/resources`. It sets the option from there. That's how I missed it  in the first place.

PR #227, opened 2019-11-02
Add type information

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature

### What is the new behavior?

Functionality: none. The code is better typed, so that should help when you're using an IDE with code completion.

The type information can now be handled by MyPy 0.740.

- [ ] Yes
- [X] No

## Notes

For two "redefine" fields the multiplicity has been fixed in the UML model: `Implementation.contact` has a multiplicity of `"*"`, as well as `Implementation.implementingClassifier`. In the code, it already worked like this, not it's also like this in the model.


danyeaw:
Nice one! :tada:

PR #228, opened 2019-11-06
Allow nested searches on namespace view

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Search was only possible on already expanded nodes in the tree view.

### What is the new behavior?

When searching, nodes that have matching children will expand, allowing for better searching.

It would probably need some sort of highlighting, to distinguish those nodes even better.

No elements are filtered out of the view.

searching has been completely disabled for tree/list views used in the Element editor.

- [ ] Yes
- [X] No


danyeaw:
:+1:

PR #231, opened 2019-11-08
Edit text popover

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

A multi-line edit window pops up. Even if the input can only be a single line; which is always the case, except for Comment's.

### What is the new behavior?

image

Popovers can be activated by a keyboard shortcut (F2).

- [X] Yes
- [ ] No

Well, I replace the `Editor` code with a new one: `InlineEditor`.


gaphor/diagram/inlineeditors.py
```diff
@@ -0,0 +1,98 @@
...
+
+from gi.repository import Gdk, Gtk
+from gaphas import Item
+from gaphas.geometry import Rectangle
+
+from gaphor import UML
+from gaphor.core import transactional
+from gaphor.diagram.presentation import (
+    Named,
+    Classified,
```
danyeaw: I don't think this import is being used.

gaphor/diagram/inlineeditors.py
```diff
@@ -0,0 +1,98 @@
...
+
+from typing import Optional, Tuple
+
+from functools import singledispatch
+
+from gi.repository import Gdk, Gtk
+from gaphas import Item
+from gaphas.geometry import Rectangle
+
+from gaphor import UML
```
danyeaw: I don't think this import is being used.

danyeaw:
@amolenaar This is beautiful now, really nice job :heart_eyes:

PR #246, opened 2019-12-11
Cleanup diagram toolbox

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature

When a new diagram item was created, one had to add it to `gaphor.ui.diagramtoolbox`.

### What is the new behavior?

Toolbox functionality is located closer to the actual `Item`.

- [ ] Yes
- [X] No


gaphor/diagram/diagramtools.py
```diff
@@ -158,7 +159,8 @@ class PlacementTool(_PlacementTool):
     PlacementTool is used to place items on the canvas.
     """
 
-    def __init__(self, view, item_factory, after_handler=None, handle_index=-1):
+    # TODO: I would expect ElementFactory to be provided as well.
```
danyeaw: Is there anything we need to do to close out this TODO?

amolenaar: Currently the element factory is accessed though `diagram.model` (`view.canvas.diagram.model` to be precise). I think this is not ideal, and violates the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter). I wanted to wait for this to fix until most of the code is in place.

I'm a bit occupied with Advent of Code at the moment, though :D.

amolenaar: It's not as bad as I thought. I think using just `diagram.model` is fine. It makes the code less cluttered at least.

danyeaw:
:tada:

Issue #138, opened 2019-04-13
Add Support for SysML


danyeaw:
The Systems Modeling Language (SysML) is a general-purpose modeling language for systems engineering applications. UML is focused on designing software, SysML allows you to design a complex system like an Autonomous Vehicle, Mars Rover, or Medical Device. It allows you to connect the behavior, structure, and requirements of a system through multiple abstraction levels. This is why I am interested in Gaphor, having an open source and simple modeling tool would be a huge boost to teaching engineers how to design complex systems. SysML is a subset and extension of UML. 

## High Level Changes
There are changes to the diagrams, including adding additional diagrams:
1. Requirements Diagram
1. Parametric Diagram
- This is a low impact for Gaphor since we only have 1 diagram type.

### Structure
Object, Profile, Deployment, and Component diagrams are removed.
- For Gaphor, the Component and Profile portions of the Palette could be hidden / removed

### Behavior
Communication, Interaction Overview, and Timing diagrams are removed. Protocol state machines are removed (behavior state machines remain).
- For Gaphor, what we support for Interactions could be renamed Sequence

## High Priority Changes

1. Blocks (Class)
- Associations have no arrowheads on both ends, which is the default we display them as well

2. Internal Blocks
- 4 properties of Blocks, but only parts, references are high priority
- Proxy ports and flow specifications

3. Requirement (Class)
- DeriveReqt, Verify, Satisfy (Trace)

4. Change Palettes 
- Hide Classes, Components, and Profiles Palettes
- Rename Interactions to Sequence
- Add Block, Internal Block, and Requirements Palettes

## Medium Priority Changes
1. Blocks – value properties and constraint properties
1. Allocation
1. Do we need a way to make an Activity (defining reusable behavior of an Action)?

## Low Priority Changes:
1. Parametric Diagram - I'm not a big fan of the usefulness of modeling parametrics
2. Model Elements support
- Conform (dependency)
- View (package)
- Viewpoint (Class)
- Rationale and Problem (Comments)
3. ValueType
4. QuantityKind / Unit
5. Block value properties and constraint properties
6. Activity extensions
- Control as Data
- Continuous Systems
- Probability
7. Copy (Trace)
8. Constraint Blocks

### I'm submitting a...
- [ ] Bug report  
- [X] Feature request
- [ ] Documentation issue or request

danyeaw:
I have created a new branch called sysml-support. In it I have created two profiles:
[...]

PR #260, opened 2020-01-12
Add Support for SysML and Safety & Reliability Analysis Profile

danyeaw:
Adds the ability to select a modeling profile beyond UML. Adds support for SysML 1.6 and the soon to be released Safety and Reliability Analysis Profile 1.0.

### PR Checklist

### PR Type

- [ ] Bugfix
- [X] Feature

Issue Number: #138 

- [ ] Yes
- [X] No


danyeaw:
The current UML profile capabilities does not include the <<import>> or <<apply>> dependencies which are part of the UML spec. In order to tell the model to import, for example UML, my plan is to add this dependency. The most general use case of this would be to allow users to import any gaphor model through a file selection dialog. What we need for being able to add the SysML and Safety & Reliability Profiles is only to import the UML and SysML profiles themselves. I think for this use case we can just import the profile through searching for a model with the package name. Any thoughts on this approach?

amolenaar:
Import and apply dependency sounds like a good plan. If all models are found in a `gaphor.models` module (UML too) we can just do a simple "import" for the dependent model entry. Or use the overrides file, but that's less fancy. I do not think the UML model needs to be treated any different from the SysML models.

I would not use a "cold" module lookup. That will inhibit us from doing any changes to Gaphor's structure. Maybe a module where all metamodels come together so we can select from there?

danyeaw:
@amolenaar I like the idea to put everything in `gaphor.models`. Can you explain more what you mean by a cold module lookup?

amolenaar:
I think I was thinking about loading and saving models. I think we want a level of indirection between how it's saved and how it's discoverable in Gaphor.

danyeaw:
## Status

Currently I am able to draw the import relationship on a diagram, and constrain the dependency between Packages. The relationship is also successfully saving in the gaphor file format, since the UML data model already had support for this.

## Design Thoughts on Referencing Across Profiles

If we want, for example, the SysML profile to define an element as an extension of a Class, I think we need to reference the Class using either the Gaphor models or the generated data models.
For the Gaphor models, the SysML profile needs to use the same ref-id as the one in the UML Gaphor model. If we use reference them using the data model, then we can look for an element called Class in the UML data model. 

One advantage of referencing the elements using the Gaphor models is this might be a nice Feature to allow any model to import another model for reusability. This Import Model option could be placed in the hamburger menu. If we were to do this, then I think the cleanest approach would be to create a new tag in the file format to reference another model. Then when that model is loading it loads the other model as well. There are ways this starts to get complicated:
1. Do we reference the other model from a relative or absolute path?
2. What happens if the referenced model can't be found? Do we then have to bring up a dialog to ask the user to select the updated location?

On the other hand, I think referencing the elements in the data model may be simpler. Especially if we use the import relationship, then we can use that to search for elements of the same name in the namespace we are adding. I do need to dig in to this a bit more.

Any other thoughts or comments on the best method to implement this?

danyeaw:
Breaking this up in to multiple more manageable PRs.

Issue #143, opened 2019-04-15
Add execution specification in UML sequence diagrams


ruimaciel:
Currently Gaphor offers some support for interaction diagrams.  However, it doesn't provide a way to add [execution specifications](https://www.uml-diagrams.org/sequence-diagrams.html#execution) in sequence diagrams.  

Therefore, although Gaphor enables users to represent lifelines and messages, it does not offer the user a way to represent where the control flow is executing and which lifeline is waiting for synchronous messages.

amolenaar:
I start working on this one. 

It's a nice case to work a bit with Gaphor. I'm updating the Interaction model to UML 2.5. And fixing some odds and ends in Gaphor itself

amolenaar:
@ruimaciel 
![image](https://user-images.githubusercontent.com/96249/75111856-8eaec180-563e-11ea-9b6e-48ed4537c49e.png)

A lot still has to be done, but the item is at least on the canvas.

amolenaar:
@ruimaciel @danyeaw 
![image](https://user-images.githubusercontent.com/96249/75592182-ec4d6e80-5a81-11ea-9250-e5cf64258513.png)

Almost there.

amolenaar:
Fixed by #273

PR #273, opened 2020-02-24
Sequence diagram updates

amolenaar:
### PR Type

- [X] Feature

Issue Number: #143 

### What is the new behavior?

- [x] An Execution Specification (ExecSpec) can be added to a Diagram
- [x] ExecSpec can be connected to a Lifeline
- [x] Messages can attach to an ExecSpec
- [x] ExecSpecs can be attached to an ExecSpec
- [x] Icon for tool box
- [x] Saving and loading
- [x] OccurrenceSpecifications should be ordered on the Lifeline [1]

Churn:
- [X] Rename the `IConnect` interface to `Connector`.

NB. detach ExecSpec with Messages or other ExecSpecs attached!

- [X] Yes
- [ ] No

### Other information

[1] If Gaphor orders the OccurrenceSpecifications in accordance to the diagram, one can actually do something with such a diagram, e.g. generate a test script that validates the messages with the real application.


amolenaar:
I'm not happy with the current state of the PR. In order to create an Execution Specification, one has ro connect a handle on the ExecSpec item. This will also make the ExecSpec a child item of the Lifeline or ExecSpec. This seems like a bit too much. For one, I find it weird to be able to connect one handle (the other handle is not _connectable_).

It should just "drop" on the lifeline. However, the current grouping code does not align child elements in a particular way.

There's still some work to do.

amolenaar:
I'll keep it as is for now: a handle on the exec spec should connect to the lifeline.
The grouping protocol is not suited to handle constraints, so that would involve a fair amount of work.

PR #277, opened 2020-03-08
Move element editor layouts to Glade

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature

All GTK+ layout is hard-coded in Python.

### What is the new behavior?

Layouts are loaded from `Gtk.Builder` (Glade) files.

This has the advantage that it's easier to visibly change the layout of a component. This will lead to less code, and can result in a more snappy UI.

- [ ] Yes
- [X] No


amolenaar:
@danyeaw I think the element editor looks better now. Also, some small issues I ran into have been fixed.

danyeaw:
@amolenaar very cool, I think this helps separate the GUI from the other code and logic :+1:

PR #278, opened 2020-03-17
Remove flatpak, as it is in a repo on the Flathub repo

amolenaar:
No need up keep Flatpak code here. It's on https://github.com/flathub/org.gaphor.Gaphor.


danyeaw:
Great idea! I think we need to also update `docs/linux.md` since it references how to build a flatpak locally by going to the flatpak directory.

PR #289, opened 2020-04-02
Remove CI workaround for console plugin

amolenaar:


danyeaw:
It looks like GitHub still hasn't fixed this bug :rofl:

amolenaar:
I tried the trimmed doen expample you made and that built, so I thought...

danyeaw:
@amolenaar Since this is now only failing with Python 3.7, we could remove 3.7 from the build matrix. I am completely fine with us only testing Gaphor on the latest Python version.

I'll keep the Issue open with GitHub, but that would allow us to merge this PR.

Issue #290, opened 2020-04-03
Application (window) hangs when loading an invalid model


amolenaar:
### I'm submitting a...
- [X] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

### Current behavior
Open a text file for a change.

### Expected behavior
An error message and the window should be usable again.

### Minimal reproduction of the problem with instructions
Load small non-xml file

danyeaw:
Closed by #291.

PR #291, opened 2020-04-03
Fix loading file with errors

amolenaar:
### PR Type

- [X] Bugfix
- [ ] Feature

Issue Number: #290 

### What is the new behavior?

An error is presented when a loading error is experienced.

Also, removing a non-existant element no longer logs an error. It's desired state,  so we're good.

- [ ] Yes
- [X] No


danyeaw:
Nice, I really like this!

PR #293, opened 2020-04-05
Add Profile Selection Combo Box for SysML

danyeaw:
This PR adds a drop down combo box to select a new profile.

### PR Checklist

### PR Type

- [ ] Bugfix
- [X] Feature

Issue Number: #138 

- [ ] Yes
- [X] No


gaphor/diagram/connectors.py
```diff
@@ -339,8 +339,10 @@ def disconnect(self, handle: Handle) -> None:
 
 
 class RelationshipConnect(UnaryRelationshipConnect):
-    """
-    """
+    """Base for relationship connections between unique elements."""
+
+    def connect_subject(self, handle: Handle) -> bool:
```
amolenaar, line 344: Why do we need this function? Now if someone forgets to implement it, it will not break

gaphor/ui/toolbox.py
```diff
@@ -85,27 +116,55 @@ def on_collapsed(widget, prop, index):
...
+    def _on_profile_changed(self, event: ProfileSelectionChanged) -> None:
+        """Reconfigures the toolbox based on the profile selected.
+
+        When the combo box drop down to select the profile changes
+        (UML, SysML, Safety), this event handler reconfigures the
+        toolbox.
+
+        Args:
+            event: The ProfileSelectionChanged event.
+
```
amolenaar, line 161: The `Args` comment does not add a lot of value. The info is in the signature already.

danyeaw: Good catch. I think a docstring could still add extra information that you can't capture in a type hint: describe what the parameter is used for and why a particular return value is chosen. You are right that I shouldn't just be restating the type hint in a docstring, that doesn't add any value. I'll keep this in mind in the future as well, I think I had in my mind that we always need a full docstring, but I agree that type hints may often make that not necessary. :+1:

gaphor/diagram/connectors.py
danyeaw: You are correct, I thought one of the linters was giving me an error that one of the methods of the ABC wasn't implemented. :+1:

PR #295, opened 2020-04-08
Move core modelling concepts to a separate package

amolenaar:
### PR Type

Part of #138.

Evenrything is in `gaphor.UML`, which makes for no logical separation when extending Gaphor with other modelling types.

### What is the new behavior?

- [ ] Move core modelling functionality to gaphor.core
  - `Element`
  - `Presentation`
  - `ElementFactory`
- [ ] Find out what to do with `Diagram` (it's a `NamedElement`)

- [X] Yes
- [ ] No


amolenaar:
I think this is done, at least for the model.

utils/model/build_core.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
+"""
+This file provides the code generator which transforms gaphor/UML/uml2.gaphor
+into gaphor/UML/uml2.py.
+
+Also a distutils tool, build_uml, is provided.
```
danyeaw, line 6: I think we need to update the docstring here for the new module

utils/model/build_core.py
```diff
@@ -0,0 +1,44 @@
...
+import os.path
+from distutils.core import Command
+from distutils.dep_util import newer
+from distutils.dir_util import mkpath
+from distutils.util import byte_compile
+
+from utils.model import gen_uml
+
+
+def generate_uml2(force=False):
```
danyeaw, line 18: Should we rename this to generate_core?

gaphor/core/modeling/diagram.py
```diff
@@ -0,0 +1,122 @@
+"""This module contains a model element Diagram which is the abstract
+representation of a UML diagram. Diagrams can be visualized and edited.
```
danyeaw, line 2: maybe just any model diagram instead of UML diagram?

utils/model/build_core.py
amolenaar: Actually I want to have a proper script, with command line arguments that I can use instead. To avoid confusion I'll change this, though ;)

danyeaw:
@amolenaar Windows builds are failing. I submitted an [issue](https://github.com/msys2/MINGW-packages/issues/6368) to help try to resolve it.

PR #297, opened 2020-04-11
Refactor Code Generator to New Module and add CLI

danyeaw:
This is a refactoring as part of in the future supporting SysML. It takes the util modules for parsing the gaphor models and generating the data model, and moves them to a separate codegen module. I also used the builtin argparse to add a CLI.

### PR Checklist

### PR Type

- [ ] Bugfix
- [ ] Feature

Issue Number: N/A

- [ ] Yes
- [X] No


danyeaw:
@amolenaar I cleaned up a handful of mypy errors that I knew how to fix. I am not sure what to do with the remaining ones, should I put ignore comments on these lines?

amolenaar:
I think that’s fine for now. There was a reason I did not run Mypy on the utils folder 🙄.

Makefile
```diff
@@ -18,12 +18,12 @@ icons:		## Generate icons from stensil (requires Inkscape)
 translate:	## Translate and update .po and .mo files (requires PyBabel)
 	$(MAKE) -C po
 
-model: gaphor/core/modeling/coremodel.py gaphor/UML/uml2.py	## Generate Python model files from Gaphor models (requires Black, MyPy)
+model: gaphor/core/modeling/coremodel.py gaphor/UML/uml.py	## Generate Python model files from Gaphor models (requires Black, MyPy)
 
-gaphor/core/modeling/coremodel.py: models/Core.gaphor models/Core.override utils/model/gen_uml.py utils/model/override.py utils/model/writer.py
-	utils/model/build_core.py && black $@ && mypy gaphor/core/modeling
+gaphor/core/modeling/coremodel.py: models/Core.gaphor models/Core.override gaphor/codegen/autocoder.py gaphor/codegen/override.py gaphor/codegen/writer.py
+	gaphor/codegen/codegen.py Core && black $@ && mypy gaphor/core/modeling
```
amolenaar, line 24: I would use specific arguments for the model, overrides and output files.

danyeaw: @amolenaar Thanks so much for the feedback, I guess I was trying to make things a little too tricky! :100: Please let me know if you like the updated version better.

gaphor/codegen/codegen.py
```diff
@@ -0,0 +1,59 @@
...
+    parser.add_argument(
+        "overrides",
+        type=Path,
+        help="override filename, default location is the "
+        "models package if full filename not given.",
+    )
+    args = parser.parse_args()
+    modelfile: Path = args.modelfile
+    outfile: Path = args.outfile
+    if str(modelfile) == modelfile.name:
```
amolenaar, line 39: I think we do not need the `if`-blocks. Let's just be specific in the arguments instead. If we can not resolve a file, just exit.

gaphor/codegen/tests/test_autocoder.py
```diff
@@ -17,32 +17,33 @@ def close(self):
 
 
 def test_loading():
-    model_file = distribution().locate_file("tests/test-model.gaphor")
+    model_file = distribution().locate_file("gaphor/codegen/tests/test-model.gaphor")
```
amolenaar, line 20: Should we move the test model to test-diagrams?

I know it's not ideal, since unit tests are packaged with the production code. Now there are several tests already that break because they depend on diagrams outside the source (code) tree. 

Either we should move all these tests out of the source tree into `/tests/` or include all test diagrams in the source tree. I'd opt for the first.

amolenaar:
Despite my comments, I think the PR is good to go. It's a major improvement, so it belongs on _master_. :+1:

gaphor/codegen/tests/test_autocoder.py
danyeaw: Thanks, I agree that we should keep the models out of the source packages. Thanks for catching that! :+1: Since you created a new models package, I thought it made sense to move test-diagrams to test-models and nest it under where the rest of the models are located. What do you think about that?

gaphor/codegen/codegen.py
danyeaw: Ok, I'm good with that! :smile_cat:

amolenaar:
I’m not sure if test-models should belong under models or if it’s more like a sibling.

amolenaar:
Oh well...

amolenaar:
Parent: #138

PR #300, opened 2020-04-12
Place UML model and diagram items closer together

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

The code is all around the UML model

### What is the new behavior?
There is a distinction between the generic Gaphor functionality and the UML model. This should provide a first example of how to implement models as part of Gaphor.

- [X] Yes - module naming
- [ ] No

## Notes:

Moved `Comment` model element to the Core module. This reduces the dependency on UML model from the diagram model to one instance: meta-classes in property pages.


amolenaar:
Part of #138

PR #301, opened 2020-04-13
Model Provider Service

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Everything depends on the UML data model.

Issue Number: #138 

### What is the new behavior?

Model specific information is obtained via a service. Having multiple services allows us to support different data models.

- [X] Yes
- [ ] No

Tasks:

- [x] populate model selection dropdown from ModelProvider
- [x] Create a model provider that aggregates individual model providers
- [x] Create separate lookups for model elements and diagram items (replace `gaphor/storage/diagramitems.py`)
- [x] Move Sanitizer service to `gaphor.UML`
- [x] `ElementDispatcher` should use new model lookup
- [x] `storage.load` should use new model lookup
- [X] `gaphorconvert,py` should load all model providers - via ModelProvider
- [x] Rename `ModelProvider` to `ModelingLanuage`, as it covers the intent better.

Postpone these tasks, otherwise the PR will grow even more:

- Add namespace filter to ModelProvider
- Should create a ModelProvider for the core model as well? - Not now
- Should Namepace also be part of the core model? (Namespace view depends on it) - not now
- Create elements via `ElementFactory.create()` by name (string). E.g. `element_factory.create("Package")`.
- move property page glade file to UML
- what to do with icons?


gaphor/diagram/propertypages.py
```diff
@@ -312,6 +311,8 @@ def __init__(self, subject: NamedElement):
         self.watcher = subject.watcher() if subject else DummyEventWatcher()
 
     def construct(self):
+        from gaphor import UML
```
danyeaw, line 314: Do we need a comment here explaining why this import can't be at the top of the module?

gaphor/services/modelprovider.py
```diff
@@ -0,0 +1,52 @@
...
+class ModelProviderService(Service, ModelProvider):
+
+    DEFAULT_PROFILE = "UML"
+
+    def __init__(self, properties={}):
+        """
+        Create a new Model Provider. It will provide all models defined
+        as entrypoints under `[gaphor.modelproviders]`.
+
+        Te `properties` argument is optional, in which case the service will
```
danyeaw, line 16: Typo

danyeaw:
This looks great, you are on a roll! :rocket:

gaphor/diagram/propertypages.py
amolenaar: Maybe it needs to move to the UML package as well. I'm not sure yet.

amolenaar:
Thanks. It's really nice to see when a plan comes together. It all falls into place :)

gaphor/diagram/propertypages.py
amolenaar: I moved the NamedElement editor to the UML package.

amolenaar:
@danyeaw Something like this? :)

gaphor/SysML/modelinglanguage.py
```diff
@@ -0,0 +1,27 @@
+"""
+The UML Model Provider is the entrypoint for an application to
+UML related assets.
```
danyeaw, line 3: Should we update this docstring for SysML?

danyeaw:
@amolenaar This look really great! :tada: 

I did notice when doing some brief operational testing that I was getting some TypeErrors. For example, put a Class on a diagram and then click off of it so it is no longer selected:
`TypeError: unsubscribe_all() takes 0 positional arguments but 1 was given`

gaphor/SysML/modelinglanguage.py
amolenaar: Done.

amolenaar:
I fixed the `TypeError`.

danyeaw:
@amolenaar Really nice job! :shipit: 🎉

amolenaar:
Thanks 🤗

PR #303, opened 2020-04-19
Generate the SysML Data Model from SysML Gaphor Model

danyeaw:
This PR adds a profile_coder module that is used to generate a SysML data model in Python from the SysML gaphor model. The module uses loads the Gaphor model in to an ElementFactory and then generates the data model from there. This is part of completing #138.

This is a WIP, so far it just opens the Gaphor model and prints all of the Class definitions to a file, but doesn't include any attributes or inheritance.

### PR Checklist

### PR Type

- [ ] Bugfix
- [X] Feature

Issue Number: N/A

- [ ] Yes
- [X] No


gaphor/codegen/profile_coder.py
```diff
@@ -0,0 +1,85 @@
...
+    services = [
+        "element_dispatcher",
+        "event_manager",
+        "component_registry",
+        "element_factory",
+        "modeling_language",
+    ]
+    session = Session(services=services)
+    element_factory = session.get_service("element_factory")
+    modeling_language = session.get_service("modeling_language")
```
amolenaar, line 49: I made a small fix on master, so we can just do:

```
element_factory = ElementFactory()
modeling_language = UMLModelingLanguage()
```

amolenaar:
That's some really nice progress. Only need to resolve the Extensions, so they can be displayed as inheritance relationships.

We'll need a little function that orders the classes, so they are are written with parent classes at the top.

gaphor/codegen/profile_coder.py
danyeaw: Thanks for updating that! I'll merge master back in to this branch and start to make use of that.

danyeaw:
My current ordering method was to just run through the list of classes three times:
1. Write UML imports
2. Write Classes with no inheritance
3. Write Classes with inheritance

I could be a little fancier and reorder the list of classes, I'll see if this makes the code more readable.

amolenaar:
The model is generated. Looks like this can be merged, right?

PR #312, opened 2020-05-08
Use custom icon when placing icon on the diagram

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

If you select a placement tool, you can not see that by the cursor in the diagram. This is inconvenient, esp. when you're using the keyboard shortcuts. At least, I found this slightly annoying last evening ;).

### What is the new behavior?

When selecting a tool in the toolbox, the cursor in the diagram reflects this selction. This makes it easier to see what action will be performed.

- [ ] Yes
- [X] No


gaphor/ui/diagrampage.py
```diff
@@ -40,6 +41,27 @@ def tooliter(toolbox_actions: Sequence[Tuple[str, Sequence[ToolDef]]]):
...
 
 
+with importlib.resources.path("gaphor.ui", "placement-icon-base.png") as f:
+    PLACEMENT_BASE = GdkPixbuf.Pixbuf.new_from_file_at_scale(str(f), 64, 64, True)
+
+_placement_pixbuf_map: Dict[str, GdkPixbuf.Pixbuf] = {}
+
+
+def get_placement_cursor(display, icon_name):
+    global _placement_pixbuf_map
```
danyeaw, line 51: Why is the global required here, can we just define the `_placement_pixbuf_map` in the `get_placement_cursor` function?

amolenaar: Indeed I think we can do without this. I’ll remove it.

amolenaar:
I noticed I missed this while tinkering with the SysML model 😄

Issue #286, opened 2020-03-31
Can't open diagram with a stalled(?) comment node


actionless:
### I'm submitting a...
- [x] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

### Current behavior
Loading stucks with the exception pritned:
```python
gaphor.storage.storage INFO Loading file deployment_diagram.gaphor
gaphor.storage.parser ERROR File corrupted, remove element 8cef44be-7376-11ea-80af-e4f89cae6ece and try again
NoneType: None
gaphor.storage.storage INFO Read 140 elements from file
gaphor.storage.storage WARNING file b'/home/lie/projects/test/deployment_diagram.gaphor' could not be loaded (module 'gaphor.UML' has no attribute 'CommentItem')
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/gaphor/misc/gidlethread.py", line 127, in __generator_executer
    result = next(self._generator)
  File "/usr/lib/python3.8/site-packages/gaphor/storage/storage.py", line 311, in load_generator
    for percentage in load_elements_generator(
  File "/usr/lib/python3.8/site-packages/gaphor/storage/storage.py", line 182, in load_elements_generator
    yield from _load_elements_and_canvasitems(
  File "/usr/lib/python3.8/site-packages/gaphor/storage/storage.py", line 222, in _load_elements_and_canvasitems
    cls = getattr(UML, elem.type)
AttributeError: module 'gaphor.UML' has no attribute 'CommentItem'
```

### Minimal reproduction of the problem with instructions
1) create a diagram with comment element (probably it was due to adding and then removing the comment node, see example files attached in one of the comments below)
2) save it
3) close the Gaphor app
4) reopen the diagram again

### System Information
OS:
- [x] Linux (Please put in notes the specific distro)
- [ ] MacOS
- [ ] Windows 10
- [ ] Windows 7

actionless:
while looking further i found weird comment item in the .gaphor diagram file:

```diff
-<CommentItem id="8cef44be-7376-11ea-80af-e4f89cae6ece">
-<matrix>
-<val>(1.0, 0.0, 0.0, 1.0, 135.90929412841797, 720.6576538085938)</val>
-</matrix>
-<width>
-<val>100.0</val>
-</width>
-<height>
-<val>94.0</val>
-</height>
-<subject>
-<ref refid="8cef3992-7376-11ea-80af-e4f89cae6ece"/>
-</subject>
-</CommentItem>
[...]

PR #317, opened 2020-05-10
Copy paste with parent items

amolenaar:
### PR Type

- [X] Bugfix
- [X] Feature

Items, part of a super item, such as Components in a Node, are not copied properly.

Issue Number: #286 

### What is the new behavior?

* When a nested item is copied, it's pasted in the same parent item
* When child and parent item are copied together, a new parent item is created and the child will be part of that

Note that items are still pasted relative to the item position (position to the parent). Gaphor still needs some work so pasted items are always pasted inside the canvas' view port.

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/UML/components/component.py
+++ b/gaphor/UML/components/component.py
@@ -52,6 +52,10 @@ class ComponentItem(ElementPresentation, Classified):
             draw=draw_border
         )
 
+    def postload(self):
+        self.update_shapes()
+        super().postload()
+
 
 def draw_component_icon(box, context, bounding_box):
     bar_width = 12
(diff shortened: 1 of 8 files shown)


amolenaar:
It's failing when I update the relationship for UseCase.subject to bi-directional. Other end is `Classifier.useCase`.

actionless:
thanks a lot! now it seems almost perfect, except for some small sporadic padding changes:

2020-05-14--1589461421_1920x1080_scrot

2020-05-14--1589461425_1920x1080_scrot

actionless:
(looks like top and left padding are multiplied by 2)

amolenaar:
Thanks a lot for testing this. I see what’s going on: as part of a paste action, the offset of the elements is changed by 10 pixels. This way it’s clear to a user the new item is pasted. Hoever, it also applies this offset to the nested items. That should not be the case, of course.

amolenaar:
Fixed it in 74c82d4ea53710b7ee27308b84ad39b6445e35e5.

actionless:
i still find it a bit weird what the whole copy-pasted fragment is padded extra 10px, i'd probably would make some sort of dotted line highlight around element, unless you click once somewhere

but that's not a blocker anymore, thanks a lot for your work!

Issue #326, opened 2020-05-24
Error while opening a previously saved project


gavr123456789:
### I'm submitting a...
- [X] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

### Current behavior

Open .gaphor file and getting error:  
![image](https://user-images.githubusercontent.com/30507409/82745909-609e6200-9d92-11ea-99bb-28aa85228690.png)

### Expected behavior
File opened successfully.

### System Information
OS:
- [X] Linux (Please put in notes the specific distro)
- [ ] MacOS
- [ ] Windows 10
- [ ] Windows 7

NOTES:
Attach a file. (Format was changed for GitHub)

[krontech Arch.txt](https://github.com/gaphor/gaphor/files/4672960/krontech.Arch.txt)

gavr123456789:
That how it looks when last time edited  
![image](https://user-images.githubusercontent.com/30507409/82745992-44e78b80-9d93-11ea-8c91-33bdf76eb60f.png)

amolenaar:
Hi @gavr123456789,
Thanks again for sharing an issue. 

The issue is looks identical to #286, where a diagram item ends up in the model.

Which version of Gaphor are you running? 1.2.0 or one of the latest development builds? 
Was saving and loading done with the same version?
Did you use the copy/paste to copy items, associations in this case? (I suspect v1.2.0 copy/paste code, so the issue should not happen for recent builds)
Anything you know that can help reproduce this issue?

I attached your fixed model: [krontech.fixed.zip](https://github.com/gaphor/gaphor/files/4673195/krontech.fixed.zip).

gavr123456789:
1.3.0  
Not sure about update, i bet 70% thats yes.  
No, or used and than deleted copy of line.  
It's been 2 days since I created this file. I don't remember exactly what led to this. I think that I was erasing automatically appearing brackets and possibly pros.

amolenaar:
Thanks.

Well, that leaves me completely blank how these elements ended up in the model :disappointed:. I assume the extra checks in #327 will at least allow for earlier warning.

PR #327, opened 2020-05-24
Add extra model integrity checks

amolenaar:
### PR Type

- [X] Bugfix
- [ ] Feature

It used to be possible for diagram items to end up in the model. This leads to issues when loading models.

Issue Number: #286, #326

### What is the new behavior?

Extra checks are done for the type sanity for both `Element Factory` (which contains the model) and `Diagram` (where diagram items are added).

This should ensure the model stays clean.

`ElementFatctory.bind()` has been removed since it was not used (nor tested).

- [ ] Yes
- [X] No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ htmlcov
 .eggs/
 gaphor.egg-info/
 tmp.gaphor
+gaphor.prof
 .gpg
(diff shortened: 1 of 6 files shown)


danyeaw:
:+1:

PR #330, opened 2020-05-30
Parts and references

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

No parts and references compartment in Block diagram

Issue Number: #138 

### What is the new behavior?

Parts and references are visible:

image

- [ ] Yes
- [X] No

NB. It looks like SysML elements are not visible in the tree view at all (!)

### DIFF
--- a/gaphor/SysML/blocks/block.py
+++ b/gaphor/SysML/blocks/block.py
@@ -20,9 +20,6 @@ from gaphor.diagram.support import represents
 from gaphor.SysML.sysml import Block
 from gaphor.UML.classes.klass import (
     attribute_watches,
-    attributes_compartment,
-    operation_watches,
-    operations_compartment,
     stereotype_compartments,
     stereotype_watches,
 )
@@ -36,12 +33,8 @@ class BlockItem(ElementPresentation[Block], Classified):
         super().__init__(id, model)
 
         self.watch("show_stereotypes", self.update_shapes).watch(
-            "show_attributes", self.update_shapes
-        ).watch("show_operations", self.update_shapes).watch(
             "show_parts", self.update_shapes
-        ).watch(
-            "show_references", self.update_shapes
-        ).watch(
+        ).watch("show_references", self.update_shapes).watch(
             "subject[NamedElement].name"
         ).watch(
             "subject[NamedElement].namespace.name"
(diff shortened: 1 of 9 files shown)


gaphor/SysML/blocks/block.py
```diff
@@ -1,9 +1,154 @@
...
+                and [operations_compartment(self.subject)]
+                or []
+            ),
+            *(
+                self.show_parts
+                and self.subject
+                and [
+                    self.block_compartment(
+                        gettext("parts"),
+                        lambda a: a.association and a.aggregation != "none",
```
danyeaw, line 108: I think this should be `a.aggregation == "composite"`

gaphor/SysML/blocks/block.py
```diff
@@ -1,9 +1,154 @@
...
+                ]
+                or []
+            ),
+            *(
+                self.show_references
+                and self.subject
+                and [
+                    self.block_compartment(
+                        gettext("references"),
+                        lambda a: a.association and a.aggregation == "none",
```
danyeaw, line 119: I think this should be `a.aggregation == "none" or a.aggregation == "composite"`

amolenaar:
I was a bit in doubt on the shared association. I'll change it accordingly.

gaphor/SysML/blocks/block.py
amolenaar: Fixed

gaphor/SysML/blocks/block.py
amolenaar: And fixed :smile:

amolenaar:
Removing attributes and operations is trivial now, since Block is a separate diagram item now, not inheriting from ClassItem.

Isn't it custom to have attributes for, for example, `maxPSI` on an `InflationValve` and a `thresholdPSI` on a `TirePressureMonitor`?

(using American units here, we use Bar as unit for pressure :smiley: )

danyeaw:
Those would be added by a 3rd Property that we haven't created yet called a Value Property. To support those we would need to add some default ValueTypes and make it possible to add additional ones, since every value property is typed by a ValueType.

amolenaar:
Then we'd better remove attributes and operations, before they are misused. Or maybe keep attributes until we have ValueTypes in place?

danyeaw:
I would be ok with just removing them I think, and then once we get initial SysML support finished, we can come back in with additions like ValueTypes.

amolenaar:
Fair enough. Sounds like a simple change to me.

amolenaar:
I removed attributes and operations.

Issue #334, opened 2020-06-01
SysML component do not show up in tree view


amolenaar:
### I'm submitting a...
- [X] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

### Current behavior

Blocks, Requirements (maybe more?) do not show up in the namespace view

### Expected behavior

They should be visible at least, and be interacted with, just like classes.

To make this more generic: we should basically show all elements. This will normally clutter the tree view with associations, properties and the like.

I think Magicdraw uses an (artificial) group that contains all (named) relationships. This may also be a solution.

### Minimal reproduction of the problem with instructions

Add a Block or Requirement to a new diagram.

NOTES:

This issue relates to #275.

amolenaar:
Fixed by #336.

PR #336, opened 2020-06-01
Namespace show all

amolenaar:
### PR Type

- [X] Bugfix
- [X] Feature

Issue Number: #334 

### What is the new behavior?

All namespaced information is shown.

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/UML/tests/test_umllex.py
+++ b/gaphor/UML/tests/test_umllex.py
@@ -145,6 +145,21 @@ def test_parse_association_end_derived_end(factory):
     assert not p.defaultValue
 
 
+def test_parse_association_end_with_type(factory):
+    """Test parsing of association end, type is ignored
+    """
+    a = factory.create(UML.Association)
+    p = factory.create(UML.Property)
+    p.association = a
+
+    UML.parse(p, "end: TypeVal")
+    assert "end" == p.name
+    assert not p.typeValue
+    assert not p.lowerValue
+    assert not p.upperValue
+    assert not p.defaultValue
+
+
 def test_parse_operation(factory):
     """Test parsing simple operation
     """
--- a/gaphor/UML/umlfmt.py
+++ b/gaphor/UML/umlfmt.py
@@ -59,10 +59,8 @@ def format_attribute(
     will not give you the same result.
     """
     name = el.name
-    if not name:
-        return name
 
-    if no_render_pat.match(name):
+    if name and no_render_pat.match(name):
         return name
 
     # Render all fields if they all are set to False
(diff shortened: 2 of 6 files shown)


amolenaar:
The models shows all elements, which is something that needs fixing:

image

danyeaw:
I really like this approach. I think autoformatting the name using the umlfmt module and removing the icon were nice touches.

A few things I have been thinking about:
1. Should we have tools in the tool palette under the blocks section to create a Composition Relationship and Aggregation Relationship directly as a shortcut to creating an association, adding navigation, and setting shared or composite aggregation?
2. Should we automatically name the Property of these types of relationships? For example, I know some of the other tools name the property name the same as the class name, except as lower camel case instead of upper camel case like the class name.

Right now if you create an unnamed association it just shows up as a blank row in the namespace:
image

amolenaar:
> 1. Should we have tools in the tool palette under the blocks section to create a Composition Relationship and Aggregation Relationship directly as a shortcut to creating an association, adding navigation, and setting shared or composite aggregation?

This makes sense, I think, since now we need to head over to the editor to change the association ends.

> 2. Should we automatically name the Property of these types of relationships? For example, I know some of the other tools name the property name the same as the class name, except as lower camel case instead of upper camel case like the class name.

For composite and aggregate associations it makes sense, for sure. Since there already is this directional aspect to it, with navigability.

> Right now if you create an unnamed association it just shows up as a blank row in the namespace:

Hmm.. It should at least show something. Even `<None>` would be better than nothing.

amolenaar:
We still need to deal with relationships, too.

amolenaar:
I made attribute formatting a little more liberal, now it will render a string, even when there is no name for the property. Also it will parse associations with a `: Type` defined and just ignore that type (needed for editing in the tree view).

amolenaar:
Let's handle relationships in a different PR.

gaphor/UML/tests/test_umllex.py
```diff
@@ -145,6 +145,22 @@ def test_parse_association_end_derived_end(factory):
...
+    p = factory.create(UML.Property)
+    p.association = a
+
+    UML.parse(p, "end: TypeVal")
+    assert "end" == p.name
+    assert not p.typeValue
+    assert not p.lowerValue
+    assert not p.upperValue
+    assert not p.defaultValue
+    assert not p.typeValue
```
danyeaw, line 161: Is there supposed to be two assert statements for `p.typeValue`?

amolenaar: Fixed

Issue #339, opened 2020-06-02
Show Element Editor by default


amolenaar:
### I'm submitting a...
- [ ] Bug report  
- [X] Feature request
- [ ] Documentation issue or request

### Current behavior

The Element Editor is hidden by default. For new users it may not be obvious where to add attributes and operations to a class, for example. By showing it by default, we make that more explicit.

Also: I find myself opening the editor as one of the first things I do when editing a model.

### Expected behavior

Element editor is visible by default. If it's hidden, it will be hidden on next start of the application.

PR #344, opened 2020-06-03
Show element editor by default

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Issue Number: #339 

### What is the new behavior?

* Element editor is shown by default
* Extra text is shown with some usage tips
* The editor section is now a scrolled window
* Remove unused space on right side of window when element editor is collapsed
* Fixed box spacing for boxes created via the layout class

 
image

image

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/ui/elementeditor.glade
+++ b/gaphor/ui/elementeditor.glade
@@ -12,13 +12,13 @@
       <object class="GtkBox">
         <property name="visible">True</property>
         <property name="can_focus">False</property>
-        <property name="margin_right">2</property>
         <property name="margin_top">2</property>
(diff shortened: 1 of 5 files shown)


danyeaw:
Great idea to have some tips displayed! I think after I have seen this a couple of times that I would want to hide it. What if we added a checkbox under the No item selected font that by default was checked to Show Tips. Then if it is unchecked the tips aren't displayed?

amolenaar:
That's a nice addition. Having no tips will make the layout more clean again.

amolenaar:
image

danyeaw:
Awesome! 👍

PR #349, opened 2020-06-11
Proxy port

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Issue Number: #138 

### What is the new behavior?

A proxy port.

image

Still needs:

- [x] Connector to block (should also raise the port's z-index
- [x] Name
- [x] Stereotype
- [x] Orientation of text should change, depending on the side of the Block it's connected to


danyeaw:
Looks great so far!

danyeaw:
@amolenaar Is this ready for review / merge?

amolenaar:
Not yet. I want to fix text placement first.

amolenaar:
Port text is updated based on where the port is placed.

image

image

amolenaar:
@danyeaw  Ok, ready for review :smiley:

amolenaar:
Ah. Thanks! 
I only added a connector for Blocks so far. I'll add one for ref. and part properties too.
I did not add a connector for the (UML) Class yet, since that's out of scope for SysML support.

- [ ] Connect Proxy Port to Reference Property
- [ ] Connect Proxy Port to Part Property

amolenaar:
@danyeaw I need some help. A Proxy port is a subtype of Property. It can be linked to a Classifier. A Part or Reference Property is a property. I can not attach the port to the 

I figured that if you place a port `myPort` on a Classifier `myBlock, `myPort.encapsulatedClassifier = myBlock`.

It's unclear to me what relation should be created at model level when a Port is attached to a Part/Reference Property. Do you have an idea? Should it be done via a Connector internally?

danyeaw:
image
Looking at what one of the other tools I thought might give us some ideas.

You can create a part property without adding a type to it. Once you add a proxy port to it it automatically creates a block (block1 in this case), types the part property with the block, nests the proxy port under the block (makes block1 the owner), creates a new Property that is typed with the Block that the ibd is representing (called Model), and creates an association between the Model property and the part property.

PR #352, opened 2020-06-14
Add Part and Shared Association to Tool Palette

danyeaw:

This PR adds buttons to the tool palette for creating a Part (Direct Composition) and Shared Associations. This save a few step when create a Block Definition Diagram, so that you don't have to create an association and change the navigability and aggregation each time.

### PR Checklist

### PR Type

- [ ] Bugfix
- [X] Feature

Issue Number: Partial #138 

- [ ] Yes
- [X] No

### TODO
- [x] Change the navigability and aggregation of the association as it is created
- [x] Prevent removal of navigability and aggregation upon disconnect

### DIFF
--- a/gaphor/SysML/toolbox.py
+++ b/gaphor/SysML/toolbox.py
@@ -42,7 +42,7 @@ def create_association(
         assoc_item.head_end.subject.aggregation = "shared"
 
 
-def part_association_config(assoc_item: uml_items.AssociationItem) -> None:
+def composite_association_config(assoc_item: uml_items.AssociationItem) -> None:
     create_association(assoc_item, True)
 
 
@@ -128,14 +128,14 @@ sysml_toolbox_actions: ToolboxDefinition = (
                 handle_index=SE,
             ),
             ToolDef(
-                "toolbox-part-association",
-                gettext("Part Association"),
-                "gaphor-part-association-symbolic",
+                "toolbox-composite-association",
+                gettext("Composite Association"),
+                "gaphor-composite-association-symbolic",
                 "<Shift>Z",
                 PlacementTool.new_item_factory(
                     uml_items.AssociationItem,
                     UML.Association,
-                    config_func=part_association_config,
+                    config_func=composite_association_config,
                 ),
             ),
             ToolDef(
--- a/gaphor/SysML/toolbox.py
+++ b/gaphor/SysML/toolbox.py
@@ -32,13 +32,32 @@ def metaclass_config(new_item):
 
 
 def part_association_config(new_item):
-    new_item.head_end.aggregation = "composition"
-    new_item.head_end.navigability = True
+    p1 = new_item.model.create(UML.Property)
+    c1 = new_item.model.create(UML.Class)
+    p1.subject = c1
+    head_type = p1.subject
+    p2 = new_item.model.create(UML.Property)
+    c2 = new_item.model.create(UML.Class)
+    p2.subject = c2
+    tail_type = p2.subject
+    assoc = UML.model.create_association(head_type, tail_type)
+    assoc.package = new_item.canvas.diagram.namespace
+
+    new_item.head_end.subject = assoc.memberEnd[0]
+    new_item.tail_end.subject = assoc.memberEnd[1]
+
+    UML.model.set_navigability(assoc, new_item.head_end.subject, True)
+    new_item.head_end.subject.aggregation = "composite"
+
+    new_item.subject = assoc
 
 
 def shared_association_config(new_item):
-    new_item.head_end.aggregation = "shared"
-    new_item.head_end.navigability = True
+    prop = new_item.model.create(UML.Property)
+    prop.name = ""
+    new_item.head_end.subject = prop
+    print(prop.type)
+    UML.model.set_navigability(new_item.subject, new_item.head_end.subject, True)
 
 
 # Actions: ((section (name, label, icon_name, shortcut)), ...)
@@ -124,7 +143,9 @@ sysml_toolbox_actions: ToolboxDefinition = (
                 "gaphor-part-association-symbolic",
                 "<Shift>Z",
                 PlacementTool.new_item_factory(
-                    uml_items.AssociationItem, config_func=part_association_config
+                    uml_items.AssociationItem,
+                    UML.Association,
+                    config_func=part_association_config,
                 ),
             ),
             ToolDef(
@@ -133,7 +154,9 @@ sysml_toolbox_actions: ToolboxDefinition = (
                 "gaphor-shared-association-symbolic",
                 "<Shift>Q",
                 PlacementTool.new_item_factory(
-                    uml_items.AssociationItem, config_func=shared_association_config
+                    uml_items.AssociationItem,
+                    UML.Association,
+                    config_func=shared_association_config,
                 ),
             ),
             ToolDef(
--- a/gaphor/SysML/toolbox.py
+++ b/gaphor/SysML/toolbox.py
@@ -31,6 +31,16 @@ def metaclass_config(new_item):
     new_item.subject.name = "Class"
 
 
+def part_association_config(new_item):
+    new_item.head_end.aggregation = "composition"
+    new_item.head_end.navigability = True
+
+
+def shared_association_config(new_item):
+    new_item.head_end.aggregation = "shared"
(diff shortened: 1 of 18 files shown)


amolenaar:
I think in the icon, the diamond on the association end may be a little smaller.

Also I'd like this in the UML palette too.

danyeaw:
@amolenaar Thanks again for the feedback.:clap:  I made those two updates. What do you think of the updated icons?

amolenaar:
> What do you think of the updated icons?

Those look really nice :+1:

gaphor/SysML/toolbox.py
```diff
@@ -31,6 +25,31 @@ def metaclass_config(new_item):
     new_item.subject.name = "Class"
 
 
+def create_association(
+    assoc_item: uml_items.AssociationItem, part_association: bool
```
amolenaar, line 29: I would mot use a boolean here. Just the type of association as a str, or maybe Enum.

gaphor/SysML/toolbox.py
```diff
@@ -31,6 +25,31 @@ def metaclass_config(new_item):
...
+) -> None:
+    assoc = assoc_item.subject
+    assoc.memberEnd.append(assoc_item.model.create(UML.Property))
```
amolenaar, line 32: Just an assignment will work here as well. It’s a feature of the model.

gaphor/ui/icons/Makefile
```diff
@@ -15,6 +15,8 @@ ICONS=diagram \
 	package \
 	interface \
 	association \
+	part-association \
```
amolenaar, line 18: Should it. E named composite-association instead? It’s more general than parts now.

gaphor/SysML/toolbox.py
danyeaw: I was originally trying to do:
```
assoc.memberEnd = assoc_item.model.create(UML.Property)
assoc.memberEnd = assoc_item.model.create(UML.Property)
```
Unfortunately that causes mypy to detect that as an error:
```
gaphor/SysML/toolbox.py:35: error: Value of type "Property" is not indexable  [index]
gaphor/SysML/toolbox.py:36: error: Value of type "Property" is not indexable  [index]
```

gaphor/SysML/toolbox.py
danyeaw: Thanks added an Enum :+1:

gaphor/UML/classes/classconnect.py
```diff
@@ -152,24 +164,18 @@ def reconnect(self, handle, port):
...
+        """Disconnect model element.
+
+        Disconnect property (memberEnd) in case of end of life for connection.
         """
         opposite = self.line.opposite(handle)
         c1 = self.get_connected(handle)
         c2 = self.get_connected(opposite)
-        if c1 and c2:
-            old: UML.Association = self.line.subject
-            del self.line.subject
```
danyeaw, line 166: @amolenaar In order to keep the UML.Association including the navigability and aggregation, does this seem reasonable to you to not delete these on disconnect?

danyeaw:
@amolenaar I think this is ready, would you mind giving me your feedback on the two items above?

1. Use of append on the memberEnd to avoid mypy errors
2. Skipping deletion of the line.subject and head/tail subject on disconnect

Thanks!

gaphor/SysML/toolbox.py
amolenaar: Ah, then using `append()` is the better way. It's more explicit anyhow.

gaphor/UML/classes/classconnect.py
amolenaar: ~My original idea was to put that information on `AssociationItem.head_end` and `AssociationItem.tail_end`. These objects are created at construction time.~

~I missed that previously (was reviewing from my phone). Sorry 'bout that.~

EDIT: by looking at it for a second time I think this is much better, actually! An Association is also a Classifier, so it has a different role than, say, Generalization. Also, this will allow us to create Association Classes (or 3/4/5-way associations for that matter).

`len(self.line.subject.presentation) == 0`  could be rewritten as `not self.line.subject.presentation`.

gaphor/UML/classes/classconnect.py
```diff
@@ -152,24 +164,18 @@ def reconnect(self, handle, port):
...
-            del self.line.head_end.subject
-            del self.line.tail_end.subject
-            if old and len(old.presentation) == 0:
-                for e in list(old.memberEnd):
-                    e.unlink()
-                old.unlink()
+        if c1 and c2 and self.line.subject and len(self.line.subject.presentation) == 0:
```
amolenaar, line 175: `len(self.line.subject.presentation) == 0` could be rewritten as `not self.line.subject.presentation`.

gaphor/UML/classes/classconnect.py
amolenaar: This will, however, disconnect the member end from everything. Including the model. So these `memberEnd` properties can not be reused... Same goes for `self.line.subject.unlink()`: the Association instance will be completely removed from the model. This causes errors when saving and loading the model.

Since the `Association` and its `memberEnd`s are created when the item is placed on the diagram, they should remain part of the model after disconnecting. Alternatively, we can just create the Association and add the memberEnds only when an association is connected. It will violate the multiplicity constraint, though (`[2..*]`).

I think this approach makes sense. An Association is also a Classifier, so it has a different role than, say, Generalization. Also, this will allow us to create Association Classes (or 3/4/5-way associations for that matter).
The trade-off is we give up the possibility to present an association in different diagrams. We can always work around this by offering such an option though the element editor, which may be fine, since I generally to not tend to use associations in multiple diagrams anyway.

danyeaw: @amolenaar Thanks for the clarification on the `unlink`!

Ideally I think the associations should be reusable in many diagrams. A user should be able to grab a few elements in a model and then ask the model to display the relationships that already exist.

I thought we had to create the Association and memberEnds to display the navigability and aggregation. This is how it was a few commits ago, the association goes back to being a plain one as soon as you disconnect it.

amolenaar: > Ideally I think the associations should be reusable in many diagrams. A user should be able to grab a few elements in a model and then ask the model to display the relationships that already exist.

That would be a nice feature. But it's different from:

1. create a diagram with 2 classes and draw an association between them (name it "foo")
2. create a new diagram, add the same 2 classes and draw another association
3. Will the association be:
   a. a new association
   b. be the already existing association named "foo"?

Currently we do option 3a. Which is fine IMHO.

> I thought we had to create the Association and memberEnds to display the navigability and aggregation. This is how it was a few commits ago, the association goes back to being a plain one as soon as you disconnect it.

Indeed you'll need both Association and the Properties at the memberEnds. Since the logic on navigability lies at the model. Not in the diagram items. It should, however, still disconnect the "type"s at the end. In the connect() method, the association ends are paired with the classes they connect to. On disconnect, this relation should be removed I think. Otherwise the model and the diagrams are no longer in sync.

gaphor/UML/classes/classconnect.py
```diff
@@ -121,14 +123,24 @@ def connect_subject(self, handle):
...
                 ):
                     return
+                else:
+                    head_nav = end1.navigability
+                    head_agg = end1.aggregation
+                    tail_nav = end2.navigability
+                    tail_agg = end2.aggregation
 
             # Create new association
             relation = UML.model.create_association(head_type, tail_type)
```
amolenaar, line 133: Does this go well, or do we end up with stale associations in the model? If the association item already has an association connected it should not create a new association again.

danyeaw: @amolenaar I refactored this part to be more clear. I think the UML.Assocation is only created if the line doesn't have a subject already, for example by creating a plain association. What do you think?

amolenaar: This is better. There is less confusion on what's done and when. Well done :+1:

gaphor/UML/classes/classconnect.py
danyeaw: Thanks good catch. I don't think I was properly connecting the the type to the element either. How does this look?

danyeaw:
@amolenaar Anything left to do before I merge this?

PR #363, opened 2020-07-17
Part connector

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Issue Number: #138 

### What is the new behavior?

- [x] Allow (assembly) connections between Ports
- [x] Allow (delegation) connections between Part and Port
- [X] Parts can be nested in Blocks
- [x] Ports can be attached to Properties as well. (no `encapsulatedClassifier` is set, though!)
- [X] Fixed issues with item placement, now that `Canvas.reparent()` also updates the matrices
- [X] Fixed SysML property pages. The module was not loaded

NB. Somehow Part/Block connection via association does not work (has it ever?)

image


amolenaar:
@danyeaw Do we miss any behavior?

PR #371, opened 2020-07-23
Render property item with type and multiplicity

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Property item only has its name rendered

Issue Number: #138 

### What is the new behavior?

- Render contents of Property item with type and multiplicity, if applicable.
- Grouping for properties that are created as part of an association, can only group with their owner.

- [ ] Yes
- [X] No


danyeaw:
Looks great, nice addition!

PR #411, opened 2020-08-19
Fix issue where copy/paste + delete of original association causes association ends to be removed

amolenaar:
### PR Type

- [X] Bugfix
- [ ] Feature

Create a class, connect both ends of an association to it. Copy + paste the association and remove the original.

See `tests/test_multiple_associations.py` for a failing scenario.

### What is the new behavior?

I hope it will work as expected: association ends remain in tact as well as the navigability.

- [ ] Yes
- [X] No

### Notes

This is also applicable for Connectors!

Scenario's I can think of:

* Class and association are pasted in a new diagram
* Class and association are pasted in a new diagram and original association is deleted
* Class and association are pasted in a new diagram and new association is deleted
* Association is pasted in a new diagram and reconnected to the class (same subject as original)
* Association is pasted and directly deleted
* Class and association are pasted in a new diagram and one end is connected to a different class

What's the behavior we're looking for?

* If an association has >1 presentations, it can only connect to those same types ()
* If an association has ==1 presentations, it can reconnect to another type, association is updated

### DIFF
--- a/gaphor/UML/classes/tests/test_associationconnect.py
+++ b/gaphor/UML/classes/tests/test_associationconnect.py
@@ -115,7 +115,24 @@ def test_disconnect_of_second_association(connected_association, clone):
     assert new.subject is asc.subject
 
 
-# test allow connect on 1st association item, where one association is already connected (reconnect)
+def test_disconnect_of_navigable_end(connected_association):
+    asc, c1, c2 = connected_association
+
+    UML.model.set_navigability(asc.subject, asc.head_end.subject, True)
+
+    assert asc.head_end.subject in c2.subject.ownedAttribute
+
+    disconnect(asc, asc.head)
+
+    assert asc.subject
+    assert len(asc.subject.memberEnd) == 2
+    assert asc.subject.memberEnd[0].type is None
+    assert asc.head_end.subject not in c2.subject.ownedAttribute
+    assert asc.tail_end.subject not in c1.subject.ownedAttribute
+    assert asc.head_end.subject.type is None
+    assert asc.tail_end.subject.type is None
+
+
 def test_allow_reconnect(connected_association, create):
     asc, c1, c2 = connected_association
     c3 = create(ClassItem, UML.Class)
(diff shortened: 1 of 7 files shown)


gaphor/UML/classes/tests/test_associationconnect.py
```diff
@@ -0,0 +1,198 @@
...
+    assert c1 is not get_connected(asc, asc.head)
+    assert c2 is not get_connected(asc, asc.tail)
+
+    assert asc.subject
+    assert len(asc.subject.memberEnd) == 2
+    assert asc.subject.memberEnd[0].type is None
+    assert asc.subject.memberEnd[1].type is None
+
+
+# test disconnect with 2 associations -> model should trmain in tact
```
danyeaw, line 107: spelling of remain

tests/test_multiple_associations.py
```diff
@@ -0,0 +1,119 @@
...
+    # Now, when the original is deleted, the model is changed and made invalid
+
+    a.unlink()
+
+    assert aa.subject.memberEnd[0].type
+    assert aa.subject.memberEnd[1].type
+    assert aa.subject.memberEnd[0].type is c.subject
+    assert aa.subject.memberEnd[1].type is c.subject
+    assert aa.subject.memberEnd[0] is aa.head_end.subject
+    assert aa.subject.memberEnd[1] is aa.tail_end.subject
```
danyeaw, line 118: I'm sure this is my misunderstanding, if the original `a` is deleted, I would think that the copy of a called `aa` is also deleted. Why would these assert statements pass?

amolenaar: `a` is an `AssociationItem`, `aa` is also an `AssociationItem`. Both point to the same association (the `subect`). If I remove one, the model should not change, since there's still a presentation of the `Association` model element in a diagram.

gaphor/UML/classes/tests/test_associationconnect.py
amolenaar: I removed the comment and improved the test case name.

tests/test_multiple_associations.py
danyeaw: Thanks for the explanation 👍

PR #415, opened 2020-08-24
Live editing for inline editors (popovers)

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

I found it problematic that I had to explicitly "enter" a name in the popover editor. Even more so since the text in the sidebar editor for fields like "name" is instantly applied.

### What is the new behavior?

Text edited in a popover editor is instantly applied. One can hit "ESC" to revert the text to before the change.

- [ ] Yes
- [X] No

### Other information

- [ ] Edit sessions should take place in one (1) transaction


danyeaw:
@amolenaar Nice idea, I found this annoying as well :+1:

PR #430, opened 2020-09-08
Improve Swimlane Behavior

### LINKED ISSUE #426: Improve Activity Partition / Swimlane Support
The current swimlane implementation needs an update:
- You create a partition
- You then have to create another partition inside the partition to name the partition
- There is no support for allocating a Block or Class to the swimlane

**Describe the solution you'd like**
Opinions that I think will simplify the design:
- A single partition isn't usable for anything. If there isn't a partition on the canvas, creating a partition should create two.
- Nesting partitions inside other partitions isn't needed for any practical reasons.
- There is no reason to have two different sets of swimlanes on the same diagram

Goals:
- No right click menu needed. We don't have a lot of right click options as it is, and I think adding right click to partitions wouldn't be very discoverable.

Proposed design:
- Creating a partition will create two swimlanes, both without a label, but with an inside line for where the allocation goes
- Using the partition tool from then on will allow you to snap another partition to the left or right of the current ones
- Dragging from the top of the partition will allow you to reorder partitions
- To make a name appear at the top of the swimlane, drag a block or class to the swimlane. I think the whole swimlane can be a drop zone
- There will be handles at the 4 outside corners of the overall group of swimlanes
- Resizing of the width happens by dragging the line between the swimlanes

danyeaw:
### PR Checklist

### PR Type

- [ ] Bugfix
- [X] Feature

Creation of partitions requires dragging PartitionItems on other PartitionItems.

Issue Number: Partially addresses #426 

### What is the new behavior?
Two partitions are created by default when a PartitionItem is created. Adding and removing partitions is handled by the partition property page editor.

- [ ] Yes
- [X] No

### Future PRs
- Allow Activities and ActivityPartitions to be deleted from the namespace treeview.
- Allow blocks / classes to be dragged on a swimlane to allocate behavior (and name the partition)


danyeaw:
I am working on a storage module upgrade for PartitionItems, and then I think this will be ready.

gaphor/UML/actions/partition.py
```diff
@@ -1,154 +1,139 @@
...
-
-        self._toplevel = self.canvas.get_parent(self) is None
-        self._subpart = len(children) > 0
-        self._bottom = not (self._toplevel or self._subpart)
-
-        if self._toplevel:
-            self._header_size = self._header_size[0], self.DELTA
-
-        handles = self.handles()
+    num_partitions: attribute[int] = attribute(
```
amolenaar, line 38: With the partition association at hand, we should be able to get rid of this counter.

gaphor/UML/actions/partition.py
```diff
@@ -1,154 +1,139 @@
...
+        for partition in self.partition:
+            if (
+                not partition.activity
+                and activity.ownedClassifier
+                and isinstance(activity.ownedClassifier[0], UML.Activity)
+            ):
+                partition.activity = activity.ownedClassifier[0]
+
+    def update_partitions(self) -> None:
+        """Add and remove UML.ActivityPartitions."""
```
amolenaar, line 64: I think this logic can move to partitionPage.

gaphor/storage/storage.py
```diff
@@ -414,3 +418,21 @@ def upgrade_element_to_2_1_0(elem):
...
+            yield item
+
+
+def upgrade_canvas_item_to_2_1_0(item: gaphas.Item) -> Optional[gaphas.Item]:
+    if item.type == "PartitionItem" and item.canvasitems:
+        item.canvasitems = list(flatten(item.canvasitems))
+        item.canvasitems = [
+            item for _ in item.canvasitems if item.type != "PartitionItem"
+        ]
+    return item
```
amolenaar, line 438: Is this logic sound? This may result in
`item.canvasitems = [item]`, which causes a circular reference. I suppose what we want is filter out the nested PartitionItems?

amolenaar:
Just loading the old model looks pretty okay, IMHO.

image

gaphor/UML/actions/partition.py
danyeaw: Ok, I'll look in to refactoring this to the partitionpage, thanks!

gaphor/storage/storage.py
danyeaw: This is no longer needed because gaphor can now still load old models.

amolenaar (approved):

PR #431, opened 2020-09-09
Cleanup session lifecycle

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature

There was a lot of hopping around the code when dealing with file loading and session creation. This was only possible through the `GtkApplication`. This is inconvenient and not handy, since `GtkApplication` is not out core application: `gaphor.application.Application` is.

Issue Number: N/A

### What is the new behavior?

This PR cleans up the code. Makes sure all file loading takes place through the `app_file_manager` and ensures session lifecycle can be completely handled in our code.

On application level there's also an event manager now, so session creation and application shutdown events can be propagated to the GtkApplication as well. Handlers in GtkApplication ensure that the main window is shown when a new session is created and make sure the Gtk main loop is ended when the Gaphor Application ends.

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/application.py
+++ b/gaphor/application.py
@@ -80,7 +80,7 @@ class Application(Service, ActionProvider):
     def active_session(self):
         return self._active_session
 
-    def new_session(self, services=None, with_ui=False):
+    def new_session(self, services=None):
         """
         Initialize an application session.
         """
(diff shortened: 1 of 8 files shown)


danyeaw (approved):
Nice cleanup!

PR #446, opened 2020-09-26
Update pre-commit hook versions

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Code style update (formatting, local variables)

Dependabot updates versions in `pyproject.toml`, but not in our `.pre-commit-config.yaml`.

### What is the new behavior?

The version of pre-commit hooks is updated to the versions defined in `pyproject.toml`.

All code has been reformatted accordingly.

- [ ] Yes
- [X] No


danyeaw (approved):
I wonder if we can get Dependabot to update pre-commit as well :thinking:

amolenaar:
I was looking into using the local packages instead, so we only have to update versions in pyproject.toml: 2b9813f94acf2da2e3d1add617be41f28960f971.

This would work, if we could install the dev-dependencies only for the linting job. There is an issue for that https://github.com/python-poetry/poetry/issues/2572.

PR #449, opened 2020-09-27
Fix is directory error with importlib resources in Python 3.9

danyeaw:

Python 3.9 is right around the corner, with release candidates out now. One of the changes is that importlib.resources now has a `files` method that returns a path and you can feed it directories. Some of the methods that did exist like `importlib.resources.path` were only supposed to be used for resources, which don't include directories. The new method is nice, because it doesn't force using a context manager to get a path to a file or directory.

This PR fixes a IsADirectoryError with Python 3.9 since it no longer allows calling directories with `importlib.resources.path`.

### PR Checklist

### PR Type

- [X] Bugfix
- [ ] Feature

Issue Number: N/A

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/ui/__init__.py
+++ b/gaphor/ui/__init__.py
@@ -4,6 +4,7 @@ and diagram windows."""
 import importlib.resources
 import logging
 import sys
+from pathlib import Path
 from typing import Optional
 
 import gi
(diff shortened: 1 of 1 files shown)


amolenaar (approved):
Excellent :)

Should we also add Py39 to the build?

Issue #373, opened 2020-07-26
Class / Block (from Package Name) Doesn't Update When Moving a Diagram


danyeaw:
### I'm submitting a...
- [X] Bug report  
- [ ] Feature request
- [ ] Documentation issue or request

### Current behavior
Changing which package a diagram is in doesn't cause the diagram to refresh the display of the Class or Block to show that they are from a different package.

### Expected behavior
The elements in the diagram update immediately.

### Minimal reproduction of the problem with instructions
Create a diagram with a Block. Create a package, and move the diagram in to the package. The Block doesn't update in the diagram to show that it is (from New model package).

### System Information
OS:
- [X] Linux (Please put in notes the specific distro)
- [ ] MacOS
- [ ] Windows 10
- [ ] Windows 7

NOTES:

amolenaar:
This is a tricky one. Diagram items do not reference the diagram directly. The change is not caused by a change in the "subject", but in the diagram.

danyeaw:
Alternatively, instead of showing that an element is part of a different package in the diagram, we could add a button to select in the treeview the item that is selected in the diagram. I think this would be useful anyway for other reasons, but as long as you can easily see where an item came from, I think showing it all the time isn't really needed. It often clutters the diagrams.

amolenaar:
You mean removing the "from" line? I agree that it would be useful to find something in the tree view from the diagram. Either by a button (IntelliJ style) or instantly (VS Code style).

PR #465, opened 2020-09-28
Update diagram when moved

amolenaar:
### PR Type

- Bugfix

When a diagram is moved, "from" lines remain, although they may be invalid.

Issue Number: #373

### What is the new behavior?

Diagram is updated when moved. Relative references are updated, as are "from" messages in classes and packages.

- [ ] Yes
- [X] No


danyeaw (approved):
You are on :fire:, another nice fix!

PR #463, opened 2020-09-28
Fix copy/paste for Transition

amolenaar:
### PR Type
What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Documentation content changes

### What is the current behavior?

Transition items are copied without a guard clause.

Issue Number: N/A

### What is the new behavior?

The `guard` Constraint is also copied.

### Does this PR introduce a breaking change?
- [ ] Yes
- [X] No

PR #466, opened 2020-09-28
Use a global copy buffer

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

When you copy some items in a diagram, you can only paste them in the same model.

Issue Number: #463 

### What is the new behavior?

You can copy, and then paste the items in a different diagram. The diagram items are backed by their model elements.

- [ ] Yes
- [X] No

### Other information

The current implementation (a global variable) is not really nice. But it proves the application can handle this copy/paste behavior. The actual behavior should work via an `gaphor.appservice`. However, there's no way to inject an app service in a session. And I'm not sure if we want to.

### DIFF
--- a/gaphor/services/copyservice.py
+++ b/gaphor/services/copyservice.py
@@ -10,6 +10,8 @@ from gaphor.core.modeling import Presentation
 from gaphor.diagram.copypaste import copy, paste
 from gaphor.ui.event import DiagramSelectionChanged
 
+copy_buffer: object = None
+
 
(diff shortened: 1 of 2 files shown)


amolenaar:
I want to leave it as it is, with the global copy variable. Alternative would be to use a application service instead. But we'd better improve this bit when (if) we migrate to GTK4.

danyeaw (approved):
Cool, I agree that this will work until we migrate to GTK4 eventually :+1:

Issue #387, opened 2020-08-07
Show descriptive error messages when a model fails to load


amolenaar:
### I'm submitting a...
- [ ] Bug report  
- [X] Feature request
- [ ] Documentation issue or request

### Current behavior

Currently, when a model fails to load a user is presented with something that looks like a stack trace.

We should do better than that

### Expected behavior

Provide a descriptive error message if:

1. A file is not a gaphor model at all (e.i. does not start with a `<gaphor>` tag)
2. It appears to be a Gaphor file, but the contents is corrupt

### Minimal reproduction of the problem with instructions

Load anything but a valid Gaphor XML file.

See also #384

amolenaar:
A message should look something like this, I think. Taking into account the latest HIG guidelines.

amolenaar:
This is what I made of it:
![image](https://user-images.githubusercontent.com/96249/94610286-5b04db80-02a0-11eb-8840-ed4a1bac71fe.png)

PR #469, opened 2020-09-29
Improve file load/save dialogs

amolenaar:
### PR Type

- [X] Bugfix
- [ ] Feature

Error messages used to be technical and hard to understand for non-technical users.

Issue Number: #387 

### What is the new behavior?

Human readable error dialog when a model does not load:

image

Or saved:

image

When overwriting a file:

image

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/ui/filemanager.py
+++ b/gaphor/ui/filemanager.py
@@ -1,6 +1,7 @@
 """The file service is responsible for loading and saving the user data."""
 
 import logging
+import pathlib
 
 from gi.repository import Gtk
 
@@ -238,11 +239,33 @@ class FileManager(Service, ActionProvider):
             gettext("Save Gaphor Model As"), action="save", filename=self.filename
         )
 
-        filename = file_dialog.selection
+        filename = self.verify_filename(file_dialog.selection)
 
         file_dialog.destroy()
 
         if filename:
+            file_path = pathlib.Path(filename)
+            if file_path.exists():
+                dialog = Gtk.MessageDialog(
+                    None,
+                    Gtk.DialogFlags.MODAL | Gtk.DialogFlags.DESTROY_WITH_PARENT,
+                    Gtk.MessageType.QUESTION,
+                )
+                dialog.props.text = gettext(
+                    "A file name “{filename}” already exists. Do you want to replace it?"
+                ).format(filename=file_path.name)
+                dialog.props.secondary_text = gettext(
+                    "The file already exists in “{dirname}”. Replacing it will overwrite its contents."
+                ).format(dirname=file_path.parent.name)
+                dialog.add_buttons(Gtk.STOCK_CANCEL, 1, gettext("Replace"), 0)
+
+                dialog.set_transient_for(self.main_window.window)
+                cancel = dialog.run()
+                dialog.destroy()
+
+                if cancel:
+                    return
+
             self.save(filename)
             return True
 
--- a/gaphor/ui/filemanager.py
+++ b/gaphor/ui/filemanager.py
@@ -22,6 +22,24 @@ MAX_RECENT = 10
 log = logging.getLogger(__name__)
 
 
+def error_message(e):
+    if not isinstance(e, IOError):
+        return gettext(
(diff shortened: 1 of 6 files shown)


gaphor/ui/filemanager.py
```diff
@@ -177,11 +180,14 @@ def save(self, filename):
...
             self.event_manager.handle(FileSaved(self, filename))
-        except Exception:
+        except Exception as e:
             error_handler(
-                message=gettext("Error while saving model to file {filename}").format(
+                message=gettext("Unable to save model “{filename}”.").format(
                     filename=filename
                 ),
+                secondary_message=gettext(
+                    "The model can not be stored at this location:\n{exc}"
```
danyeaw, line 189: In the secondary messages, could we instead focus on the ways the user might be able to remedy the problem? For example: Please try to save a different location.

gaphor/ui/filemanager.py
```diff
@@ -177,11 +180,14 @@ def save(self, filename):
 
             self.filename = filename
             self.event_manager.handle(FileSaved(self, filename))
-        except Exception:
+        except Exception as e:
             error_handler(
-                message=gettext("Error while saving model to file {filename}").format(
+                message=gettext("Unable to save model “{filename}”.").format(
```
danyeaw, line 185: Should we decode the filename to remove the `b' '` wrapping?

gaphor/ui/filemanager.py
```diff
@@ -94,9 +94,12 @@ def load(self, filename):
             self.event_manager.handle(FileLoaded(self, filename))
         except Exception:
             error_handler(
-                message=gettext(
-                    "Error while loading model from file {filename}"
-                ).format(filename=filename),
+                message=gettext("Unable to open model “{filename}”.").format(
```
danyeaw, line 97: Can we use f-strings with the gettext calls?

danyeaw (commented):
Great idea to improve the error messages :+1:

gaphor/ui/filemanager.py
amolenaar: It looks fine with me (see screenshot above).

gaphor/ui/filemanager.py
amolenaar: No. Because f-strings will substitute in place, hence, the gettext text can not be resolved.

E.g. I have a text `"hello {name}"`. When I do `name = "Arjan"; gettext("hello {name}")`, gettext will see the text `"hello Arjan"`, while it should resolve `"hello {name}"` and after that replace `{name}` with `Arjan`.

amolenaar:
I added a dialog for file overwrites.

amolenaar:
I want to move the "replace file" logic into the FileDialog class.

- [ ] Move replace file check in FileDialog
- [ ] Show dialog when replace question is asked
- [ ] Check export functions to see if they work as expected (export diagram and export XMI)

gaphor/ui/filemanager.py
danyeaw: Thanks for the explanation! :smile:

gaphor/ui/filemanager.py
danyeaw: It looks good now, thanks!

gaphor/ui/filemanager.py
```diff
@@ -22,6 +22,24 @@
...
+    if e.errno == 13:
+        return gettext(
+            "You do not have the permissions necessary to save the model.\nPlease check that you typed the location correctly and try again."
+        )
+    elif e.errno == 28:
+        return gettext(
+            "You do not have enough free space on the device to save the model.\nPlease free up some disk space and try again or save it in a different location."
+        )
+    return gettext(
+        "The model can not be stored at this location:\n{exc}\nPlease check that you typed the location correctly and try again."
```
danyeaw, line 39: This should be "cannot"

danyeaw (approved):
These error messages look great! One small nit on spelling.

gaphor/ui/filemanager.py
amolenaar: Thanks

Issue #386, opened 2020-08-07
Show relationships in tree view


amolenaar:
### I'm submitting a...
- [ ] Bug report  
- [X] Feature request
- [ ] Documentation issue or request

### Current behavior

Relations are missing from the tree view. This is also the cause of issue #275.

### Expected behavior

Show relationships in the diagram.

A simple version can *just show* all NamedElements. Note that there are also NamedElements that are not assigned a namespace.

Next, items can be categorized. E.g. relationships can be grouped under a "Relationships" node.

Actions:
 
- [X] update the UML model to UML 2.5.1 layout to ease editing. 
- [ ] Ensure all elements are created with an owner
- [ ] Allow all elements to be dragged on a diagram and provide sensible behavior for relationships.
- [ ] Provide all elements with a default owner and add migrations to the storage module to retrofit existing diagrams.
- [X] Group relationships in a "pseudo-package" in the tree view. for readability.
- [ ] Provide useful names for relationships (e.g. many associations do not have a name) -> can be done by adding clauses to the `format()` generic function.

amolenaar:
I found the tree view is to limited, since it only takes into account Namespace relationships.

To obtain all ownership relationships, we should check for `Element.owner` (opposite is `Element.ownedElement`). This shows all elements and their relationships.

However, these relationships are not always filled in correctly in the UML model (which is now a mix of UML 2.2 and UML 2.5) We should move to UML 2.5, but for that to work all associations should be visible, so they can be migrated.

So I should:

1. Make associations visible, so I can update the relevant parts to UML 2.5.1.
2. Ensure all items have the appropriate ownership relation (e.g. when loading a model, when adding elements)
3. Structure the model in such a way that relationships appear under their own special subnode.

amolenaar:
I'm trying to show all elements in the namespace/tree view. To view the elements the tree view tracks the `Element.owner` - `Element.ownedElement` relationship. 

However, there are a whole bunch of elements (e.g. Action, LifeLine) that an not have a Package as their owner. Now these elements are "dropped" in the root of the tree view. E.g. a `State` Element is supposed to be placed in a `StateMachine`.

I'm looking into what to do with those elements:
1. Hide them? 
    * Would that mean we keep a blacklist/whitelist of items to track?
    * That's a hassle when new modeling languages are added to Gaphor. I like to be able to integrate those without a need to change the tree view.
2. For all elements that can not be owned by a Package, add an association between the element and a Package?
[...]

PR #478, opened 2020-10-04
In-app notification

amolenaar:
### PR Type

- [X] Feature

DnD to the diagram is not supported for all element. A message is only shown in the console, not to the user.

Issue Number: #386 

### What is the new behavior?

When you drag something like an Action to the diagram, a notification is shown that this behavior is not yet supported.

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/ui/diagrampage.py
+++ b/gaphor/ui/diagrampage.py
@@ -30,7 +30,7 @@ from gaphor.diagram.painter import ItemPainter
 from gaphor.diagram.support import get_diagram_item
 from gaphor.transaction import Transaction
 from gaphor.ui.actiongroup import create_action_group
-from gaphor.ui.event import DiagramSelectionChanged
+from gaphor.ui.event import DiagramSelectionChanged, Notification
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


gaphor/ui/diagrampage.py
```diff
@@ -370,8 +370,12 @@ def _on_drag_data_received(self, view, context, x, y, data, info, time):
             if not isinstance(
                 element, (UML.Classifier, UML.Package, UML.Property)
             ) or isinstance(element, UML.Association):
-                log.warning(
-                    f"DnD is (temporarily) limited to Classifiers, Packages and Properties, not {type(element).__name__}"
+                self.event_manager.handle(
+                    Notification(
+                        gettext(
+                            "Drag to diagram is (temporarily) limited to Classifiers, Packages and Properties, not {type}."
```
danyeaw, line 376: Sorry for the nitpicks: could we use an oxford comma? Ideally it would read better if we could made the type plural, but I think it would at least be better if we added an indefinite article "a" before the type.

danyeaw (approved):
I really like the in-app notifications, it makes Gaphor look really professional to give the user a warning using this!

PR #502, opened 2020-10-26
Fix macos build

amolenaar:
### PR Type

- [X] Bugfix

The build breaks: Homebrew has updated our dependency (`gobject-introspection`) to depend on Python 3.9

### What is the new behavior?

The Python version is determined on the Python used by `gobject-introspection`. This way the build should no longer break on Python updates.

- [ ] Yes
- [X] No


danyeaw (approved):
Nice @amolenaar !

PR #530, opened 2020-11-30
Fix copy paste

amolenaar:
### PR Type

- [X] Bugfix
- [ ] Feature

Copy/paste breaks on Ubuntu 20.10/Wayland.

### What is the new behavior?

Copy buffer is only cleared if a view has no focus.

- [ ] Yes
- [X] No

### Other information

It looks like I get multiple clipboard `OwnerChange` events in Wayland. This broke the logic with semaphores.

Oddly enough it worked fine in VSCode (installed as Snap).


danyeaw (approved):
Looks good!

Issue #468, opened 2020-09-29
Make diagram items 100% Gaphor model elements


amolenaar:
This is an idea on the inner workings of Gaphor's data model and diagrams in particular

> I use the term "model element" for any `gaphor.core.modeling.Element` subclass that represents items in a UML, SysML, etc. model. "Diagram items" are the visible elements in a diagram, backed by a model element.

## How it came to be

* Diagram items used to be simple `gaphas.Item` instances. Over the course of the 2.0 release, diagram items also inherited from `UML.Presentation`, now `core.modeling.Presentation`.
* Diagrams are stored in a hierarchical way in the model files. There's special logic to deal with diagram items

## What I want

At this point diagram items are 50% Gaphas items and 50% Gaphor model elements. I think it would be a huge improvement (read: simplification) if diagram items become 100% model element.

We do not make use of the properties of `Element`. such as the `owner` union relation. Wouldn't it be nice if the graphical items are just simple model elements? Changes to diagram items are handled the same way as elements in the model. No glue code, nothing special. All elements and diagram items will be managed by the `ElementFactory`. They can be queried for, etc.

Diagrams have a hierarchy: a diagram contains items, and those items can contain other items. This is much like the `Element.owner` - `Element.ownedElement` relationship in the data model. I think we should move towards this model.

## What it needs

Most of the complexity is in the intersection of Gaphor and Gaphas (the canvas library). The big challenge will be to make Gaphas extensible to a level we can use the Gaphor model elements and not rely on 

## What we gain

There is no need for the global (undo) event handler in Gaphas. No special cases for Gaphas items.
Items can be broken down in building blocks and those building blocks can be used to create a canvas representation based on our model.

# To do

There are a couple of improvements that can be done to the current code base:

- [X] Reference `diagram` instead of `diagram.canvas`. #565
- [x] Remove `canvas` from `Presentation`: only use `diagram`
- [x] Make `Presentation.diagram` a normal association (opposite: `Diagram.presentation`).
- [x] Add association `Presentation.children` <---<> `Presentation.parent` and use it instead of Gaphas' tree structure
- [x] Use Core Modeling attributes and associations wherever possible. 
- [ ] Storage package should no longer check for `Canvas` -> requires changes to the storage format?

amolenaar:
[...]

PR #565, opened 2020-12-22
Use our Diagram instead of Gaphas Canvas

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature

We have `Diagram`, but a lot of diagram logic is delegated to a Gaphas `Canvas` instance.

The idea is to bring diagram items' behavior closer to normal model elements. That should make it easier to work with them.

Issue: #468 

### What is the new behavior?

`Diagram` is used directly. It can be passed to a GtkView for rendering. It's used to deal with connections and other stuff that was previously the task of `diagram.canvas`.

This is a first step to make diagram items behave like "normal" model elements. Diagram items already inherit all desired behavior from `Presentation`, but a lot of what we need can not be used due to the dependency on Gaphas.

- [X] Yes
- [ ] No

`diagram.canvas` is no longer there. Use `diagram` instead.

### DIFF
--- a/gaphor/SysML/blocks/connectors.py
+++ b/gaphor/SysML/blocks/connectors.py
@@ -19,7 +19,7 @@ class BlockProperyProxyPortConnector:
         block: Union[BlockItem, PropertyItem],
         proxy_port: ProxyPortItem,
     ) -> None:
-        assert block.canvas is proxy_port.canvas
+        assert block.diagram is proxy_port.diagram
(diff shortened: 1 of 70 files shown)


danyeaw (approved):
I really like this simplification, nice job @amolenaar!

amolenaar:
The changes are part of #567 as well. Closing this PR.

PR #567, opened 2020-12-22
Use associations for diagram and parent/child relationships

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature

We do special things for Presentation relationships. These should be ordinary relationships, as defined in the core model.

Issue Number: #468

### What is the new behavior?

* `Diagram` now owns all presentations (`Diagram.ownedPresentation`).
* `Presentation.diagram` is a normal association. We can watch it. No need for canvas variables anymore.
* Removed `DiagramCanvas`. Functionality is merged with `Diagram`
* `Presentation` elements use a `parent`-`children` relationship. The tree structure is no longer centrally managed by the Canvas.
* Watched paths are directly registered. `EventWatcher.subscribe_all()` has be removed.
* Removed `Presentation.setup_canvas()` and `teardown_canvas()`. We no longer need them.

- [X] Yes
- [ ] No

### DIFF
--- a/gaphor/SysML/blocks/connectors.py
+++ b/gaphor/SysML/blocks/connectors.py
@@ -19,7 +19,7 @@ class BlockProperyProxyPortConnector:
         block: Union[BlockItem, PropertyItem],
         proxy_port: ProxyPortItem,
     ) -> None:
-        assert block.canvas is proxy_port.canvas
+        assert block.diagram is proxy_port.diagram
         self.block = block
         self.proxy_port = proxy_port
 
@@ -70,10 +70,7 @@ class PropertyConnectorConnector(UnaryRelationshipConnect):
         return super().allow(handle, port)
 
     def connect_subject(self, handle):
-        element = self.element
         line = self.line
-
-        assert element.canvas
 
         c1 = self.get_connected(line.head)
         c2 = self.get_connected(line.tail)
--- a/gaphor/SysML/blocks/connectors.py
+++ b/gaphor/SysML/blocks/connectors.py
@@ -38,8 +38,8 @@ class BlockProperyProxyPortConnector:
             proxy_port.subject.encapsulatedClassifier = self.block.subject
 
         # This raises the item in the item hierarchy
-        assert proxy_port.canvas
-        proxy_port.canvas.reparent(proxy_port, self.block)
+        assert proxy_port.diagram
+        proxy_port.diagram.reparent(proxy_port, self.block)
 
         return True
 
@@ -48,7 +48,7 @@ class BlockProperyProxyPortConnector:
         if proxy_port.subject and proxy_port.canvas:
             subject = proxy_port.subject
             del proxy_port.subject
-            proxy_port.canvas.reparent(proxy_port, None)
+            proxy_port.diagram.reparent(proxy_port, None)
             subject.unlink()
 
 
--- a/gaphor/SysML/blocks/connectors.py
+++ b/gaphor/SysML/blocks/connectors.py
@@ -45,7 +45,7 @@ class BlockProperyProxyPortConnector:
 
     def disconnect(self, handle: Handle) -> None:
         proxy_port = self.proxy_port
-        if proxy_port.subject and proxy_port.canvas:
+        if proxy_port.subject and proxy_port.diagram:
             subject = proxy_port.subject
             del proxy_port.subject
             proxy_port.diagram.reparent(proxy_port, None)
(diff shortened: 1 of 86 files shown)


amolenaar:
@danyeaw I did a little more cleanup. Presentation types start a lot to look like ordinary model elements :smiley:.

gaphor/storage/storage.py
```diff
@@ -67,8 +67,7 @@ def save_element(name, value, writer):
 
     A value may be a primitive (string, int), a
     gaphor.core.modeling.collection (which contains a list of references
-    to other UML elements) or a gaphas.Canvas (which contains canvas
```
danyeaw, line 70: With this change does the concept of a `canvas item` still exist in Gaphor?

gaphor/storage/storage.py
```diff
@@ -112,11 +111,11 @@ def save_canvas(value):
 
         The extra attribute reference can be used to force UML
         """
-        assert isinstance(value, gaphas.Item)
```
danyeaw, line 115: Would it make more sense now to rename this function and the one below it to `save_diagram` and `save_diagram_item` now?

gaphor/storage/storage.py
```diff
@@ -188,7 +186,7 @@ def _load_elements_and_canvasitems(
     elements, factory, modeling_language, gaphor_version, update_status_queue
 ):
     def create_canvasitems(diagram, canvasitems, parent=None):
-        """Canvas is a read gaphas.Canvas, items is a list of
```
danyeaw, line 191: Same thing here as well, would now be a good time to rename this function from `canvasitems` to `diagramitems`?

gaphor/SysML/blocks/connectors.py
```diff
@@ -38,17 +38,17 @@ def connect(self, handle: Handle, port: Port) -> bool:
...
         return True
 
     def disconnect(self, handle: Handle) -> None:
         proxy_port = self.proxy_port
-        if proxy_port.subject and proxy_port.canvas:
+        if proxy_port.subject and proxy_port.diagram:
             subject = proxy_port.subject
             del proxy_port.subject
-            proxy_port.canvas.reparent(proxy_port, None)
+            proxy_port.parent = None  # type: ignore[assignment]
```
danyeaw, line 51: I really like the simplifications here and throughout the refactoring to be able to call `.parent` :+1:

gaphor/SysML/propertypages.py
```diff
@@ -50,7 +50,7 @@ def text_handler(event):
                 buffer.set_text(event.new_value)
                 buffer.handler_unblock(changed_id)
 
-        self.watcher.watch("text", text_handler).subscribe_all()
```
danyeaw, line 53: I like that we don't have to explicitly subscribe now :+1:

gaphor/UML/classes/interface.py
```diff
@@ -158,8 +158,8 @@ class InterfaceItem(ElementPresentation, Classified):
     RADIUS_REQUIRED = 14
 
     def __init__(self, connections, id=None, model=None):
-        super().__init__(connections, id, model)
         self._folded = Folded.NONE
+        super().__init__(connections, id, model)
```
danyeaw, line 162: Did you mean to swap the super class constructor so that it is below the assignment here?

danyeaw (approved):
Awesome refactoring as always! :100:

gaphor/SysML/propertypages.py
amolenaar: In the old setup (Gaphas 2) we needed to. Now we do no longer have to. :smiley:

gaphor/UML/classes/interface.py
amolenaar: Yes, it's on purpose. `update_shapes()` is called from `ElementPresentation`'s constructor, and it requires this attribute.

gaphor/storage/storage.py
amolenaar: Done

gaphor/storage/storage.py
amolenaar: Should be updated

gaphor/storage/storage.py
amolenaar: Updated.

gaphor/UML/classes/interface.py
danyeaw: Ahh got it, i knew there had to be a reason for it, but it wasn't immediately clear

danyeaw:
Nice job @amolenaar!

PR #572, opened 2020-12-26
Presentation requires diagram

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, ~no~ api changes)

It's a bit awkward that a `Presentation` takes a Gaphas construct (`Connections`) as one of it's arguments.
A presentation item is tied to a diagram. They have their connections (constraints) centrally managed.
It's not explicit that a `Presentation` is tied to a `Diagram`.

### What is the new behavior?
`Presentation` items take a `Diagram` as their first argument. This way the code is telling us that a `Presentation` item can not exist without a diagram. Also, the presentation is instantly unlinked when the diagram reference changes. This is desired, since we need to clean up constraints and connections made on the diagram (those can not be easily transferred to another diagram).

Because `Presentation` takes a `diagram` argument now, there's no longer need for a `model` to be provided. The `Presentation` class handles the "translation" to normal `Element`s.

`Presentation` also inherits the `gaphas.item.Matrices` mixin. It provides the `matrix` and `matrix_i2c` fields we require anyway.

- [X] Yes
- [ ] No

### Other information

In the cases where we also inherit from Gaphas' items (`Element`, `Line`) the constructor takes an extra `connections` argument, consumed by the Gaphas item. Any kw-args are passed on to the next (super) class in the MRO sequence.

So in those cases the constructor make look a bit funny:

```python
class ElementPresentation(gaphas.Element, Presentation[S]):
    def __init__(self, diagram: Diagram, id=None, shape=None):
        super().__init__(connections=diagram.connections, diagram=diagram, id=id)  # type: ignore[call-arg]
        ...
```

I think this is acceptable, since we're still able to call super-classes via `super()` and we need to "work around" the different method signatures.

### DIFF
--- a/gaphor/diagram/diagramtools/dropzone.py
+++ b/gaphor/diagram/diagramtools/dropzone.py
@@ -1,5 +1,6 @@
+from typing import Type
+
 from gaphas.aspect.move import Move as MoveAspect
-from gaphas.connections import Connections
 from gaphas.guide import GuidedItemMove
 from gaphas.tool.itemtool import item_at_point
 from gaphas.view import GtkView
@@ -13,13 +14,15 @@ from gaphor.diagram.presentation import (
 )
 
 
-def drop_zone_tool(view, item_class):
+def drop_zone_tool(
+    view: GtkView, item_class: Type[Presentation]
+) -> Gtk.EventController:
     ctrl = Gtk.EventControllerMotion.new(view)
     ctrl.connect("motion", on_motion, item_class)
     return ctrl
 
 
-def on_motion(controller, x, y, item_class):
+def on_motion(controller, x, y, item_class: Type[Presentation]):
     view: GtkView = controller.get_widget()
     model = view.model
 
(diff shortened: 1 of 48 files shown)


gaphor/diagram/diagramtools/dropzone.py
```diff
@@ -29,13 +32,10 @@ def on_motion(controller, x, y, item_class):
...
-        # create dummy adapter
-        connections = Connections()
-        adapter = Group(parent, item_class(connections))
-        if adapter and adapter.can_contain():
-            view.selection.dropzone_item = parent
-        else:
-            view.selection.dropzone_item = None
+        adapter_type = Group.registry.get_registration(type(parent), item_class)
+        # Do not do in depth check id we can actually connect.
+        # We can do that only when we have and item_class instance.
```
danyeaw, line 37: This comment isn't very clear due to grammar issues. Maybe something like:

We do not do an in depth check of the id to see if we can connect, since the we do not have the item_class instance.

danyeaw (approved):
Everyday the code gets more and more clear thanks to these great refactorings, nice one! Sorry for the single minor nitpick on grammar :+1:

gaphor/diagram/diagramtools/dropzone.py
amolenaar: Thanks for catching this.

PR #581, opened 2020-12-29
Release Gaphor as an AppImage

### LINKED ISSUE #534: Appimages for Gaphor
It would be nice to have an Appimage for Gaphor.

danyeaw:

This PR adds support for and automatically builds an AppImage as part of our release process. The website needs to also be updated to include a link to the latest released version.

### PR Checklist

### PR Type

- [ ] Bugfix
- [X] Feature

Flatpak only for Linux.

Issue Number: Closes #534 

### What is the new behavior?
Flatpak and AppImage.

- [ ] Yes
- [X] No


danyeaw:
Launching the CI built AppImage is giving me an error (works fine locally):
```
  File "/tmp/.mount_gaphorXPoPFe/opt/python3.9/bin/gaphor", line 6, in <module>
    from gaphor.ui import main
  File "/tmp/.mount_gaphorXPoPFe/opt/python3.9/lib/python3.9/site-packages/gaphor/ui/__init__.py", line 10, in <module>
    import gi
  File "/tmp/.mount_gaphorXPoPFe/opt/python3.9/lib/python3.9/site-packages/gi/__init__.py", line 40, in <module>
    from . import _gi
ImportError: libffi.so.7: cannot open shared object file: No such file or directory
```
I'll need to troubleshoot this before the PR is ready.

danyeaw:
@amolenaar This one is ready for a review, the AppImages are now being built by freezing the app using PyInstaller and then copying the right metadata and files in place. 🎉

packaging/appimage/excludelist
```diff
@@ -0,0 +1,165 @@
+# This file lists libraries that we will assume to be present on the host system and hence
```
amolenaar, line 1: I'm afraid this exclude list will become a maintenance nightmare. After every major OS update, we can expect some incompatibility here.

packaging/appimage/org.gaphor.Gaphor.appdata.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
amolenaar, line 1: Isn't this the same file we use for Flatpak?

packaging/appimage/org.gaphor.Gaphor.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
```
amolenaar, line 1: Same here: this file is also in our Flatpak repo

packaging/windows/README.md
```diff
@@ -9,10 +9,10 @@ because that is what PyGObject currently supports.
 - Run `pacman -Syu` to update packages
 - Run `pacman -S git` to install git
 - Run `git clone https://github.com/gaphor/gaphor.git`
-- Execute `win-installer/msys2-install.sh` to install all the needed dependencies.
+- Execute `windows/msys2-install.sh` to install all the needed dependencies.
```
amolenaar, line 12: `packaging/windows/msys2-install.sh`?

packaging/gaphor.spec
```diff
@@ -33,13 +33,13 @@ exe = EXE(pyz,
           a.scripts,
           options=[],
           exclude_binaries=True,
-          name='gaphor',
+          name='gaphor-exe',
```
amolenaar, line 36: Why the name change?

packaging/appimage/org.gaphor.Gaphor.appdata.xml
danyeaw: Yes it is, any ideas on how we could keep the files common across the repos? The only way I know is to put these files in a git submodule.

packaging/windows/README.md
danyeaw: Good catch, I'll update this.

packaging/gaphor.spec
danyeaw: In Windows, the executable gets a .exe extension. But for Linux and macOS the executable doesn't get an extension. This causes the executable name to conflict with the name of the gaphor directory. We could call it something else if you have any ideas.

packaging/appimage/excludelist
danyeaw: Ya that's a good point, but I'm not sure what to do about that, it is kind of how AppImage is setup. I could change the script to always get the latest version from here: https://github.com/AppImage/pkg2appimage/blob/master/excludelist

packaging/appimage/org.gaphor.Gaphor.appdata.xml
amolenaar: Maybe we could a checkout of the Gaphor repo from the build script in our flatpak repo. I suppose multi-repo checkouts are supported, as they are for Azure Pipelines.

packaging/gaphor.spec
amolenaar: Ah, I see. That sounds fair.

packaging/appimage/org.gaphor.Gaphor.appdata.xml
danyeaw: I like this idea! Flathub uses a buildbot and I'm not sure how configurable things are from our repo. I will look in to this more.

packaging/appimage/excludelist
danyeaw: Another option is we could only include the items in the excludelist that pertain to us based on the files it found from the wider list:
```
libstdc++.so.5
libX10.so.6
libgio-3.0.so.0
libgdk_pixbuf-3.0.so.0
libfontconfig.so.0
libexpat.so.0
libgcc_s.so.0
libglib-3.0.so.0
libgpg-error.so.-1
libp10-kit.so.0
libuuid.so.0
libz.so.0
libgobject-3.0.so.0
libpangoft1-1.0.so.0
libpangocairo-2.0.so.0
libpango-2.0.so.0
```
Do you think this would be a little more manageable?

packaging/appimage/org.gaphor.Gaphor.appdata.xml
danyeaw: I looked into whether we can have the build files (appdata.xml and desktop) centralized a bit more. It looks like projects like [Lollypop](https://gitlab.gnome.org/World/lollypop) are using meson to build the project and move the files needed since flatpak-builder supports meson. Then all the files we have in the share folder of the flathub repo could be moved to our main repo. I think this would be a whole new project to move things to meson, so I suggest we a separate PR for that.

amolenaar (approved):
LGTM

Issue #578, opened 2020-12-28
RuntimeError When Opening Safety.gaphor


danyeaw:
**Describe the bug**
A RuntimeError is thrown when trying to open the Safety.gaphor model.

**To Reproduce**
1. Open Gaphor
2. Load `models/Safety.gaphor`
4. Loading the model throws a runtime error

```
Traceback (most recent call last):
  File "/home/dan/Projects/gaphor/gaphor/ui/actiongroup.py", line 138, in _action_activate
    method(from_variant(param))
  File "/home/dan/Projects/gaphor/gaphor/ui/appfilemanager.py", line 121, in action_open_recent
    self.load(filename)
  File "/home/dan/Projects/gaphor/gaphor/ui/appfilemanager.py", line 49, in load
    file_manager.load(filename)
  File "/home/dan/Projects/gaphor/gaphor/ui/filemanager.py", line 109, in load
    worker.reraise()
  File "/home/dan/Projects/gaphor/gaphor/ui/gidlethread.py", line 115, in reraise
    raise exc
  File "/home/dan/Projects/gaphor/gaphor/ui/gidlethread.py", line 119, in __generator_executer
    result = next(self._generator)
  File "/home/dan/Projects/gaphor/gaphor/storage/storage.py", line 308, in load_generator
    factory.model_ready()
  File "/home/dan/Projects/gaphor/gaphor/core/modeling/elementfactory.py", line 159, in model_ready
    self.handle(ModelReady(self))
  File "/home/dan/Projects/gaphor/gaphor/core/modeling/elementfactory.py", line 182, in handle
    self.event_manager.handle(event)
  File "/home/dan/Projects/gaphor/gaphor/core/eventmanager.py", line 55, in handle
    self._events.handle(e)
  File "/home/dan/Projects/gaphor/.venv/lib/python3.9/site-packages/generic/event.py", line 64, in handle
    handler(event)
  File "/home/dan/Projects/gaphor/gaphor/ui/mainwindow.py", line 237, in _new_model_content
    for diagram in self.element_factory.select(
  File "/home/dan/Projects/gaphor/gaphor/core/modeling/elementfactory.py", line 116, in select
    yield from (e for e in self._elements.values() if expression(e))
  File "/home/dan/Projects/gaphor/gaphor/core/modeling/elementfactory.py", line 116, in <genexpr>
    yield from (e for e in self._elements.values() if expression(e))
RuntimeError: OrderedDict mutated during iteration
```

**Expected behavior**
[...]

PR #583, opened 2020-12-29
Create style sheet on load and on new model

amolenaar:
### PR Type

- [X] Bugfix
- [ ] Feature

`StyleSheet` was created instantly, by `Diagram`. This could cause issues when iterating the element factory (causing the elements dict to change unexpectedly.

Issue Number: #578 

### What is the new behavior?

Do no longer create it instantly, as it can cause problems with model loading.

Also the default style is set as a default attribute value.
 
- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/core/modeling/diagram.py
+++ b/gaphor/core/modeling/diagram.py
@@ -5,7 +5,6 @@ Diagrams can be visualized and edited.
 from __future__ import annotations
 
 import logging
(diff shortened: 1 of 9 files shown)


danyeaw (approved):
LGTM!

danyeaw:
Thanks for fixing this @amolenaar!

Issue #585, opened 2020-12-31
Copy/Paste Causes Dispatcher Warnings and KeyError


danyeaw:
**Describe the bug**
Copy and paste is giving me warning and errors.

**To Reproduce**
1. Add a Class to a diagram
2. Click the Undo button:
```
ElementDispatcher WARNING Handler <function Presentation.__init__.<locals>.update at 0x7fc7e0e950d0> is not registered for <gaphor.UML.uml.Class element 768acf82-4bb5-11eb-9208-18cf5efc6bb9>.<association ownedOperation: Operation[0..*] <>-> class_>
```
repeated
The Class is removed from the model.
3. Hit the Redo button and the Class is added back to the model, but is not restored on the diagram.
4. Hit Undo again:
```
Traceback (most recent call last):
  File "/home/dan/Projects/gaphor/gaphor/core/modeling/elementdispatcher.py", line 237, in on_element_change_event
    handler(event)
  File "/home/dan/Projects/gaphor/gaphor/core/modeling/diagram.py", line 239, in _presentation_removed
    self._update_views(removed_items=(event.old_value,))
  File "/home/dan/Projects/gaphor/gaphor/core/modeling/diagram.py", line 351, in _update_views
    v.request_update(dirty_items, dirty_matrix_items, removed_items)
  File "/home/dan/Projects/gaphor/.venv/lib/python3.9/site-packages/gaphas/view/gtkview.py", line 293, in request_update
    self._qtree.remove(item)
  File "/home/dan/Projects/gaphor/.venv/lib/python3.9/site-packages/gaphas/quadtree.py", line 180, in remove
    bounds, data, clipped_bounds = self._ids[item]
KeyError: <gaphor.UML.classes.klass.ClassItem element 768bc3ba-4bb5-11eb-9208-18cf5efc6bb9>
```

**Expected behavior**
The Class is removed from the model for Undo and then restored to the model and diagram upon Redo.

**OS**
- [X] Linux - openSUSE Tumbleweed with GNOME on Wayland
- [ ] MacOS
- [ ] Windows

NOTES:

**Version**
Version of Gaphor: Master

**Additional context**

amolenaar:
Nice catch! It looks like this has to do with changes in the update feature in Gaphas (or the absence thereof).

amolenaar:
[...]

PR #587, opened 2021-01-02
Replace diagram item properties by attributes

amolenaar:
### PR Type

- [ ] Bugfix
- [ ] Feature

Some properties of Presentation items are implemented as `reversible_property`, a construct from Gaphas.

Issue Number: #585

### What is the new behavior?

Implement the reversible properties as association and attribute properties, as is used in the Gaphor data model. This will trigger state changes and work nicer with the undo system.

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/UML/actions/actionspropertypages.py
+++ b/gaphor/UML/actions/actionspropertypages.py
@@ -15,6 +15,7 @@ class ObjectNodePropertyPage(PropertyPageBase):
     ORDERING_VALUES = ["unordered", "ordered", "LIFO", "FIFO"]
 
     subject: UML.ObjectNode
+    item: ObjectNodeItem
 
     def __init__(self, item):
         self.item = item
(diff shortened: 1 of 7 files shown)


gaphor/UML/actions/actionspropertypages.py
```diff
@@ -59,6 +60,7 @@ def _on_ordering_change(self, combo):
     @transactional
     def _on_ordering_show_change(self, button, gparam):
         self.item.show_ordering = button.get_active()
+        print("show_ordering", self.item.show_ordering)
```
danyeaw, line 63: Is this print statement still needed?

gaphor/UML/actions/activitynodes.py
```diff
@@ -179,22 +177,9 @@ def __init__(self, diagram, id=None):
...
-            self._combined = value
-        else:
-            super().load(name, value)
-
-    @observed
-    def _set_combined(self, value):
-        self._combined = value
-
-    combined = reversible_property(lambda s: s._combined, _set_combined)
+    combined: relation_one[UML.ControlNode] = association(
```
danyeaw, line 180: I like how adding these as associations reduced the number of methods needed :+1:

danyeaw (approved):
@amolenaar This looks great!

gaphor/UML/actions/actionspropertypages.py
amolenaar: Removed in #588

amolenaar:
Merge #588 instead

Issue #390, opened 2020-08-08
Guide Users to Create Valid Relationships


danyeaw:
This issue is part of the wider epic to make it more clear that Gaphor has an underlying data model and that it isn't just a drawing tool.

### I'm submitting a...
- [ ] Bug report  
- [X] Feature request
- [ ] Documentation issue or request

### Current behavior
It isn't very apparent that Gaphor is constraining which elements a relationship can connect to. You can only really tell that it was connected after you fully drop the relationship and it turns red. I also don't think the green and red relationship ends are very obvious to a new user or clear what they mean

### Expected behavior
1. When you select a relationship from the palette, it should gray out (50% opacity) anything that it isn't allowed to connect to.
1. Once you hover over any valid object it should highlight the object (potentially in green), if you hover over an invalid object it should highlight it in red
1. Starting the drag anywhere on top of a valid element should make the connection (not just the border). Starting a drag over an invalid element should be ignored
1. Once you have connected the head of the relationship, it should gray out anything that the tail can't connect to
1. If there is only one valid element to connect to, it should show an icon for that element to coach the user
1. Hovering over a valid object should highlight it as well
1. Dropping anywhere over the valid object should complete the connection. Dropping over an invalid object should create a new node in the path.

I would also want similar behavior when moving one end of a relationship to another element. I think we could then remove the colors from the ends of the relationships, because then there would be no way to create an unconnected relationship.

![dragging](https://user-images.githubusercontent.com/10014976/89718321-a5966580-d98b-11ea-8f14-453267f69a26.png)

![hoverhighlight](https://user-images.githubusercontent.com/10014976/89718310-931c2c00-d98b-11ea-9e96-e3b0f203af2c.png)

### If this is a feature request please fill out the following:
<pre><code>
As a user:
I want to be guided on what relationships are valid:
So that it is very clear that there is a data model constraining what I can draw,
so that it is easier to model,
and so that there are no unglued relationships in the model:
</code></pre>

NOTES:

amolenaar:
> Hovering over a valid object should highlight it as well

@danyeaw Do you think the connection should be placed at the intersection of the line and the port? Or should the line point to the center of the object (when you drag the element, the line keeps pointing at the center)? Or maybe a combination of both?

danyeaw:
> Do you think the connection should be placed at the intersection of the line and the port? Or should the line point to the center of the object

@amolenaar I think the intersection of the line and the port.

amolenaar:
[...]

PR #603, opened 2021-01-14
In element select

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Lines only connect when handles are connected to the edge of an element.

Issue Number: #390 

### What is the new behavior?

A line will connect to an element (shape) even if the mouse cursor is dragged inside the element.

- [X] Yes
- [ ] No

### DIFF
--- a/gaphor/tests/testcase.py
+++ b/gaphor/tests/testcase.py
@@ -83,12 +83,8 @@ class TestCase(unittest.TestCase):
         adapter = Connector(item, line)
         return adapter.allow(handle, port)
 
-    # def connect(self, line, handle, item, port=None):
     def connect(self, line, handle, item):
-        """Connect line's handle to an item.
-
-        If port is not provided, then first port is used.
-        """
+        """Connect line's handle to an item."""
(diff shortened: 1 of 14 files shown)


gaphor/tests/testcase.py
```diff
@@ -83,24 +83,23 @@ def allow(self, line, handle, item, port=None):
         adapter = Connector(item, line)
         return adapter.allow(handle, port)
 
-    def connect(self, line, handle, item, port=None):
+    # def connect(self, line, handle, item, port=None):
+    def connect(self, line, handle, item):
         """Connect line's handle to an item.
 
         If port is not provided, then first port is used.
```
danyeaw, line 90: I think this docstring needs to be updated since port is no longer a parameter.

danyeaw (approved):
This is working really well! What do you think about us adding blue element highlighting like Gaphas is using?

gaphor/tests/testcase.py
amolenaar: :+1:

Issue #605, opened 2021-01-16
Association cut+paste not working


amolenaar:
**Describe the bug**

When cut + pasting a 2 classes and an association (all connected), the association is not pasted correctly. It looks like the head/tail subjects are not pasted.

**To Reproduce**

Unit test error

**Expected behavior**

**Screenshots**

NOTES:

**Version**
Version of Gaphor: master

PR #607, opened 2021-01-16
Fix association copy paste

amolenaar:
### PR Type

- [X] Bugfix
- [ ] Feature

Some diagram items have extra associations. These are not copied propertly. When doing a _cut-n-paste_, the paste data is incomplete.

* Association items missed elements for connected head and tail
* Partition item consisted of only one partition after pasting

Issue Number: #605

### What is the new behavior?

Copy, cut and paste works as expected.

- [ ] Yes
- [X] No

### Other information

This PR turned out bigger than I expected. Copy/paste did not support the copy of extra model elements, such as the partitions for PartitionItem. To support this I "flattened" the list of elements to copy. While updating, I also ran into some ordering issues, which I solved by making the `paste()` function a generator for normal elements.

Downside is that the "external" interfaces for `copy()` and `paste()` are different from the "internal" one (used to copy individual objects and their mandatory relations). 

Upside is that making a copy of an item, along with a bunch of model elements is really simple. I did not need special paste functions anymore :). 

Exception handling around Connector.connect() has been removed. Instead the connectors should just work. If they fail, a rollback should occur.


gaphor/UML/states/copypaste.py
```diff
@@ -1,58 +1,20 @@
...
-        paste_element(copy_data.exit, diagram, lookup)
-    if copy_data.do_activity:
-        paste_element(copy_data.do_activity, diagram, lookup)
-    return paste_element(copy_data.state, diagram, lookup)
-
-
-class TransitionCopy(NamedTuple):
-    transition: ElementCopy
-    guard: Optional[ElementCopy]
+    yield element.id, copy_named_element(element)
```
danyeaw, line 7: Wow, really nice simplifications!

danyeaw (approved):
Looks great!

PR #608, opened 2021-01-17
Allow spaces in attribute and operation name and type

### LINKED ISSUE #601: Association Multiplicity Doesn't Work for End Names with Spaces
**Describe the bug**
Multiplicity doesn't work if there is a space in the name of an association end. Also an extra + is added in front of the name on the diagram.

**To Reproduce**
1. Create two classes
2. Connect an association between them
3. Add a name with a space in it like `target Event` and add a multiplicity like `[1]`

**Expected behavior**
The multiplicity is recognized and the 1 is moved to the multiplicity box

**Screenshots**
With a space:
Screenshot from 2021-01-12 20-06-20

Without a space:
Screenshot from 2021-01-12 20-08-53

**Version**
Version of Gaphor: master

amolenaar:
### PR Type

- [ ] Bugfix
- [x] Feature

Spaces are not allowed in attribute and operation names and types.

Issue Number: #601

### What is the new behavior?

Spaces can be used in attributes and operations. Expressions like this can be parsed:

```
+ attr with spaces: some type
operation with spaces(some param : some type) : some return type
```

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/UML/tests/test_umlfmt.py
+++ b/gaphor/UML/tests/test_umlfmt.py
@@ -31,7 +31,7 @@ def add_tag_is_foo_metadata_field(e, factory):
     "text,formatted_text",
     [
         ("", ""),
-        ("not an attribute===foobar", "+ not an attribute===foobar"),
+        ("an attribute===foobar", "+ an attribute = ==foobar"),
         ("myattr", "+ myattr"),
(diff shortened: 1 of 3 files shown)


danyeaw (approved):
LGTM!

Issue #632, opened 2021-01-31
Highlight (drop) and gray-out (disabled) style via CSS


amolenaar:
We have CSS, but we also have custom code that deals with highlighting items and grayed-out elements.

This styling should be defined via CSS as well.

We already have a `drop` pseudo selector. 
We need one for grayed-out/disabled element, still.

- [x] Set grayed out style (`:disabled`) via style sheet
- [x] Create a separate system style sheet
- [ ] Remove diagram fallback style
- [x] Remove `highlight-color` style property
- [x] Render multiple style sheets (maybe add user style sheet as well?)
- [x] Allow setting line-width for line-like elements
- [ ] Maybe we need an extra property (`opacity`) to make disabling items easier to describe.

danyeaw:
Closed by #638

PR #638, opened 2021-02-06
Drop and gray out style

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

There was special code that dealt with drop hover and grayed out states.

Issue Number: #632 

### What is the new behavior?

All style changes are managed via CSS.

- [ ] Yes
- [X] No

### Other information

* Config grayed out (`disabled`) items via CSS
* Add `opacity` CSS property
* Added absolute size values for fonts (`small`, `large`, etc.)
* Add a system style sheet

Items that can be dropped upon now display slightly different, 'cause I removed the `highlight-color` property. Use the `:drop` pseudo selector instead.

### DIFF
--- a/docs/style_sheets.md
+++ b/docs/style_sheets.md
@@ -132,8 +132,6 @@ family, size, weight.
                         ``background-color: rgb(255, 255, 255);`` |br|
                         ``background-color: hsl(130, 95%, 10%);``
 ``color``               Color used for lines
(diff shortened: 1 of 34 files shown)


danyeaw (approved):
This looks great, I like that the gray out is styled and it also like the refactoring as well!

Issue #614, opened 2021-01-19
Add support for the C4 model


amolenaar:
C4 model is a lean graphical notation technique for modelling the architecture of software systems.

#### Describe the solution you'd like

Support in Gaphor for drawing C4 models.

The C4 toolbox should at least also support sequence diagrams.

#### Additional context

* https://c4model.com/#Notation

C4 defines the following component types:
- [x] Person (Actor + description + location)
- [x] Software System (Package + description  + location)
- [x] Container (Component + description + technology)
- [x] Container: Database (Component + description + technology)
- [x] Component (Component + description + technology)
- [x] Relationships -> (Dependency + technology)
- [x] Code components -> UML class, sequence diagrams

Software System, Container and Component are presented as a rectangular box.
Database is presented as a classic database, with ellipse on top.
If there are elements shown in the box, it's drawn as a dashed line box with the text in the bottom-left corner.

In the editor, the _Technology_ field should auto-complete, based on technologies already used in the model.

Maybe apply description, technology and location to Software System, Container and Component? Allow for other container types as well?

Should we also color code, based on type?

PR #645, opened 2021-02-12
C4 model

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

C4 model is missing from Gaphor. It would be a nice addition, since it's more accessible than UML.

Issue Number: #614 

### What is the new behavior?
A C4 model menu with C4 model elements

image


amolenaar:
I think pretty much everything is in the C4 model.

Only exception is auto-complete for the technology field.

gaphor/C4Model/toolbox.py
```diff
@@ -0,0 +1,118 @@
...
+            new_item_factory(uml_items.DependencyItem),
+        ),
+    ),
+)
+
+
+c4model_toolbox_actions: ToolboxDefinition = (
+    general_tools,
+    c4,
+    classes,
```
danyeaw, line 114: Is it common for people to mix C4 with UML? I was expecting that only the general and c4 toolbars would be used, but if this is common then maybe this is a good approach.

danyeaw:
Do we need to add icons for the namespace treeview?

image

danyeaw:
Would this be similar to what I was taking about on gitter to use an svg for an element icon?
image
could be replaced by the shape of a person?
image

gaphor/C4Model/toolbox.py
amolenaar: UML Classes can be used for class modeling. C4 focuses on components and software systems. See https://c4model.com, the Metamodel section.

amolenaar:
@danyeaw Forgot to push :see_no_evil:. To my defense: it was almost midnight :sweat_smile: 

image

I have to add a description field for Person objects still.

danyeaw:
@amolenaar No problem, I can review the PR this evening!

gaphor/C4Model/toolbox.py
danyeaw: That makes sense, I was just trying to understand the overlap. If it was extremely common to overlap then we should keep them together on the same toolbar, otherwise users could select UML from the dropdown when they want to draw some UML stuff.

danyeaw (approved):
@amolenaar This looks really great! I am excited for C4 model support in gaphor. :tada: :1st_place_medal:

gaphor/C4Model/tests/test_grouping.py
```diff
@@ -0,0 +1,54 @@
+from gaphor import UML
+from gaphor.C4Model.c4model import C4Container
+from gaphor.C4Model.diagramitems import C4ContainerItem
+from gaphor.C4Model.grouping import ContainerGroup
+
+
+def container(diagram, element_factory):
+    subject = element_factory.create(C4Container)
+    print(diagram)
```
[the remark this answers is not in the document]
amolenaar: Sure thing!

gaphor/codegen/profile_coder.py
```diff
@@ -65,46 +69,33 @@ def get_class_extensions(cls: UML.Class):
             yield meta_cls
 
 
-def create_class_trees(classes: List[UML.Class]) -> Dict[UML.Class, List[UML.Class]]:
+def create_class_hierarchy(
+    classes: List[UML.Class],
+) -> Dict[UML.Class, List[UML.Class]]:
     """Create a tree of UML.Class elements.
```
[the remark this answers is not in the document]
amolenaar: It's a hierarchy. a tree. So I think this is fine.

PR #658, opened 2021-02-15
Convert tests to pytest

TiemenSch:
### PR Checklist

- [x] I have read, and am following the [Contributor guide](https://github.com/gaphor/gaphor/blob/master/CONTRIBUTING.md)
- [x] I have read and understand the [Code of Conduct](https://github.com/gaphor/gaphor/blob/master/CODE_OF_CONDUCT.md)

### PR Type

- [x] Refactoring (no functional changes, no api changes)

Issue Number: 129

### What is the new behavior?

Converted all unittest tests to pytest.

- [x] No

### Other information

``pytest-mock`` was added as a dev dependency so ``unittest.mock.Mock`` could be deprecated.


TiemenSch:
@danyeaw, would you mind taking a look at this PR? 

There's multiple ways to achieve the same goal with pytest and some depend on some kind of 'style'.

I tried to retain as much of the test class names so that filtering tests still pretty much works with the same words (i.e. when using "-k" as a pytest arg). 

Please let me know what you think.

amolenaar:
Hi @TiemenSch,

I think this is a huge step forward :+1: . 

The style we used so far, is to create separate fixtures for the services and wire them together. This makes it easier to structurally move tests from the big "case" to smaller units.

amolenaar (approved):

TiemenSch:
Good to hear it's appreciated :)! I'm going to push some last cleanups in a minute. I would love to hear what you would deem the best way forward.

I was thinking a little more on the fixtures and there's an overridable pattern, where you override the services on a local basis (e.g. per module or class), while leaving the rest untouched. I tried to write up a minimal example:

```python
@pytest.fixture
def services():
    return [
        "event_manager",
        "component_registry",
        "element_factory",
        "element_dispatcher",
        "modeling_language",
        "sanitizer",
    ]


@pytest.fixture
def case(services):
    return Case(services=services)


class TestWithinClass:
    @pytest.fixture
    def services(self):
        return ["event_manager", "element_factory"]

    def test_within_class(self, case):
        print(case.services)  # prints the two services


def test_out_of_class(case):
    print(case.services)  # prints the original set of services
```

However, there are some Cases with an overridden ``__init__``, for which this example does not account yet. 

Food for thought.

danyeaw:
Congrats on your first code contribution, we really appreciate it, and I hope this is the first of many to come! :tada:

This is a huge step forward in cleaning up our tests by finally cutting out unittest. In the future, we can continue to refactor these tests to add more fixtures and move away from classes to test functions.

@all-contributes please add @TiemenSch for code

danyeaw:
@all-contributors please add @TiemenSch for code

Issue #690, opened 2021-03-14
Add containment link for the requirement diagram


freezed-or-frozen:
**Is your feature request related to a problem? Please describe.**
I cannot find how to add a containment link between two requirements block in the requirement diagram
![Capture d’écran de 2021-03-14 10-06-16](https://user-images.githubusercontent.com/28558919/111063064-01d6e000-84ad-11eb-8a10-9e4a6f5e574c.png)

**Describe the solution you'd like**
Add a containment link after the "verifiy" link
![Capture d’écran de 2021-03-14 10-05-38](https://user-images.githubusercontent.com/28558919/111063070-0bf8de80-84ad-11eb-96bb-60226c0027d0.png)

Gaphor looks very promising. Congratulations for your work and thank you !

danyeaw:
@all-contributors please add @freezed-or-frozen for bug

danyeaw:
@freezed-or-frozen Thanks for the report, you are right - it looks like we are missing the containment relationship! :+1: 

This would be pretty straightforward to add I think and we are always looking for new contributors. Is this something you would be interested in helping with?

danyeaw:
@freezed-or-frozen In #707 we have completed support for this and we should be able to release version 2.4 soon:

![image](https://user-images.githubusercontent.com/10014976/113615105-6d7e1a00-9621-11eb-866f-bd31e238a871.png)

freezed-or-frozen:
Thank you very much for your work, it will be very useful in my school.

danyeaw:
@freezed-or-frozen We would be very interested in how you are using Gaphor. Are you taking a class on SysML?

freezed-or-frozen:
Yes, I use Gaphor when working on project with students. SysML requirements diagram helps to see what it is needed by the client. I use it to in UML class but just as a drawing tool. When dealing with generation/reverse code, I switch to StarUML.

danyeaw:
@freezed-or-frozen Thanks! Are these students taking a course at a university? I am just curious on how it is being used so that we can tailor things to the users.

freezed-or-frozen:
We used it with "A levels" and "BTEC Higher National Diploma" students.

danyeaw:
Thanks you for sharing!

PR #707, opened 2021-03-26
Add Support for Containment Relationship

danyeaw:

This PR adds support for the containment relationship with UML/SysML.

### PR Checklist

### PR Type

- [ ] Bugfix
- [X] Feature

It is possible to nest objects under a package, but the containment relationship is not supported.

Issue Number: #690 

### What is the new behavior?
Support for the containment relationship for those elements that can currently be nested.

- [ ] Yes
- [X] No

### TODO

- [x] Containment relationship works to nest packages
- [x] Containment relationship allows nesting of other currently nestable elements

### DIFF
--- a/gaphor/SysML/requirements/requirementstoolbox.py
+++ b/gaphor/SysML/requirements/requirementstoolbox.py
@@ -7,6 +7,7 @@ from gaphor.diagram.diagramtoolbox import ToolDef, ToolSection, namespace_config
 from gaphor.diagram.diagramtools import new_item_factory
 from gaphor.SysML import diagramitems as sysml_items
 from gaphor.SysML import sysml
+from gaphor.UML.classes import ContainmentItem
 
 requirements = ToolSection(
     gettext("Requirements"),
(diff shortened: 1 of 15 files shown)


danyeaw:
There isn't a containment relationship in UML, instead, the relationship in the diagram represents a `UML.Package.ownedElement`. I think this is similar to how a CommentLine represents an association but it isn't a unique element itself.

@amolenaar If you have a min to look at this, I must be doing something silly because when I run Gaphor in a debugger `ContainmentConnect` `allow()` method isn't being called when I create a new ContainmentItem.

amolenaar:
@danyeaw Shouldn't the containment relationship be in the SysML profile? I can see no reference of it in the UML specs.

danyeaw:
@amolenaar I don't think the UML spec actually uses the term containment, but it is shown in the spec. Here is section 12.2.5 where it shows the notation for a package:
image

I have seen it called the Containment relationship in most of the tools and books. This is from UML 2 and the Unified Process book:
image

amolenaar:
Indeed. It's referred to as "ownership notation". :smiley: 

I think it's working. I added a little behavior for the disconnect action as well. When disconnected, it moves the item to the same namespace as the diagram. Is this what you expect?

danyeaw:
Thanks again for your help with this @amolenaar! I think I was stuck because I was missing `containmentconnect.py` from the `__init__.py`, I knew it had to be something silly like that :sweat_smile:.

amolenaar:
Is it good to merge now? Or are there features missing still?

Issue #482, opened 2020-10-06
When I open a file an already opened model file, show the window with already opened model


amolenaar:
**Is your feature request related to a problem? Please describe.**

When i open a model, and open the same model again, I have two windows, showing the same model.

**Describe the solution you'd like**

When I open an already opened model, present the window with the already opened model.

danyeaw:
@amolenaar What if a user wants to have two diagrams open side by side? I don't think our current tabs scheme allows that, which I think is fine. However, there might be some use cases where users want to have the model open twice.

amolenaar:
You want to keep it as is, for now?

danyeaw:
I would be open to changing the behavior, I think we just need to keep in mind this use case as well.

danyeaw:
Closed by #711.

PR #711, opened 2021-03-28
Focus already opened model

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

If an already opened model is opened again, it's loaded in a new window.

Issue Number: #482

### What is the new behavior?

When an already opened model is "opened", the window is focused.

### Notes

* `gaphor.ui.event.FileLoaded/Saved` have been replaced by `gaphor.event.ModelLoaded/Saved`

### DIFF
--- a/gaphor/application.py
+++ b/gaphor/application.py
@@ -20,6 +20,8 @@ from gaphor.entrypoint import initialize
 from gaphor.event import (
     ActiveSessionChanged,
     ApplicationShutdown,
+    ModelLoaded,
+    ModelSaved,
     ServiceInitializedEvent,
     ServiceShutdownEvent,
     SessionCreated,
@@ -169,6 +171,7 @@ class Session(Service):
     def __init__(self, services=None):
         """Initialize the application."""
         services_by_name: Dict[str, Service] = initialize("gaphor.services", services)
+        self._filename = None
 
         self.event_manager: EventManager = cast(
             EventManager, services_by_name["event_manager"]
@@ -181,6 +184,8 @@ class Session(Service):
             logger.info(f"Initializing service {name}")
             self.component_registry.register(name, srv)
             self.event_manager.handle(ServiceInitializedEvent(name, srv))
+
+        self.event_manager.subscribe(self.on_filename_changed)
 
     @property
     def is_new(self):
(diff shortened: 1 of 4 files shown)


danyeaw (commented):
This implementation looks really good!

Since the app is a singleton, I am concerned about not allowing someone to open a model twice. I think there are some use cases where you would want to be able to reference a diagram while viewing another diagram in the same model.

Is the use case here that the user doesn't know they already have the model open or do they know they have it open and they are using the open menu to switch projects?

Do we want to at least add an option to the open menu or a popup dialog to allow a user to open the model in a new window if they need to?

amolenaar:
> Since the app is a singleton, I am concerned about not allowing someone to open a model twice. I think there are some use cases where you would want to be able to reference a diagram while viewing another diagram in the same model.

This can also lead to strange situations, where both instances of the same model are changed differently. 

This could also be resolved by allowing users to detach diagrams from the notebook view, or open it in a second notebook, next to the active one, like one can do with files in VSCode (ctrl-click).

> Is the use case here that the user doesn't know they already have the model open or do they know they have it open and they are using the open menu to switch projects?

They could also use the file explorer.

> Do we want to at least add an option to the open menu or a popup dialog to allow a user to open the model in a new window if they need to?

We can do this if the user explicitly opens the model though te open menu. If it's "opened" by double clicking a model file in their file manager, I suppose it's better to just show the already opened model instead. That's the behavior I see with text editors at least ;).

danyeaw:
> This could also be resolved by allowing users to detach diagrams from the notebook view, or open it in a second notebook, next to the active one, like one can do with files in VSCode (ctrl-click).

Yup, I agree.

> They could also use the file explorer.

Ahh yes, I hadn't thought about that use case.

> We can do this if the user explicitly opens the model though te open menu. If it's "opened" by double clicking a model file in their file manager, I suppose it's better to just show the already opened model instead. That's the behavior I see with text editors at least ;).

That's a good idea, I like that!

amolenaar:
@danyeaw I added a little dialog if a user wants to open an already opened model.

I'm not really happy with the phrasing: the buttons should be more descriptive than "Yes"/"No", but that's something I want to flesh out separately, since that means all question dialogs do need an update.

danyeaw (approved):
I really like this, nice one! :stars:

Issue #618, opened 2021-01-21
Use GTK 4 


amolenaar:
Gaphor is running on GTK+ 3, currently.

There are a couple of improvements in GTK 4 that make an (early?) migration an option, IMHO:

* Better cross platform support: a lot has been done to improve interoperability with Windows and macOS. I assume GTK4 will run smoother on those platforms.
* Better support for copy-and-paste: we can purely rely on GTK API's.
* More and better event controllers. Event controllers were backported from GTK 4 to GTK+ 3.

During migration we should still support GTK+ 3.

What we need before migrating:

- [x] Good availability of GTK 4 on the latest linux platforms (Arch :heavy_check_mark:, Ubuntu :heavy_check_mark: , Fedora :heavy_check_mark: ).
- [x] [PyGObject](https://gitlab.gnome.org/GNOME/pygobject) should be ready (works with master)
- [x] A GTK 4 base in Flatpak.
- [x] A GTK 4 package in MSys.
- [x] A GTK 4 package in Homebrew.
- [X] gtk-mac-integration? Still needed for GTK integration on macOS? _RESOLVED_: looks like [GTK can handle it by itself](https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gtk/gtkapplication-quartz.c#L108).

What needs changing:

- [x] The tool bar should be rebuild using `Gtk.Expander` and `Gtk.FlowBox` (affects #498).
- [x] Event controllers are instantiated differently
- [x] See the migration guide: https://developer.gnome.org/gtk4/stable/gtk-migrating-3-to-4.html.
- [x] Implement Copy/paste - use `Gdk.Clipboard`. #274 
- [x] Implement DnD for Toolbox.
- [x] Toolbox: [Button remains in "pressed" state when used with GtkDragSource](https://gitlab.gnome.org/GNOME/gtk/-/issues/4416).
- [x] Fix issues with slow toolbox (`Gio.Action`), and recent files popup.
- [x] ~DnD is [not working for `Gtk.TreeView`](https://gitlab.gnome.org/GNOME/gtk/-/issues/3649) for GTK 4 ([see also](https://gitlab.gnome.org/GNOME/gtk/-/issues/3656)).~
[...]

PR #746, opened 2021-04-18
Preparations for GTK4 migration

amolenaar:
### PR Type

- [ ] Bugfix
- [X] Feature

Do as much as possible of the migration without breaking GTK3 compatibility.

https://developer.gnome.org/gtk4/stable/gtk-migrating-3-to-4.html

Issue Number: #618 

### What is the new behavior?

Create `.ui` files for GTK 4, leave the `.glade` files for GTK 3.

This PR covers most of the items as defined in the pre-migration section.

- Use styles instead of GtkRelief and GtkShadowType
- Use controllers instead of `@binding-set` in CSS
- fix deprecation errors

- Split the big glade file from gaphor.diagram to make it easier to migrate

- [ ] Yes
- [X] No


amolenaar:
Not needed. Just merge #756 instead.

Issue #777, opened 2021-05-05
Reordering attributes or operation not working


icfmp:
**Describe the bug**
When I drag and drop attributes or operations it doesn't apply to the class

**To Reproduce**
1. Open Gaphor
2. Select a class
3. Move an attribute or an operation in the edit box
4. The attribute or operator moved in the edit box but not in the view.
5. Choose another class and get back to the class you modified : modifications are lost.

**Expected behavior**
Attributes and operator order should be saved.

**Screenshots**
![image](https://user-images.githubusercontent.com/71736258/117165644-f5dd0f80-adc5-11eb-939d-294ad16f1df7.png)

**OS**
Ubuntu 21.04 + flatpak + flathub

**Version**
Version of Gaphor: 2.4.0 from flathub

amolenaar:
Hi @icfmp, Thanks for reporting this.

Indeed, if you reorder the attributes with drag-n-drop they are not reordered. I'll investigate this further.

As a workaround (for now), you can use <key>+</key> and <key>-</key> to reorder the elements. Then they will be updated in the diagram as well.

danyeaw:
@all-contributors please add @icfmp for bug

icfmp:
> Hi @icfmp, Thanks for reporting this.
> 
> Indeed, if you reorder the attributes with drag-n-drop they are not reordered. I'll investigate this further.
> 
> As a workaround (for now), you can use + and - to reorder the elements. Then they will be updated in the diagram as well.

@amolenaar Ok, I tried it. I can indeed drag a question down with =, but + isn't working.

PR #780, opened 2021-05-08
DnD for element editor lists

amolenaar:
### PR Type

- [X] Bugfix
- [ ] Feature

Attributes and operations do not reorder when reordered with drag-and-drop.

Issue Number: #777

### What is the new behavior?

The model is updated if items are reordered via DnD.

- [ ] Yes
- [X] No

### Other information

Also changed the abstract methods not to be protected (starting with a `_`).


danyeaw (approved):
Except for the tests failing, this looks great to me!

Issue #748, opened 2021-04-18
Add Support for STPA from RAAML


danyeaw:
**Is your feature request related to a problem? Please describe.**
I would like to model System Theoretic Process Analysis (STPA) using the RAAML language. The [STPA Handbook](https://psas.scripts.mit.edu/home/get_file.php?name=STPA_handbook.pdf) and the RAAML normative examples would be the primary references.

**Describe the solution you'd like**
Add support for STPA. Since we don't currently support tables, we should be able to model it using diagrams instead. Here are the different diagrams we should be able to create based on the RAAML Examples:

![image](https://user-images.githubusercontent.com/10014976/115162575-653acb80-a072-11eb-95d1-e64139a0296a.png)
![image](https://user-images.githubusercontent.com/10014976/115162595-797ec880-a072-11eb-9901-cd67a9f138a1.png)
![image](https://user-images.githubusercontent.com/10014976/115162604-8996a800-a072-11eb-9100-550a010a0dde.png)
![image](https://user-images.githubusercontent.com/10014976/115162623-9ca97800-a072-11eb-84ed-949dae7b6fd0.png)
![image](https://user-images.githubusercontent.com/10014976/115162645-afbc4800-a072-11eb-837f-4f2da99af059.png)
![image](https://user-images.githubusercontent.com/10014976/115162658-c5ca0880-a072-11eb-8bcc-8aab61fa5186.png)
![image](https://user-images.githubusercontent.com/10014976/115162678-dbd7c900-a072-11eb-8155-e60585f5509a.png)
![image](https://user-images.githubusercontent.com/10014976/115162714-fdd14b80-a072-11eb-8450-677c0d4f8ddb.png)

**Describe alternatives you've considered**
No support for STPA.

**Additional context**

danyeaw:
![image](https://user-images.githubusercontent.com/10014976/118382172-b29e5000-b5c0-11eb-9e4c-c05eb3f1246b.png)
First diagram in Gaphor.

danyeaw:
![image](https://user-images.githubusercontent.com/10014976/118408988-43733b00-b656-11eb-9ad8-51337302919f.png)

danyeaw:
![image](https://user-images.githubusercontent.com/10014976/119199570-52dcf480-ba59-11eb-931d-deab065f9a50.png)

danyeaw:
[...]

PR #781, opened 2021-05-08
Add Initial Support for STPA

danyeaw:
[System-Theoretic Process Analysis (STPA)](https://psas.scripts.mit.edu/home/get_file.php?name=STPA_handbook.pdf) is the next portion of RAAML we will support.

### PR Checklist

- [X] I have read, and I understood the [Code of Conduct](https://github.com/gaphor/gaphor/blob/master/CODE_OF_CONDUCT.md)

### PR Type

- [ ] Bug fix
- [X] Feature

The RAAML profile only support FTA

Issue Number: #748 

### What is the new behavior?
RAAML now has FTA and STPA

- [ ] Yes
- [X] No


danyeaw:
@amolenaar I'll submit a PR to fix distlib in MSYS2, I think this PR is ready to review when you get the chance.

gaphor/SysML/requirements/connectors.py
```diff
@@ -46,7 +46,7 @@ class DeriveReqtConnect(DirectedRelationshipPropertyPathConnect):
 
 
 @Connector.register(Classified, RefineItem)
-class RefinefyConnect(DirectedRelationshipPropertyPathConnect):
+class RefineConnect(DirectedRelationshipPropertyPathConnect):
```
amolenaar, line 49: If this was already in 2.4, it should also require a migration.

gaphor/codegen/codegen.py
```diff
@@ -17,12 +17,19 @@ def main() -> None:
...
+        "--sysml_profile", help="generate a SysML profile", action="store_true"
+    )
     args = parser.parse_args()
     print(f"Generating {args.outfile} from {args.modelfile}...")
     print("  (warnings can be ignored)")
-    if args.profile:
+    if args.uml_profile:
         profile_coder.generate(args.modelfile, args.outfile, args.overrides)
+    elif args.sysml_profile:
+        profile_coder.generate(args.modelfile, args.outfile, args.overrides, True)
```
amolenaar, line 32: `True` is more descriptive when used as keyword argument: `includes_sysml=True`.

amolenaar (changes requested):
That's a lot of functionality you added!

gaphor/codegen/codegen.py
danyeaw: Fixed, thanks!

gaphor/SysML/requirements/connectors.py
danyeaw: I don't think the Connectors get loaded in the gaphor model files as an element right? I can load a model with a RefineItem that was saved in version 2.4.2 successfully.

amolenaar: Yes. For connectors it fine. Sorry. I misread the decorator :see_no_evil: .

amolenaar (approved):

PR #804, opened 2021-05-14
Improve CSS text editor

vanillajonathan:
This improves the CSS text editor to insert spaces instead of tabs, and to use 2-space indention and auto-indent.

### DIFF
--- a/gaphor/ui/elementeditor.py
+++ b/gaphor/ui/elementeditor.py
@@ -252,7 +252,7 @@ class SettingsStack:
         self.style_sheet_view.set_auto_indent(True)
         self.style_sheet_view.set_indent_width(2)
         self.style_sheet_view.set_input_hints(Gtk.InputHints.NO_EMOJI)
-        self.style_sheet_view.set_insert_spaces_instead_of_tabs(True) 
+        self.style_sheet_view.set_insert_spaces_instead_of_tabs(True)
 
         self.event_manager.subscribe(self._model_ready)
         self.event_manager.subscribe(self._style_sheet_created)
--- a/gaphor/ui/elementeditor.py
+++ b/gaphor/ui/elementeditor.py
@@ -249,6 +249,10 @@ class SettingsStack:
     def open(self, builder):
         self.style_sheet_buffer = builder.get_object("style-sheet-buffer")
         self.style_sheet_view = builder.get_object("style-sheet-view")
+        self.style_sheet_view.set_auto_indent(True)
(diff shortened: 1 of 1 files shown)


gaphor/ui/elementeditor.py
```diff
@@ -249,6 +249,10 @@ def tx_update_style_sheet(style_sheet, text):
     def open(self, builder):
         self.style_sheet_buffer = builder.get_object("style-sheet-buffer")
         self.style_sheet_view = builder.get_object("style-sheet-view")
+        self.style_sheet_view.set_auto_indent(True)
+        self.style_sheet_view.set_indent_width(2)
+        self.style_sheet_view.set_input_hints(Gtk.InputHints.NO_EMOJI)
+        self.style_sheet_view.set_insert_spaces_instead_of_tabs(True)
```
danyeaw, line 255: `style_sheet_view` is a `Gtk.TextView` and the 3 indention methods you added are for the `Gtk.Source.View`.

danyeaw (changes requested):
@vanillajonathan I am really excited to have you contribute this. Have you been able to get a development environment setup? We would be glad to answer any questions on Gitter as well. :+1: 

It looks like you are trying to apply the Gtk.Source.View methods to a Gtk.TextView and if you try to run Gaphor like this it throws an AttributeError. You might be on the right track that using the SourceView is what we want to improve the CSS editing experience. I couldn't find Gtk.Source.View in Glade though, so this may require some investigation on how to swap in the SourceView in to the `elementeditor.glade` file.

vanillajonathan:
@danyeaw No, I haven't. I also used Gtk.Source so much that I thought those functions were available on Gtk.TextView too, but they're not.

Yeah, using Gtk.SourceView would be nice, because then you can get syntax highlighting, and auto indention, and insert tabs-as-spaces. The drawback is that it introduces a dependency.

I don't know how to declare GtkSourceView in Glade. It is not part of GTK, it is part of the gtksourceview library.

I found by searching: https://cjenkins.wordpress.com/2012/05/08/use-gtksourceview-widget-in-glade/
Which suggests:
```python
GObject.type_register(GtkSource.View)
```

danyeaw:
Hi @vanillajonathan, how is your development environment setup going?

vanillajonathan:
I haven't set it up, I just dropped by for a drive-by commit.

danyeaw:
Closing this. Glad to have you contribute if you decide to set up a dev environment. :+1:

PR #808, opened 2021-05-15
Add GTK4 build on macos

amolenaar:
On macOS it's easier to install GTK4 than on Ubuntu ;)


amolenaar:
PyGObject [needs an update](https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/172)

danyeaw:
It is too bad that GitHub Actions doesn't allow for an allow-failure option for some builds, we don't want the other builds to cancel or the overall status to show as failed if GTK4 fails.

amolenaar:
Excellent suggestion! There is such an option :smiley:.

danyeaw:
Nice one!

Issue #814, opened 2021-05-20
Composite Association AttributeError


danyeaw:
**Describe the bug**
I am an AttributeError when trying to add a composite association to a diagram.

```
  File "/home/dan/Projects/gaphor/gaphor/transaction.py", line 29, in _transactional
    return func(*args, **kwargs)
  File "/home/dan/Projects/gaphor/gaphor/UML/classes/classespropertypages.py", line 672, in _on_end_navigability_change
    UML.model.set_navigability(
  File "/home/dan/Projects/gaphor/gaphor/UML/modelfactory.py", line 328, in set_navigability
    owner.ownedAttribute = end
AttributeError: 'NoneType' object has no attribute 'ownedAttribute'
gaphor.services.undomanager ERROR Could not rollback transaction
```
I then get a really long traceback as it attempts to rollback the transaction.

**To Reproduce**
1. Add two classes
2. Add a composite association

**Expected behavior**

**Screenshots**

**OS**
- [X] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

NOTES:

**Version**
Version of Gaphor: latest master version

**Additional context**

PR #820, opened 2021-05-20
Property page should only set navigability if both ends are connected

amolenaar:
### PR Type

- [X] Bug fix
- [ ] Feature

An error is raised when a composite association is added to a diagram and connected to two classes.

Issue Number: #814

### What is the new behavior?

No exception, association is properly set.


danyeaw (approved):
Thanks for the quick fix @amolenaar! :1st_place_medal:

danyeaw:
It looks like this bug is in 2.4.1 as well, we should roll a new release this weekend.

PR #851, opened 2021-06-03
Diagram items in element factory

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Currently diagram items (Presentation) are managed by diagrams, outside the view of ElementFactory.

Issue Number: N/A

### What is the new behavior?

All elements, model and presentation are managed by the Element Factory.

This simplifies the loading/saving, the undo management and the way we need to think about out models.

- [X] Yes
- [ ] No

### Other information

- [x] To do: update ModelingLanguage. We make no distinction between model and diagram items anymore.
- [X] I noticed some misplacement of element, when they're loaded the new way, instead of the old (nested).
      E.g. open `test-models/all-elements.gaphor`, save it and open it again. you'll see some lines and items are wrong. See #855.
- [x] Fix item placement for nested items

### DIFF
--- a/gaphor/diagram/copypaste.py
+++ b/gaphor/diagram/copypaste.py
@@ -212,7 +212,8 @@ def _paste_all(copy_data: CopyData, diagram, lookup) -> Set[Presentation]:
             return new_elements[ref]
 
         looked_up = lookup(ref)
-        if looked_up:
+        # TODO: Presentation element should be created anyhow
+        if looked_up and not isinstance(looked_up, Presentation):
             return looked_up
 
         elif ref in copy_data.elements:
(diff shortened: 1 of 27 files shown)


gaphor/diagram/copypaste.py
```diff
@@ -212,7 +212,8 @@ def element_lookup(ref: str):
             return new_elements[ref]
 
         looked_up = lookup(ref)
-        if looked_up:
+        # TODO: Presentation element should be created anyhow
```
danyeaw, line 215: Sorry to always  bring up these, but is there anything we need to do with this TODO before merging?

danyeaw (approved):
You are on a roll with nice refactoring and cleanup!

gaphor/diagram/copypaste.py
amolenaar: You're right :smiley: . I hadn't noticed this one.

Issue #894, opened 2021-07-13
Recursion limit met when opening document


mrmonkington:
**Describe the bug**
Gaphor crashes when I open this diagram, which I created using Gaphor 2.5.1 (flatpak) and saved, then closed earlier (with no errors at the time).

[classes.zip](https://github.com/gaphor/gaphor/files/6808802/classes.zip)

Logs/traceback below

**To Reproduce**
1. Open (after unzipping) the attached file `classes.gaphor`

**Expected behavior**
To be able to view and edit the diagram.

**OS**
- [x] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

NOTES:
Ubuntu 20.04

**Version**
Version of Gaphor: master (run from dev environment w/ python 3.9.6) and 2.5.1 (flatpak)

**Additional context**

Traceback:

```
$ gaphor ~/Documents/Work/najork/docs/modeling/classes.gaphor 
gaphor.application INFO Initializing service component_registry
gaphor.application INFO Initializing service event_manager
gaphor.application INFO Initializing service properties
gaphor.application INFO Initializing service modeling_language
gaphor.application INFO Initializing service element_dispatcher
gaphor.application INFO Initializing service element_factory
gaphor.application INFO Initializing service export_menu
gaphor.application INFO Initializing service tools_menu
gaphor.application INFO Initializing service main_window
gaphor.application INFO Initializing service consolewindow
gaphor.application INFO Initializing service toolbox
gaphor.application INFO Initializing service diagrams
gaphor.application INFO Initializing service copy
gaphor.application INFO Initializing service diagram_export
gaphor.application INFO Initializing service elementeditor
gaphor.application INFO Initializing service file_manager
gaphor.application INFO Initializing service namespace
gaphor.application INFO Initializing service recent_files
gaphor.application INFO Initializing service undo_manager
gaphor.application INFO Initializing service sanitizer
gaphor.application INFO Initializing service xmi_export
gaphor.storage.storage INFO Loading file classes.gaphor
gaphor.storage.storage INFO Read 180 elements from file
gaphor.storage.storage WARNING file b'/home/mark/Documents/Work/najork/docs/modeling/classes.gaphor' could not be loaded (maximum recursion depth exceeded while calling a Python object)
> /home/mark/Documents/Work/gaphor/gaphor/core/modeling/properties.py(359)_get_one()
-> v: T | None = getattr(obj, self._name, None)
(Pdb) step
Traceback (most recent call last):
[...]

PR #895, opened 2021-07-13
Add extra rules to avoid cyclic references

amolenaar:
This should avoid errors such as #894.

### PR Type

- [X] Bug fix
- [ ] Feature

It is possible to construct a model in such a way that, for a `ClassItem` and `InterfactItem`, where `obj.parent is obj`.

Issue Number: #894

### What is the new behavior?

An exception is raised if `obj.parent` is set to `obj`. This rule is universal for associations, so it can be checked for at model level.

- [ ] Yes
- [X] No

### Other information

This is possibly no a solution to the root cause for #894, but at least it will not create an error in model.


danyeaw (approved):
Nice checks added @amolenaar!

PR #903, opened 2021-07-17
Upgrade Windows Build Script from Bash to Python

danyeaw:
In order to eventually move our Windows build to gvsbuild, this PR upgrades the Windows build script from bash to Python to make it more cross-platform.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: N/A

- [X] Yes
- [X] No


packaging/windows/build-win-installer.py
```diff
@@ -0,0 +1,99 @@
...
+dist: Path = dir / "dist"
+Path(dist / "gaphor").mkdir(parents=True, exist_ok=True)
+
+gaphor_dist: Path = dist / "gaphor"
+seven_zip: Path = Path("C:/Program Files/7-Zip/7z.exe")
+payload: Path = dist / "payload.7z"
+portable: Path = dist / f"gaphor-{version}-portable"
+
+
+def clean_files() -> None:
```
amolenaar, line 22: Can we add the paths (in this case `portable` and `payload`) as arguments to the function? That would make it more clear what will happen to whom.

packaging/windows/build-win-installer.py
```diff
@@ -0,0 +1,99 @@
...
+    if sfx_path.is_file():
+        print(f"Sfx file found at {sfx_path}")
+    else:
+        print("Sfx file not found")
+    concatenate_files(
+        [seven_zip.parent / "7z.sfx", payload], dist / f"gaphor-{version}-portable.exe"
+    )
+
+
+clean_files()
```
amolenaar, line 95: Move these statements under a `if __name__ == "__main__"` clause. Otherwise things like Pytest's test discovery may fail (e.g. in VSCode).

Also the global folders can be defined in the if block then and passed to the functions.

amolenaar (changes requested):
Nice work. I think the resulting Python script is simple and readable :+1:

packaging/windows/build-win-installer.py
danyeaw: Great suggestion, I updated the function to pass in a list of paths as an argument.

packaging/windows/build-win-installer.py
danyeaw: Also a great idea, updated!

amolenaar (approved):

PR #905, opened 2021-07-19
Replace main Makefiles with Poetry Runner (poethepoet)

danyeaw:
Replaces the Makefiles with a Poetry runner called poethepoet in order to improve compatibility for building in Windows.

### PR Checklist
Please check if your PR fulfills the following requirements:

- [X] I have read, and I am following the [Contributor guide](https://github.com/gaphor/gaphor/blob/master/CONTRIBUTING.md)
- [X] I have read, and I understand the [Code of Conduct](https://github.com/gaphor/gaphor/blob/master/CODE_OF_CONDUCT.md)

### PR Type
What kind of change does this PR introduce?

- [ ] Bug fix
- [ ] Feature
- [X] Chore (refactoring, formatting, local variables, other cleanup)
- [ ] Documentation content changes

### What is the current behavior?

Issue Number: N/A

### Does this PR introduce a breaking change?
- [ ] Yes
- [X] No

PR #912, opened 2021-08-01
Convert CI from mingw to gvsbuild

danyeaw:
Gvsbuild allows use to build GTK for Windows using MSVC instead of GCC, which is a more native Windows environment. The main advantage is that we can use the official version of Python for Windows instead of the patched version to run with mingw. This should improve the reliability of builds in Windows just working.

This PR also replaces the Makefiles with a Poetry runner called poethepoet in order to improve compatibility for building in Windows.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

For Windows, our CI uses mingw to test and build for Windows.

Issue Number: N/A

### What is the new behavior?
Move to use gvsbuild

- [ ] Yes
- [X] No


amolenaar:
Nice! Have to try the binary though.

This PR also includes #905?

danyeaw:
@amolenaar I think #905 is ready to go. For this one the build is successful, but I am getting an OSError WinError 87 The parameter is incorrect: ' ', when I try to launch the installed Windows version built using this.

danyeaw:
Well, it works fine in console mode with debug enabled and upx disabled.

danyeaw:
I am having an issue with PyInstaller with gvsbuild GTK apps and I opened an issue here:
https://github.com/pyinstaller/pyinstaller/issues/6100

danyeaw:
@amolenaar I think this one is finally ready!

danyeaw:
@amolenaar I have tested this in macOS as well. Did you want to review it before I merge?

.github/workflows/build.yml
```diff
@@ -91,63 +92,107 @@ jobs:
         if: github.event_name == 'release'
         run: poetry publish -u __token__ -p ${{ secrets.PYPI_TOKEN }}
 
-  windows:
+  gvsbuild:
```
amolenaar, line 95: I'd rather have the build step be named windows. This makes more sense when looking at the build results.

danyeaw: Hi @amolenaar, thanks for the feedback :+1:.

I split the windows build in to two steps:

1. Build GTK for Windows, or grab it from cache if it is already built
2. Everything else (install dependencies, test, package)

This was the first step that was renamed to gvsbuild, but I just renamed it to `windows-gtk-build`, hopefully that is more clear.

I am also working on packaging GTK for Windows and putting it on [winget](https://docs.microsoft.com/en-us/windows/package-manager/winget/) (Microsoft's new Chocolately packaging system that will be built in to Windows). So hopefully we can get rid of step 1 in the near future completely. I created a repo to work on that here: https://github.com/gaphor/gtk-win-installers

PR #991, opened 2021-09-15
Use flat buttons in the header bar

amolenaar:
### PR Type

- [ ] Bug fix
- [X] Feature

We have some buttons in the header.

The GTK4 (new libawaita) look will make those buttons flat. I wondered what it would look like for Gaphor.

Issue Number: N/A

### What is the new behavior?

The items in the header are flat. This does indeed make the header less cluttered:

image

- [ ] Yes
- [X] No

### Other information

https://blogs.gnome.org/alexm/2021/09/10/cleaning-up-header-bars/

Not sure if we should actually merge this though? Will it look okay and is it intuitive on Windows and macOS?

### DIFF
--- a/gaphor/ui/mainwindow.glade
+++ b/gaphor/ui/mainwindow.glade
@@ -188,6 +188,9 @@
                 <property name="icon_size">1</property>
               </object>
             </child>
+            <style>
+              <class name="flat"/>
+            </style>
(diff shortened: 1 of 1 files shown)


danyeaw (approved):
Nice, I agree this looks really!

Issue #1027, opened 2021-09-26
Hungarian not showing followup


ovari:
Thank you @danyeaw and @amolenaar for the 2.6.1 release.

Here is the Gaphor localisation followup.

### Describe the bug
Hungarian translations on Weblate are at 100%; however, Hungarian is not completely shown in the Gaphor GUI.

### To Reproduce
https://github.com/gaphor/gaphor/issues/1024

### Expected behavior
Not all strings are showing in Hungarian.

### Screenshots
1. `Open`
2. Popups: `New diagram`, `Show/Hide Editor`, `Undo`, `Redo`, `Zoom In`, `Zoom 100%`, `Zoom Out`, 
3. `No item selected` and the rest of the right panel
![kép](https://user-images.githubusercontent.com/17465872/134790937-4912186c-7e53-4c7c-a243-8f8d9b24a969.png)

1. `Preferences` (`Preferences` button has no popup, should the popup be added to the `Preferences` button)
2. `Preferences` and the rest of the right panel
![kép](https://user-images.githubusercontent.com/17465872/134791025-c2e4d928-8d43-4218-b95b-c7404cb2153d.png)

1. When adding a `Class` or `Package`, the default name is `NewClass` or `NewPackage`, respectively. Can the default name please be translatable? And the strings be added to [Weblate](https://hosted.weblate.org/projects/gaphor/gaphor/)
2. The right panel
![kép](https://user-images.githubusercontent.com/17465872/134791219-6cfe63f1-41bc-4ac6-808c-0c830a9fded3.png)

1. `Gapor is the simple modeling tool written in Python`
2. `Fork me on GitHub`
3. `Copyright © 2001-2021 Arjan J. Molenaar and Dan Yeaw`
4. The About dialog heading shows `Gaphor névjegye` when it should show `A Gaphor névjegye`
![kép](https://user-images.githubusercontent.com/17465872/134791251-bab00e23-e4f0-4e40-8117-c6a0c524c13c.png)

### OS
- [X] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.6.1

### Additional information
https://github.com/gaphor/gaphor/issues/943

ovari:
1. `All Gaphor Models`
![kép](https://user-images.githubusercontent.com/17465872/134791419-38b892bc-9bff-4bb8-96cc-be99d777e3df.png)

ovari:
1. `Console` in the menu
[...]

PR #1031, opened 2021-09-26
Fix localization in UI files

amolenaar:
### PR Type

- [X] Bug fix
- [ ] Feature

Parts of the application are translated, others are not.

I found out that the translatable text from the UI files was not translated.

Issue Number: #1027 

### What is the new behavior?

We use the `locale` library. This lib is using libintl (the C library). Now translations are also picked up by `Gtk.Builder`.

- [ ] Yes
- [X] No

### Other information

Does this also work on Windows?

### DIFF
--- a/gaphor/diagram/uibuilder.py
+++ b/gaphor/diagram/uibuilder.py
@@ -1,0 +1,15 @@
+import functools
+import importlib.resources
+import xml.etree.ElementTree as etree
+
+from gaphor.i18n import gettext
+
+
+@functools.lru_cache(maxsize=None)
+def translated_ui_string(package: str, ui_filename: str) -> str:
+    with importlib.resources.path(package, ui_filename) as ui_path:
+        ui_xml = etree.parse(ui_path)
+    for node in ui_xml.findall(".//*[@translatable='yes']"):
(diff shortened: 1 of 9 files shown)


amolenaar:
I cant get the translations for Windows to work. The UI files do not translate :/

amolenaar:
It looks like Python on Windows is compiled without libintl support. So far I had no success with ctypes.

I see two ways forward:
1. Fix libintl support to the extend that it works on Windows (does GLib do something funny? Should Python be recompiled with libintl support?)
2. Fix it on Python side: read the UI files and perform transformations in the XML or parsed builder object.

gaphor/diagram/uibuilder.py
```diff
@@ -0,0 +1,15 @@
+import functools
+import importlib.resources
+import xml.etree.ElementTree as etree
+
+from gaphor.i18n import gettext
+
+
+@functools.lru_cache(maxsize=None)
+def translated_ui_string(package: str, ui_filename: str) -> str:
```
danyeaw, line 9: That was a creative way to solve this :+1:

danyeaw (approved):

PR #1050, opened 2021-10-03
Simpler services

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Help service is an "appservice". However, it depends on the current active session, to display its dialogs.

### What is the new behavior?

HelpService is an ordinary service now.

As a result the Session appservice could also be removed.

- [X] Yes - fewer appservices
- [ ] No


danyeaw (approved):
Nice simplification!

PR #1052, opened 2021-10-03
Remove importlib_metadata

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

`importlib_metadata` was used to apply Python 3.8+ features in Python 3.6/7.
We only support Python 3.9 currently.

### What is the new behavior?

Nothing changed, we use the standard library now directly.
By caching entry points we speed up unit tests by roughly 10%.

- [ ] Yes
- [X] No


danyeaw (approved):
Yay, less dependencies is :fire: 

I was getting a lock file is out of date warning, so I updated the lock file :+1:

Issue #929, opened 2021-08-20
Swap Relationship Directions from Source to Target


Gytree:
### Is your feature request related to a problem?
When we are going to create a relation, like generalization or a class dependency, currently I need to select the target object first and then the source. but in all the previous software that I used before the behavior is on the contrary, first select the source of the relation then the target.

Example if I'm going to make a class generalization the normal thinking is the class Dog extends Animal, so the logical is select the class that is going to extend the other first (Dog) and after that the extended class (Animal)

### Describe the solution you'd like
Is possible change this behavior to a selecting the source first ? or a better solution maybe will be add a setting for this behavior.

danyeaw:
@Gytree Thanks, I have noticed this as well, and I'm glad someone else pointed it out!

@all-contributors please add @Gytree for bug

danyeaw:
@Gytree I created a PR to swap the generalization direction. I don't have MagicDraw available since I'm on vacation. Do you know what other relationships are backwards?

danyeaw:
Gaphas right now defines the head end as the side of the line that is placed first. Another solution if all the relationships are backwards is to make the head and tail swappable in gaphas.

I tested things with [Papyrus](https://www.eclipse.org/papyrus/).

Class like dependencies:
- Generalization: opposite
- Dependency: opposite
- Containment: same
- Directed Association: opposite
- Extension: opposite
- Import: opposite

Others:
- Include and extend: opposite
- Transition: same
- Message: same

So what we have is inconsistent, and it wouldn't work to just allow the head and tail to be swapped in Gaphas. Should we create migrations like I did for #935 for the other relations?

amolenaar:
@danyeaw We will probably be the most flexible if we swap which node is placed first. Make it configurable in the placement tool. We can then set the direction per item and swap them if we need to.

Are Papyrus and MagicDraw in sync with their placement behavior?

danyeaw:
@amolenaar Yes, it looks like Papyrus and MagicDraw use the same arrow directions

amolenaar:
- [ ]    Generalization: opposite
- [ ]    Dependency: opposite
- [ ]  Directed Association: opposite
- [ ]    Extension: opposite
- [ ]   Import: opposite
- [ ] Include and extend: opposite

danyeaw:
@amolenaar I tested with Papyrus again. I think our directed association is backwards, did you see something different when you were testing?

We draw the arrow on the first handle, Papyrus draws the diamond on the first handle.

PR #1129, opened 2021-11-16
Swap relationship direction for Generalization, Dependency, Import, Include, and Extend

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Some relationships are drawn in an unexpected direction. E.g. Generalization is drawn from specific to general type

Issue Number: #929

### What is the new behavior?

Directions have been inverted when drawing. This should feel more natural.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/C4Model/toolbox.py
+++ b/gaphor/C4Model/toolbox.py
@@ -119,6 +119,7 @@ c4 = ToolSection(
             "gaphor-dependency-symbolic",
             "d",
             new_item_factory(uml_items.DependencyItem),
+            handle_index=0,
         ),
     ),
 )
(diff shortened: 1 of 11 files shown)


danyeaw:
What kind of magic is this to swap directions using the handle_index? 🪄 

If it is OK, I would like to go back through this and compare MagicDraw to this PR and make sure we have them correct.

amolenaar:
No magic. This is a feature already used for all box-line elements. Only I had to swap the handle positions itself to make the initial connect of the "opposite" handle work as expected.

danyeaw:
@amolenaar Should we fix directed associations and then merge this one in as well before we release 2.7.0?

amolenaar:
@danyeaw I would opt to merge this and create a separate PR that changes the orientation of Associations. Associations pop up in a lot of places in the toolbox(es), and we should change all of them.

danyeaw:
Yup, that works!

Issue #1147, opened 2021-11-22
Reconnecting a Generalization creates a second Generalization


amolenaar:
### To Reproduce
1. Open the RAAML model
2. Open diagram `Profiles.RAAML.FTA.FTA Library.Events.Events`
3. Move one of the generalization relations around
4. Save the model
5. Try to generate the model: `poe raaml`
6. See an error like this:

```
gaphor/RAAML/raaml.py:176: error: Duplicate base class "EventDef"
```

### Expected behavior
Onlt one generalization relation needs to exist.

### OS
- [x] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: master (593b0d5e)

PR #1149, opened 2021-11-23
Reuse generalization

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

When a generalization between two classes is created on two diagrams, two generalizations are created in the model

Issue Number: #1147 

### What is the new behavior?

Fix old behavior where an existing generalization is used from the model. This requires both associations (`UML.Generalization.specific` and `UML.Generalization.general`) to be bi-directional associations, so we can discover an existing . The `UML.Generalization.general` association was uni-directional and a model update has been made to make it bi-directional. This may also help traversing the model in code generation.

image

- [ ] Yes
- [x] No

### Other information

I also added a similar test for `InterfaceRealization`.

### DIFF
--- a/gaphor/UML/classes/tests/test_generalizationconnect.py
+++ b/gaphor/UML/classes/tests/test_generalizationconnect.py
@@ -1,0 +1,129 @@
+import pytest
+
+from gaphor import UML
+from gaphor.core import Transaction
+from gaphor.core.modeling import Diagram
+from gaphor.diagram.tests.fixtures import allow, connect, get_connected
+from gaphor.UML.classes.generalization import GeneralizationItem
+from gaphor.UML.classes.klass import ClassItem
+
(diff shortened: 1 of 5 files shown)


gaphor/UML/classes/tests/test_generalizationconnect.py
```diff
@@ -0,0 +1,131 @@
...
+    def _create(item_class, element_class=None):
+        return other_diagram.create(
+            item_class,
+            subject=(element_factory.create(element_class) if element_class else None),
+        )
+
+    return _create
+
+
+def test_generalization_multiple_connection(create, other_create):
```
amolenaar, line 31: This is the new test that covers this behavior.

danyeaw (approved):
Looks great!

PR #1153, opened 2021-11-26
Use Jedi for auto-complete in the Python Console

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

We use `rlcompleter`, which has two issues: auto-complete does not work smooth, and it's causing crashes on our CI machine (probably due to the readline library).

Issue Number: N/A

### What is the new behavior?

We use [Jedi](https://github.com/davidhalter/jedi) for auto-complete. In general auto-complete is a lot better now. It's easier to discover the model now.

This PR also contains a bunch of UX improvements for the console.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/plugins/console/console.py
+++ b/gaphor/plugins/console/console.py
@@ -43,11 +43,13 @@ prompt = {
 
 style = dict(
     [
-        _text_tag("banner", foreground="saddle brown"),
-        _text_tag("ps1", foreground="DarkOrchid4", editable=False),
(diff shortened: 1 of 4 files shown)


danyeaw (approved):
Nice one @amolenaar!

danyeaw:
I pushed this to a packaging branch as well, I think we should make sure macOS and Windows are working smoothly prior to merging since we are adding an extra dependency.

amolenaar:
I think this is by far the best GTK based Python REPL :laughing: .

PR #1160, opened 2021-11-27
No GTK required anymore for generating docs

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Although we did not use a GUI, GTK and GtkSourceView had to be provided as dependencies when rendering docs with Sphinx.

Issue Number: N/A

### What is the new behavior?

All you need is GI-repository and Pango for text handling.

- [ ] Yes
- [x] No

### Other information

I moved the `gi.require_versions()` commands out of `gaphor/__init__.py`. PangCairo has been moved to `gaphor.diagram` and GTK and GtkSourceView versions are set from `gaphor.ui`.

This works fine if you're just generating docs. With our app, we ran into a little trouble. Since the application itself uses `gi` and calls `gi.require_version()`, we have to either install the GTK typelibs or implement a workaround so `gi.require_version()` is not called when generating docs. I choose to create the workaround: `GAPHOR_USE_GTK=NONE` will skip setting the required versions.


danyeaw (approved):
This looks great!

Issue #452, opened 2020-09-27
Rewrite UML code generator


amolenaar:
**Is your feature request related to a problem? Please describe.**

The UML code generator (coder) is a very hard to read piece of code. This makes it hard to maintain and it's also hard to extend it. 

**Describe the solution you'd like**

I would like a rewrite of the `uml_coder` that makes full use of Gaphor's file loading capabilities (`storage`, `ElementFactory`), and from that creates a new model. This version could act as an example of how to write a (complex) code generator for Python or any other language.

The Profile coder can act as a starting point. Maybe it's sufficient to extend it.

**Additional context**

One of the hard things in the current generator is that derived unions are defined in one go. It would be nicer to allow extension of derived unions. E.g. the Core model can define the `owner-ownedElement` derived union and UML, SysML, etc. can add "subsets" to that derived union.

amolenaar:
Implemented by #1170.

PR #1168, opened 2021-12-01
New code generator for UML

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

* Our current code generator is in a state where it's almost impossible to change.
* Our current code generator uses its own parser to parse the model. We can use the `storage` module and `ElementFactory` for that.
* Our current code generator is not a nice example of how to write a code generator

Issue Number: #452

### What is the new behavior?

The new code generator is more clean.
In addition it allows for relations to be added to derived unions (instead of declaring it at once). This feature we need if we want to add derived values over models.

The new coder is built so we can reuse it for SysML and RAAML as well.

Black and isort have been disabled. Some Flake8 warnings are silenced. This ensures the coder output passes pre-commit without need for (manual) change.
 
- [ ] Yes
- [x] No


amolenaar:
Superseded by #1170.

PR #1181, opened 2021-12-06
Core as a separate ModelingLanguage

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Currently the Core modeling language is accessed via the UML modeling language. (e.g. for loading a model)

### What is the new behavior?

The Core Modeling Language has its own `ModelingLanguage`, to obtain types like `Diagram` and `Presentation`.
No extra imports need to be done in `gaphor/UML/uml.py`.

All imports for `Comment`, `Diagram`, `Presentation`, and `StyleSheet` have to be done on the core model. Hence the many updated test cases.

We do not want to show the Core language in the dropdown menu, so I made a workaround by proving an empty name (`""`).

Previously, for most tests we could get away with loading the UML language. Now we need a combined Core+UML language. Hence the new `MockModelingLanguage` class. I'm not sure it's the right name, though.

- [x] Yes
- [ ] No

### Other information

I like the idea that Core is a standalone language now. It's no longer dependent on UML, which it was for historical reasons.

On the other hand, it was convenient to know that everything we need is in the UML model. and the workaround with the Core ML name is not ideal.

### DIFF
--- a/gaphor/C4Model/tests/conftest.py
+++ b/gaphor/C4Model/tests/conftest.py
@@ -1,28 +1,18 @@
 import pytest
 
 from gaphor.C4Model.modelinglanguage import C4ModelLanguage
-from gaphor.core.modeling import ElementFactory
-from gaphor.core.modeling.elementdispatcher import ElementDispatcher
-from gaphor.diagram.tests.fixtures import diagram, event_manager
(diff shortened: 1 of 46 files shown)


danyeaw (approved):
@amolenaar I like separating out the core model as well. Everything looks good to me!

Issue #1185, opened 2021-12-07
Static main method with array parameter not set.


jcbritobr:
### Describe the bug
When trying to put a main static method with a string array in class, gaphor change it to simple variable

### To Reproduce
1 - Seclect a class
2 - Create a method with static mark
3 - Set "+ main(args: String[]): void"

### Expected behavior
A method with + main(args: String[]): void in class

### Screenshots
![image](https://user-images.githubusercontent.com/6579862/145098508-98961e70-3dc7-42ce-a6d9-6b986f702966.png)

### OS
Windows

### Version
Version of Gaphor: 2.7.1

amolenaar:
Hi,

The official syntax would be `main(args: String[*])`. The items between square brackets is the multiplicity. I can see this is confusing, since in most programming languages `[]` denotes an array of unspecified size. It would probably be better to interpret `[]` as `[*]`. What do you think?

See also https://www.uml-diagrams.org/operation.html.

jcbritobr:
> Hi,
> 
> The official syntax would be `main(args: String[*])`. The items between square brackets is the multiplicity. I can see this is confusing, since in most programming languages `[]` denotes an array of unspecified size. It would probably be better to interpret `[]` as `[*]`. What do you think?
> 
> See also https://www.uml-diagrams.org/operation.html.

Agreed and it works. I will close the issue here. :)

amolenaar:
Never the less. I think Gaphor can help us by interpreting `[]` as `[*]`.

PR #1186, opened 2021-12-07
Empty square bracket notation

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

For an attribute or parameters, it's not possible to have an undefined multiplicity: `attr[]`, `myfunc(arg: string[]): result[]`. The brackets `[]` are not parsed.

However, `[]` is a common construct in programming languages to define an unbounded array. We can simply help the user here by interpreting `[]` as `[*]`.

Issue Number: #1185

### What is the new behavior?

Attributes and parameters with undefined multiplicity are treated as if it has a `*` multiplicity. `attr[]` will parse and render as `attr[*]`.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/UML/umllex.py
+++ b/gaphor/UML/umllex.py
@@ -23,7 +23,7 @@ derived_subpat = r"\s*(?P<derived>/)?"
 name_subpat = r"\s*(?P<name>[a-zA-Z_]\w*( +\w+)*)"
 
 # Multiplicity ::= '[' [mult_l ..] mult_u ']'
-mult_subpat = r"\s*(\[\s*((?P<mult_l>[0-9]+)\s*\.\.)?\s*(?P<mult_u>([0-9]+|\*))\s*\])?"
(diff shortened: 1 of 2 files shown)


gaphor/UML/umllex.py
```diff
@@ -23,7 +23,7 @@
 name_subpat = r"\s*(?P<name>[a-zA-Z_]\w*( +\w+)*)"
 
 # Multiplicity ::= '[' [mult_l ..] mult_u ']'
-mult_subpat = r"\s*(\[\s*((?P<mult_l>[0-9]+)\s*\.\.)?\s*(?P<mult_u>([0-9]+|\*))\s*\])?"
+mult_subpat = r"\s*(?P<has_mult>\[\s*((?P<mult_l>[0-9]+)\s*\.\.)?\s*(?P<mult_u>([0-9]+|\*))?\s*\])?"
```
danyeaw, line 26: These regex are wild :magic_wand:

danyeaw (approved):
Looks good!

gaphor/UML/umllex.py
amolenaar: @danyeaw Pure magic, with a *lot* of testing :wink:.

PR #1194, opened 2021-12-11
Ensure model consistency on save

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

We've had a couple of issues in the past where saved models could not be loaded. Although I think this issue has not reoccurred since the model has been simplified (diagram items are now fully part of the model), it's good to do an extra consistency check.

### What is the new behavior?

While saving a model, every reference is checked against the model to ensure it is resolvable -- the element it's referring to will be saved as well.

References to elements not part of the model are skipped and a warning is printed.

- [ ] Yes
- [x] No

### Other information

I'm not sure if we should bother the user with this (e.i. show an in-app notification/toaster).


danyeaw (approved):
Nice cleanup, simplification, and extra checks!

Issue #1189, opened 2021-12-08
Grow container when dropping an element


amolenaar:
### Is your feature request related to a problem?

When you drop an item ("child") on another item ("parent"), the child may partly be placed outside the boundaries of the parent item.

In this example, the Component is dropped on the code:
![image](https://user-images.githubusercontent.com/96249/145290431-47bee455-924b-4954-bde5-9739bceded14.png)

### Describe the solution you'd like

When you drop an item on another, make the parent item grow so it "contains" the child item.

![image](https://user-images.githubusercontent.com/96249/145290505-9c1b054e-3866-46ee-bde4-47daa66d8959.png)

### Describe alternatives you've considered

The alternative, current implementation, is to make users resize the parent item by themselves.

PR #1203, opened 2021-12-12
Grow item when an item is dropped on it

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Items can be dropped inside other items (like a Component in a Node), but the component will most likely be partly outside the node, since the node does not adapt its size.

Issue Number:  #1189

### What is the new behavior?

Resize the node whenever a (new) item is dropped on it.

- [ ] Yes
- [x] No


danyeaw (approved):
Nice enhancement @amolenaar!

PR #1207, opened 2021-12-14
Notification when model elements are automatically removed

### LINKED ISSUE #971: Add Option and Dialog to Prevent Automatic Removal of Last Usage of an Element When the Diagram is Deleted
As a user I would like to be able to create a model element in the model (by pressing a button, or right mouse click, etc. and selecting from a list), without the need of creating a new diagram.

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

When you remove a diagram with diagram items, model elements that are no longer presented in any diagram will be removed. A dialog is shown (somewhere on screen) with a message.

Fixes #971.

### What is the new behavior?

Instead of a dialog, an in-app notification is shown. An option has been added to the Preferences pane to toggle automatic deletion of no longer presented model elements.

When a diagram is deleted with items, a notification is shown. The notification is also shown if an item is removed on a diagram.

- [ ] Yes
- [x] No

### Other information

This solution leverages the solutions already present in Gaphor: in-app notifications and our simple preferences pane.
If items are deleted by accident, they can simply be undone.

Note that the message does not tell *which* items have been removed. This is partly due to where the message is constructed and partly because it's likely to break localization.

### DIFF
--- a/gaphor/UML/sanitizerservice.py
+++ b/gaphor/UML/sanitizerservice.py
@@ -8,6 +8,8 @@ from gaphor.core import event_handler
 from gaphor.core.modeling import Diagram, Element, Presentation
 from gaphor.core.modeling.event import AssociationDeleted, AssociationSet, DerivedSet
 from gaphor.diagram.general import CommentLineItem
+from gaphor.event import Notification
+from gaphor.i18n import gettext
 
 
 def undo_guard(func):
@@ -58,6 +60,14 @@ class SanitizerService(Service):
             old_subject = event.old_value
             if old_subject and not old_subject.presentation:
                 old_subject.unlink()
+
+                self.event_manager.handle(
+                    Notification(
+                        gettext(
+                            "Model element(s) have been removed: they're not shown in a diagram anymore."
+                        )
+                    )
+                )
 
     @event_handler(AssociationSet)
     @undo_guard
--- a/gaphor/UML/sanitizerservice.py
+++ b/gaphor/UML/sanitizerservice.py
@@ -70,7 +70,7 @@ class SanitizerService(Service):
             self.event_manager.handle(
                 Notification(
                     gettext(
-                        "Dangling model element(s) have been removed: they're not shown in a diagram anymore."
+                        "Removed unused elements from the model: they are not used in any other diagram."
                     )
                 )
             )
--- a/gaphor/UML/sanitizerservice.py
+++ b/gaphor/UML/sanitizerservice.py
@@ -30,8 +30,9 @@ class SanitizerService(Service):
     """Does some background cleanup jobs, such as removing elements from the
     model that have no presentations (and should have some)."""
 
-    def __init__(self, event_manager, undo_manager=None):
+    def __init__(self, event_manager, undo_manager=None, properties=None):
         self.event_manager = event_manager
+        self.properties = properties or {}
         self.undo_manager = undo_manager
 
         event_manager.subscribe(self._unlink_on_subject_delete)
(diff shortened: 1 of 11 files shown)


gaphor/UML/sanitizerservice.py
```diff
@@ -54,10 +57,23 @@ def _unlink_on_subject_delete(self, event):
...
+            return
+
+        old_subject = event.old_value
+        if old_subject and not old_subject.presentation:
+            old_subject.unlink()
+
+            self.event_manager.handle(
+                Notification(
+                    gettext(
+                        "Dangling model element(s) have been removed: they're not shown in a diagram anymore."
```
amolenaar, line 73: Maybe the wording can be improved.

danyeaw: What about: Since you deleted the last usage of elements in a diagram, they  were automatically removed from the model.

danyeaw (approved):
This is nice and clean solution, I really like it!

gaphor/UML/sanitizerservice.py
amolenaar: That sounds friendly. Let me try that.

image

Can it be a little shorter?

"You deleted the last usage of elements, they are removed from the model too."?

"You deleted the last usage of elements in a diagram, they are removed from the model too."?

danyeaw: I like the last one, I think it is short and clear.

amolenaar: "Removed model elements no longer in use in a diagram." it is then.

amolenaar:
@danyeaw The preferences item should also be renamed, I think. Since the term "Dangling" has no relation with the message:

image

Would "Remove unused elements" cover it? It's only about removing the elements at the point their last appearance in a diagram has been removed.

gaphor/UML/sanitizerservice.py
sz332: As I user i would like to understand what has happened in the system. The description "Removed model elements no longer in use in a diagram" is something I don't understand as a user.

"The deleted element was removed from the model as well because it was not used in any other diagram"?

danyeaw:
> Would "Remove unused elements" cover it? It's only about removing the elements at the point their last appearance in a diagram has been removed.
> 
> We can also add a tooltip to explain this feature somewhat.

Yup, that sounds great.

sz332:
> 

Yes, I agree. Remove unused elements is a great name for the properties. The notification message is something we need to work on a little bit I think.

gaphor/UML/sanitizerservice.py
amolenaar: The message should ideally be short, since it disappears after a couple of seconds and you need time to read it. I agree that @sz332's and @danyeaw's initial phrasing is more self explaining (less cryptic if you like).

This message occurs in two situations:
1. An element has been removed that's no longer in any diagram
2. A diagram has been removed that contains elements that are now no longer present in any diagram.

amolenaar:
@danyeaw @sz332 How 'bout:

`Removed unused elements from the model: they are not used in any other diagram.`

It relates to the preference _Remove Unused Elements_, and is relatively short.

danyeaw:
@amolenaar Sounds good to me!

sz332:
Sounds great!

Dan Yeaw ***@***.***> ezt írta (időpont: 2021. dec. 21., K
4:38):

amolenaar:
Done

danyeaw:
Nice one @amolenaar :1st_place_medal:

Issue #1012, opened 2021-09-18
Add new action: pusher and magnet to toolbar


sz332:
### Is your feature request related to a problem?

In case of working with a big diagram, making space is not always easy with selecting items, and moving them around.

### Describe the solution you'd like

MagicDraw has a very useful tool for this problem, called pusher/magnet. (Visual Paradigm calls this sweep/magnet). It works the following way:

Pusher (make space):

- User selects pusher from the toolbar
- User presses left mouse button and keeps it pressed, and
- Starts moving the mouse to the right
  - A vertical line is shown
  - Every element to the right of the vertical line are pushed to the right (farther from each other)
- Starts moving the mouse downwards
  - A horizontal line is shown
  - Every element below of the horizontal line are moved downwards (farther from each other)

Magnet (remove space):

- User selects magnet from the toolbar
- User presses left mouse button and keeps it pressed, and
- Starts moving the mouse to the left
  - A vertical line is shown
  - Every element to the right of the vertical line are pushed to the left (closer to each other)
- Starts moving the mouse upwards
  - A horizontal line is shown
  - Every element below of the horizontal line are moved upwards (closer to each other)

sz332:
https://user-images.githubusercontent.com/8182138/135477321-bc31a1c6-d321-48ad-8c09-5b5e3dfe50c0.mp4

danyeaw:
Maybe you could combine it if you always made right to left and down to up close space (magnet), and the opposite open space (push). I am not sure how intuitive that would be though since I have only played with them as separate tools, but might be worth a try to play with.

danyeaw:
Closed by #1208 which includes pusher and magnet support :+1:

PR #1208, opened 2021-12-14
A Magnet-tool

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

You'll have to select all elements using rubberband selection and move them around.

Issue Number: #1012 

### What is the new behavior?

A simple magnet tool that pulls/pushes all elements on the right/bottom of the bar.

- [ ] Yes
- [x] No


danyeaw:
The guides are getting stuck on for me after I use the magnet tool:
image

Should we create a new icon for this instead of the GNOME foot?

amolenaar:
I created a magnet 🧲 this morning. I have to look into the guide lines. Also, if there are no item on the right/bottom the bar is not showing. I think it should. I'm also wondering if the tool can give better visual hints of what's happening. Maybe gray out the left/top side?

danyeaw:
@amolenaar I like the icon! I also noticed that the tool isn't resetting back to the pointer.

sz332:
@danyeaw  Shall it reset back the pointer?

danyeaw:
> @danyeaw  Shall it reset back the pointer?

Yes, I think so, as long as the option to automatically reset is set.

amolenaar:
> > @danyeaw  Shall it reset back the pointer?
> 
> Yes, I think so, as long as the option to automatically reset is set.

I'm not sure how easy that is to do, since the reset functionality is made to trigger when a new item has been created, not after some tool has been executed.

amolenaar:
Okay, tools are reset and the guides are removed after the drop is finished.

danyeaw:
:shipit: Nice job @amolenaar!

Issue #841, opened 2021-05-29
Allow Nesting and Stereotypes for Properties


danyeaw:
**Is your feature request related to a problem? Please describe.**
With our initial SysML support we allowed for the creation of part and reference properties and then to display them as part of an Internal Block Diagram. Next it would be great if we allowed the properties to be nested. We also need to be able to display Stereotypes for Controller, Actuator, and ControlledProcess.

For example, if a Brake System Control Unit (BSCU) contains an Autobrake Controller and Hydraulic Controller:
![image](https://user-images.githubusercontent.com/10014976/120086310-4f2a1d00-c0ac-11eb-9014-a91782d01a04.png)

How I would like to be able to create a nested Internal Block Diagram that showing that the BSCU contains these things:
![image](https://user-images.githubusercontent.com/10014976/120086401-957f7c00-c0ac-11eb-8e64-fc6b6301d972.png)

Right now we can drag a property on top of another, but the relationship (composition or aggregation) isn't taken into account. Also the name of the property is centered in the property which makes this difficult. 

**Describe the solution you'd like**
When a property is selected, allow sub-properties (part or reference) to be displayed from the propertiespages.

When a sub-property is displayed, move the name and stereotype of the box back to the top and create a new compartment.

**Describe alternatives you've considered**
Move the name and stereotype to the top and keep things as they are with no nested properties.

**Additional context**

amolenaar:
Somewhat related to #897.

amolenaar:
@danyeaw Is this about nesting `Property`'s or `Block`'s? The first diagram shows blocks, but the text talks about properties.

We can already nest `Property` in `Block`, but not `Property` in `Property` and `Block` in `Block`.

`Block` in `Block` should be pretty straightforward. I'm not sure how `Property` in `Property` would be possible.

danyeaw:
@amolenaar The first diagram shows that the BSCU contains an Autobrake Controller and a Brake Controller. Then in an IBD you would like to be able to show that as well without redrawing it from scratch. You would drag the BSCU to your IBD to create a part property of the BSCU (a usage of if), then you want to be able to select the BSCU and show what is inside the BSCU (the AutoBrake Controller and Brake Controller).

If the first diagram was showing aggregation instead of composition, you would still want to be able to display those in an IBD as well except as a reference property.

amolenaar:
[...]

PR #1211, opened 2021-12-16
Stereotypes for ItemFlow properties

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: #841

### What is the new behavior?

Add example for car braking system.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/SysML/propertypages.py
+++ b/gaphor/SysML/propertypages.py
@@ -1,9 +1,9 @@
 from gaphor import UML
 from gaphor.core import transactional
-from gaphor.core.format import format, parse
 from gaphor.diagram.propertypages import (
     PropertyPageBase,
     PropertyPages,
+    combo_box_text_auto_complete,
     new_resource_builder,
 )
 from gaphor.SysML import sysml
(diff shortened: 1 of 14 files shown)


amolenaar:
@danyeaw 

Should stereotyped types of `itemProperty` show like this?

image

danyeaw (commented):
Controller, Feedback, and Control Action are already defined in the RAAML profile under the STPA section. For the sysml-brake-system example, did we need to redefine them in a new profile?

gaphor/SysML/propertypages.py
```diff
@@ -206,24 +215,43 @@ def _on_item_flow_active(self, switch, gparam):
             subject.informationFlow = iflow
             iflow.informationSource = subject.end[0].role
             iflow.informationTarget = subject.end[1].role
-            iflow.itemProperty = subject.model.create(sysml.Property)
+            iflow.itemProperty = subject.model.create(UML.Property)
```
danyeaw, line 218: Why the UML Property instead of the sysml one?

amolenaar: It's the same. `sysml` imports `Property` from `gaphor.UML`

amolenaar:
> Controller, Feedback, and Control Action are already defined in the RAAML profile under the STPA section. For the sysml-brake-system example, did we need to redefine them in a new profile?

The default stereotypes are currently not picked up (they are only defined in the diagram item). Only the ones defined in the model. It would take a little more work to also display the default stereotypes (like `primitive` in the image above). It may be useful for other purposes to have a (single-dispatch) function that provides an optional stereotype name.

gaphor/SysML/propertypages.py
danyeaw: Ya, I get that they are the same. It just seems clearer to me that the SysML profile uses SysML elements.

danyeaw:
> The default stereotypes are currently not picked up (they are only defined in the diagram item). Only the ones defined in the model.

I am concerned if users have to redefine elements already in the RAAML profile, then we aren't really solving the original issue.

Do you think we should enhance that for at least Controller, Feedback, and Control Action? Should they only show when we have the RAAML profile selected?

amolenaar:
> I am concerned if users have to redefine elements already in the RAAML profile, then we aren't really solving the original issue.

That would be weird, actually ;).

> Do you think we should enhance that for at least Controller, Feedback, and Control Action? 

We can do that, for starters. And come up with a more generic solution later, matching the "default" stereotype on classifiers.

> Should they only show when we have the RAAML profile selected?

I don't think we should do that. The profile is in no way related to the diagram.

amolenaar:
Like this?

image

danyeaw:
@amolenaar Yes, that looks great! How do I assign the Controller stereotype to the part autobrakeController and hydraulicController part properties?

amolenaar:
We don't have those items in our toolbox currently. I think we should add them.

amolenaar:
@danyeaw Is this ready for mainline?

PR #1214, opened 2021-12-17
Allow allocation of structural types to swimlane partitions

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Swimlanes can only have a name.

Issue Number: #426 

### What is the new behavior?

Swimlanes can be assigned a type, via a drop-down in the property editor. Swimlane types are restructed to UML Classifiers (Classes, Interfaces, SysML Blocks, etc..

image

- [ ] Yes
- [x] No

### Other information

I'll leave out the option to DnD a block on a swimlane for now.

### DIFF
--- a/gaphor/UML/actions/partitionpage.py
+++ b/gaphor/UML/actions/partitionpage.py
@@ -1,5 +1,7 @@
 """Activity partition property page."""
 
+from gi.repository import Gtk
+
 from gaphor import UML
 from gaphor.core import gettext, transactional
 from gaphor.diagram.propertypages import (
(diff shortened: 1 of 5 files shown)


gaphor/UML/actions/propertypages.glade
```diff
@@ -248,54 +346,48 @@ renderers and such.</property>
...
-      <object class="GtkTreeView" id="partition-treeview">
-        <property name="visible">True</property>
-        <property name="can-focus">True</property>
-        <child internal-child="selection">
-          <object class="GtkTreeSelection"/>
+        <child>
+          <object class="GtkLabel">
+            <property name="visible">True</property>
+            <property name="can-focus">False</property>
+            <property name="label" translatable="yes">Type</property>
```
danyeaw, line 364: I think it would be clearer if we called it Allocated Type

gaphor/UML/actions/partitionpage.py
```diff
@@ -23,60 +23,74 @@ class PartitionPropertyPage(PropertyPageBase):
...
-        renderer_editable_text.set_property("editable", True)
+        builder.get_object("partition-name").set_text(partition.name or "")
 
-        treeview.append_column(
-            Gtk.TreeViewColumn(
-                title=gettext("Name"), cell_renderer=renderer_editable_text, text=1
-            )
-        )
+        combo = builder.get_object("partition-type")
+        for c in self.item.model.select(UML.Classifier):
```
danyeaw, line 58: I think we should limit this to Blocks and Classes. Being able to select an Activity to allocate an Activity Partition to isn't quite right, since the partition should be allocated to a structural element, not behavioral.

danyeaw (commented):
I like this functionality from the property page :+1:

gaphor/UML/actions/partitionpage.py
amolenaar: I think it makes sense to stick with classifiers, since that includes interfaces too. We can exclude behaviors.

danyeaw: This looks good now :+1:

gaphor/UML/actions/propertypages.glade
danyeaw: Thanks for the update!

danyeaw:
Should we expand the header vertically if the name + type doesn't fit?
image

amolenaar:
> Should we expand the header vertically if the name + type doesn't fit?

Nice find. I'll have a look. It shouldn't be that hard.

amolenaar:
image

Voila

danyeaw (approved):
This looks outstanding @amolenaar!

PR #1216, opened 2021-12-18
Diagram types

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

We only had one diagram type. The request has come up several times for specific diagrams. 

Fixes ##973.

### What is the new behavior?

Apart from the generic type, modeling language specific modeling types can be selected.

- [x] Language specific diagram types show in the New Diagram dropdown.
- [x] Diagram have a label in the top-left, indicating the diagram type.
- [x] Toolbox is updated to reflect the diagram type.
- [x] Diagram types for:
  - [x] UML
  - [x] SysML
  - [x] RAAML
  - [x] C4 Model
- [ ] How to distinguish different diagram types in the tree view?
 
- [x] Yes
- [ ] No

### DIFF
--- a/gaphor/SysML/toolbox.py
+++ b/gaphor/SysML/toolbox.py
@@ -65,12 +65,12 @@ sysml_toolbox_actions: ToolboxDefinition = (
 
 # Not implemented: Parameter Diagram
 sysml_diagram_types: DiagramTypes = (
-    DiagramType("act", gettext("Activity Diagram"), (actions,)),
-    DiagramType("blk", gettext("Block Definition Diagram"), (blocks,)),
-    DiagramType("ibl", gettext("Internal Block Diagram"), (internal_blocks,)),
-    DiagramType("pkg", gettext("Package diagram"), (blocks,)),
-    DiagramType("req", gettext("Requirement diagram"), (requirements,)),
-    DiagramType("sd", gettext("Sequence Diagram"), (interactions,)),
-    DiagramType("stm", gettext("State Machine Diagram"), (states,)),
-    DiagramType("uc", gettext("Use Case Diagram"), (use_cases,)),
+    DiagramType("act", gettext("New Activity Diagram"), (actions,)),
+    DiagramType("blk", gettext("New Block Definition Diagram"), (blocks,)),
+    DiagramType("ibl", gettext("New Internal Block Diagram"), (internal_blocks,)),
+    DiagramType("pkg", gettext("New Package Diagram"), (blocks,)),
+    DiagramType("req", gettext("New Requirement Diagram"), (requirements,)),
+    DiagramType("sd", gettext("New Sequence Diagram"), (interactions,)),
+    DiagramType("stm", gettext("New State Machine Diagram"), (states,)),
+    DiagramType("uc", gettext("New Use Case Diagram"), (use_cases,)),
 )
--- a/gaphor/SysML/toolbox.py
+++ b/gaphor/SysML/toolbox.py
@@ -65,8 +65,8 @@ sysml_toolbox_actions: ToolboxDefinition = (
 
 # Not implemented: Parameter Diagram
 sysml_diagram_types: DiagramTypes = (
-    DiagramType("blk", gettext("New Block Definition Diagram"), (blocks,)),
-    DiagramType("ibl", gettext("New Internal Block Diagram"), (internal_blocks,)),
+    DiagramType("bdd", gettext("New Block Definition Diagram"), (blocks,)),
+    DiagramType("ibd", gettext("New Internal Block Diagram"), (internal_blocks,)),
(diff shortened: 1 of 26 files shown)


amolenaar:
I added a little info box to all non-generic diagram types, so you know what you're looking at:

Screenshot from 2021-12-19 21-09-38

danyeaw (commented):
I'm still reviewing, but things look superb so far!

Should we also rename the new diagram to match the diagram type, like "New activity diagram" instead of "New model diagram"?

amolenaar:
@danyeaw @sz332 Any idea on how to display the diagram type in the tree view.

@danyeaw I updated the text of the labels.

Also: are there any standard diagram types for RAAML? Does it make sense to have diagram types for C4?

danyeaw:
> @danyeaw @sz332 Any idea on how to display the diagram type in the tree view.

We could create a diagram icon for each diagram type. I can help with this if you want.

> @danyeaw I updated the text of the labels.

Looks great, thanks!

> Also: are there any standard diagram types for RAAML? Does it make sense to have diagram types for C4?

RAAML doesn't specify any, but I think we will have some like a Hazard Analysis table once we add table support.

amolenaar:
> We could create a diagram icon for each diagram type. I can help with this if you want.

That'll be a lot of icons ;). I was more thinking about the diagram icon with the diagram abbreviation in the bottom. Haven;t figured out how to pull that off, though.

> RAAML doesn't specify any, but I think we will have some like a Hazard Analysis table once we add table support.

I added special diagrams for FTA and STPA. Should I revert those?

danyeaw:
We don't have to do it like them, but this is how one of the commercial tools does it. The diagram is always the leaf of the tree in this picture:

image

Notice that the UML and SysML diagrams have slightly different color pallete which differentiates the diagrams even if the SysML diagram type is the same. I think this causes a lot of confusion to new users to be honest.

amolenaar:
We can do the icons in a separate PR, maybe? The icons you show have a great deal of detail. I think too much.

sz332:
I think it would be nice to have a customizable icon set so that users can use whatever icons their please while providing a default one. What do you think?

amolenaar:
I'd like to keep the level of customization to a minimum. It'll keep our work a lot simpler, since we only have to test a single configuration. It's easier to spot and reproduce errors, etc.

danyeaw:
@amolenaar I am not seeing diagram types for RAAML (FTA / STPA), am I missing them?

amolenaar:
@danyeaw Sorry, forgot to push :see_no_evil:

gaphor/UML/toolbox.py
```diff
@@ -19,3 +25,16 @@
...
+    DiagramType("cls", gettext("New Class Diagram"), (classes,)),
+    DiagramType("pkg", gettext("New Package Diagram"), (classes,)),
+    DiagramType("cmp", gettext("New Component Diagram"), (components,)),
+    DiagramType("dep", gettext("New Deployment Diagram"), (components,)),
+    DiagramType("act", gettext("New Activity Diagram"), (actions,)),
+    DiagramType("sd", gettext("New Sequence Diagram"), (interactions,)),
+    DiagramType("com", gettext("New Communication Diagram"), (interactions,)),
+    DiagramType("stm", gettext("New State Machine Diagram"), (states,)),
+    DiagramType("uc", gettext("New Use Case Diagram"), (use_cases,)),
+    DiagramType("prf", gettext("New Profile Diagram"), (profiles,)),
```
danyeaw, line 39: I don't think UML has an abbreviation for Class Diagrams or Profile Diagrams, and I think many of the tools just write it out. I think it makes sense, though, using an abbreviation for them like you did :+1:

danyeaw (commented):
This is such a huge capability you added to the tool, really nice job!

Should adding some distinguishing characteristics for the tree view be part of a future PR?

gaphor/SysML/toolbox.py
```diff
@@ -59,3 +61,16 @@
     states,
     use_cases,
 )
+
+
+# Not implemented: Parameter Diagram
+sysml_diagram_types: DiagramTypes = (
+    DiagramType("blk", gettext("New Block Definition Diagram"), (blocks,)),
+    DiagramType("ibl", gettext("New Internal Block Diagram"), (internal_blocks,)),
```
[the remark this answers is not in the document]
amolenaar: fixed.

gaphor/UML/toolbox.py
amolenaar: It's kinda weird to have a full name, where in all other diagram types an abbreviation is used.

amolenaar:
> This is such a huge capability you added to the tool, really nice job!

Thanks :)

> Should adding some distinguishing characteristics for the tree view be part of a future PR?

I think it should. It would be nice to integrate this with the templates support and provide the right types of diagrams for the different levels in a model.

Issue #1157, opened 2021-11-26
Add predefined viewpoints for the C4 model


vanillajonathan:
### Is your feature request related to a problem?
A user opens Gaphor and changes profile to the C4 model.
The user now is unclear on how to proceed, on what to do.

### Describe the solution you'd like
Add the 4 predefined C4 viewpoints.
- Level 1: Context diagrams
- Level 2: Container diagrams 
- Level 3: Component diagrams
- Level 4: Code diagrams

So that the user can quickly get a grasp of C4 and have a starting point.

amolenaar:
A similar issue has come up with SysML, where you want different layers defined in the starting image. We probably need something like templates to accommodate different modeling styles.

vanillajonathan:
Yes, templates would be give a huge increase in usability by providing guidance and laying the foundation to guide proper usage of the modeling language,

amolenaar:
I've been looking for examples. I think the create-project plugin of GNOME Builder is a nice example.

![image](https://user-images.githubusercontent.com/96249/144808702-86544884-aaba-4159-b0e6-47313c33969d.png)


* https://gitlab.gnome.org/GNOME/gnome-builder/-/blob/main/src/plugins/create-project/gbp-create-project-template-icon.ui
* https://gitlab.gnome.org/GNOME/gnome-builder/-/blob/main/src/plugins/create-project/themes/shared.css

amolenaar:
... How are we dealing with templates? Will we define a fix numer of templates in Gaphor? Can we make it extensible by users, and if so, how?

danyeaw:
Probably the simplest way to do that is we could publish gaphor models with some initial diagrams and views setup so that users just have to go and edit them. We could ship the models with gaphor, or make them available separate on our website. This wouldn't require any changes to Gaphor to make it happen, only some modeling.

vanillajonathan:
@amolenaar I only think it is important with a set of predefined templates for the standards modeling notations.

I personally don't think they need to be extensible by users. The standards already exist, end-users have no need to edit templates, just needs to to pick a notation and be confined to the scope of that notation.

amolenaar:
Any ideas for what templates we might need?

1. C4 model, with packages as defined above
2. Something SysML-ish?
3. Does RAAML have it's own template?

vanillajonathan:
[...]

PR #1230, opened 2021-12-24
Improve the welcoming experience

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

When Gaphor is started, it shows an empty model.

Issue Number: #1157 

### What is the new behavior?

When gaphor is started, we open with a "Greeter", where you can pick a recently opened file (and later a new file from a Template/Scaffold).

The experience is based on [GNOME Builder](https://wiki.gnome.org/Apps/Builder). You're welcomed with your recently opened files in big button-ish things. There is an option to start a new model or open an existing model.

- [x] Remove "Open" button and add an entry in the hamburger dropdown.
- [x] Do no longer replace the model in an "empty" window.
- [x] Show something useful when there are no recent files.
- [ ] Test experience on Windows and macOS.

NB. This PR does not contain model templates yet.

- [x] Yes
- [ ] No

### Other information

Models are now displayed by their name, without extension.

image

The main window looks cleaner now:

image


danyeaw (approved):
I really like this @amolenaar, you blew me away, nice one!

I tested it on Windows as well 👍

gaphor/ui/mainwindow.py
```diff
@@ -245,8 +212,8 @@ def set_title(self):
 
         if self.filename:
             p = Path(self.filename)
-            title = p.name
-            subtitle = str(p.parent).replace(HOME, "~")
+            title = p.stem
+            subtitle = str(p).replace(HOME, "~")
```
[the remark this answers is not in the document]
amolenaar: The same argument goes for unexperienced Linux/macOS users.

I think they'll figure it out. It's a lot shorter than `C:\Users\myusername` at least ;)

What would be the alternative? `Development/test/testmodel.gaphor` or start it with ellipses instead?

danyeaw:
Ya, I think it is fine. I don't think there is really a good equivalent in Windows since I doubt most Windows users know what `%USERPROFILE%` is and that would be ugly to put as a path in the GUI. At least PowerShell understands ~, so maybe it is becoming more familiar. :+1:

Where are the recently used models stored?

amolenaar:
> Where are the recently used models stored?

The recent files are stored by GTK somewhere. This way it integrates well with GNOME.

danyeaw:
> The recent files are stored by GTK somewhere. This way it integrates well with GNOME.

Got it, looks like removing ~/.local/share/recently-used.xbel does the trick to reset it so that I can see the welcome screen.

gaphor/ui/greeter.glade
```diff
@@ -0,0 +1,191 @@
...
+        <property name="can-focus">False</property>
+        <property name="label" translatable="yes">&lt;big&gt;&lt;b&gt;Welcome to Gaphor&lt;/b&gt;&lt;/big&gt;
+
+It's time to create your first model
+or open an existing model.
+
+Happy modeling!
+
+• &lt;a href="https://gaphor.org"&gt;Gaphor homepage&lt;/a&gt;
+• &lt;a href="https://github.com/gaphor"&gt;GitHub&lt;/a&gt;
```
danyeaw, line 99: Did you mean to link this to the organization instead of the repo?

amolenaar: Yes, that was intentional :)

PR #1234, opened 2021-12-27
Start a new model from a template

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

You always start with a bare model (package + diagram).

Issue Number: #1157 

### What is the new behavior?

A user was already presented with a welcome dialog (#1230). The welcome dialog now contains templates to start a model from.

image

Templates need to be added to the `TEMPLATES` list in `gaphor/ui/greeter.py`. The models are loaded from `gaphor/templates`.

- [x] Set the modeling language from the template.
- [x] Translate texts (mostly names) in template models.

- [x] Yes
- [ ] No

API changes.

### Other information

The file manager has been extended to support the loading of a model without setting the filename. This way we can load templates.


danyeaw (approved):
This looks great!

We can work to add and update templates overtime. I contacted OMG about us [getting permission](https://www.omg.org/legal/tm_guidelines.htm) to use the UML, SysML, and RAAML trademarks and permission to use their official logos for those languages as well. We can then use those on the select template screen.

amolenaar:
> We can work to add and update templates overtime. I contacted OMG about us [getting permission](https://www.omg.org/legal/tm_guidelines.htm) to use the UML, SysML, and RAAML trademarks and permission to use their official logos for those languages as well. We can then use those on the select template screen.

Excellent! I completely forgot about that.

Issue #1247, opened 2021-12-31
Delete model element does not work


sz332:
### Describe the bug
Existing model elements cannot be removed from the model.

### To Reproduce
1. Create a new project (generic)
2. Add a class
3. Right mouse click in the navigation, delete is disabled

### Expected behavior
I expect to remove any element from the model

### OS
- [X] Linux
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: git latest

amolenaar:
This behavior was intentional: you can not remove an item that's still part of a diagram.

With the recent changes and the ability to disable the "Remove Unused Elements" behavior, we have to address this as well.

The simple solution is to just allow deletion. This will automatically trigger the deletion of element presentations in diagrams.

PR #1250, opened 2021-12-31
Allow for deleting elements in the tree view

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

It is not possible to simply delete an element from the tree view.

Issue Number: #1247

### What is the new behavior?

All Elements can be removed from the tree view. Exception rules can be defined where elements should not be allowed to be deleted, e.g. a property that acts as an association end.

- [ ] Yes
- [x] No

### Other information

Q: Are all cases covered?


danyeaw (approved):
Nice solution to use singledispatch and the deletable register.

Issue #1246, opened 2021-12-31
a join node looses its size after close and reopen of the program on linux


deifemu:
### Describe the bug
a join node looses its size after close and reopen of the program on linux

### To Reproduce

- draw a join node with connections
- save
- close 
- open

### Expected behavior
stay the same

### Screenshots
![image](https://user-images.githubusercontent.com/81991548/147817481-9c2d9455-b133-42c4-be84-8e29a6ac5d46.png)
![image](https://user-images.githubusercontent.com/81991548/147817510-cd7b8088-8b88-4279-8562-f809cec5dec0.png)

### OS
- [x] Linux (Please put in notes the specific distro)
Linux Mint 20.2, installed with flatpack

- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.7.1

amolenaar:
Hi @deifemu,

Thank you for reporting. I can reproduce this issue. 

@all-contributors please add @deifemu for bug

danyeaw:
@deifemu Thanks again for submitting the issue, this will be fixed in version 2.8.0 which we are getting ready to release soon.

PR #1252, opened 2022-01-01
Fork node loading

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: #1246

- [ ] Yes
- [x] No

### Other information

This is a regression: at some point during the Gaphas 3.0 migration, positions were no longer "normalized" (`handle[0]` was reset to `(0, 0)` and the item `matrix` was updated accordingly).

This fixes the regression, without breaking backwards compatibility. 
A nicer approach may be to implement it like [ExecutionSpecification](https://github.com/gaphor/gaphor/blob/master/gaphor/UML/interactions/executionspecification.py#L116).

NB. `ForkNodeItem` does not save the lines connected to itself.

### DIFF
--- a/gaphor/UML/actions/tests/test_objectnode.py
+++ b/gaphor/UML/actions/tests/test_objectnode.py
@@ -2,25 +2,26 @@ from gaphor import UML
 from gaphor.UML.actions.objectnode import ObjectNodeItem
 
 
-class TestObjectNode:
-    def test_object_node(self, case):
-        case.create(ObjectNodeItem, UML.ObjectNode)
+def test_object_node(case):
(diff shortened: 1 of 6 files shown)


gaphor/UML/actions/tests/test_objectnode.py
```diff
@@ -2,25 +2,26 @@
 from gaphor.UML.actions.objectnode import ObjectNodeItem
 
 
-class TestObjectNode:
```
danyeaw, line 5: Nice job getting rid of all these test classes :+1:

danyeaw (approved):
Nice bug fix @amolenaar!

PR #1262, opened 2022-01-05
Full Copy/Paste of model elements

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Only a shallow (link) type of paste is supported for the diagram items

Issue Number: #791

### What is the new behavior?

A new key binding <key>Ctrl</key>+<key>Shift</key>+<key>p</key> can be used to copy the presentation along with the underlying model elements.

I introduced a `paste_link()` and `paste_full()` function to distinguish between the two types of pasting.

- [ ] Yes
- [x] No

### Other information

Model elements were already copied, because they were needed for cut/paste functionality.

* A lot of changes have to do with changing the test functions `copy_and_paste` and `copy_clear_and_paste` to `*_link`, to distinguish them from the deep-copy functionality.
* Newly created elements now always get a new id.
* I refactored the copy code a little to get rid of some duplication &mdash; introduced a `blacklist` parameter for `copy_element`.

### DIFF
--- a/gaphor/core/modeling/tests/test_elementfactory.py
+++ b/gaphor/core/modeling/tests/test_elementfactory.py
@@ -49,7 +49,7 @@ def test_flush(factory):
 
     gc.collect()
 
-    assert len(list(factory.values())) == 0, list(factory.values())
+    assert not list(factory.values()), list(factory.values())
 
 
 def test_without_application(factory):
@@ -73,7 +73,7 @@ def test_unlink(factory):
 
     p.unlink()
 
-    assert len(list(factory.values())) == 0, list(factory.values())
+    assert not list(factory.values()), list(factory.values())
 
     p = factory.create(Parameter)
     p.defaultValue = "l"
@@ -83,7 +83,7 @@ def test_unlink(factory):
     p.unlink()
     del p
 
-    assert len(list(factory.values())) == 0, list(factory.values())
+    assert not list(factory.values()), list(factory.values())
 
 
 # Event handlers are registered as persisting top level handlers, since no
--- a/gaphor/core/modeling/tests/test_elementfactory.py
+++ b/gaphor/core/modeling/tests/test_elementfactory.py
@@ -1,4 +1,5 @@
 import gc
+from collections.abc import Container, Iterable
 
 import pytest
 
@@ -20,6 +21,14 @@ from gaphor.UML import Parameter
 def factory():
     event_manager = EventManager()
     return ElementFactory(event_manager)
+
+
+def test_element_factory_is_an_iterable(factory):
+    assert isinstance(factory, Iterable)
+
+
+def test_element_factory_is_a_container(factory):
+    assert isinstance(factory, Container)
 
 
 def test_create(factory):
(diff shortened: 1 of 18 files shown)


gaphor/core/modeling/tests/test_elementfactory.py
```diff
@@ -40,7 +49,7 @@ def test_flush(factory):
 
     gc.collect()
 
-    assert len(list(factory.values())) == 0, list(factory.values())
+    assert not list(factory.values()), list(factory.values())
```
danyeaw, line 52: I like this change to be more readable :+1:

gaphor/diagram/copypaste.py
```diff
@@ -3,8 +3,9 @@
 The `copy()` function will return all values serialized, either as string
 values or reference id's.
 
-The `paste()` function will resolve those values. Based on the elements
+The `paste_link()` function will resolve those values. Based on the elements
 that are already in the model,
+`paste_full()` will try copy every element involved.
```
danyeaw, line 8: How about:

The `paste_link()` function will resolve those values and place instances of them on the diagram using the same defining element. The `paste_full()` function is similar, but it creates new defining elements for the elements being pasted.

danyeaw (approved):
This looks great and adds a really nice enhancement to Gaphor :tada: 

My only question is the terms "Paste (link)" and Paste (full)" as clear as they can be?

Even something like Paste with Same Data and Paste with New Data might make it more clear what is going to happen before trying it.

gaphor/core/modeling/tests/test_elementfactory.py
amolenaar: Thanks to Sourcery :wink:

gaphor/diagram/copypaste.py
amolenaar: I improved the docstring with your suggestion :+1:

amolenaar:
> This looks great and adds a really nice enhancement to Gaphor tada
> 
> My only question is the terms "Paste (link)" and Paste (full)" as clear as they can be?
> 
> Even something like Paste with Same Data and Paste with New Data might make it more clear what is going to happen before trying it.

I agree that the terminology could be improved upon. It's like a shallow vs a deep copy. In the docstring it now mentions a _defining element_.  Maybe _link defining elements_ and _copy defining elements_?

danyeaw:
> I agree that the terminology could be improved upon. It's like a shallow vs a deep copy. In the docstring it now mentions a _defining element_. Maybe _link defining elements_ and _copy defining elements_?

That is funny, I thought of the exact same shallow vs. deep copy in Python, but we are doing it on the paste side of things.

In the HIG they just have Paste and Paste Special, maybe we should keep the Paste name and make the full one Paste Special (New Data). Naming things is hard :smile:

amolenaar:
Only _Paste Special_ doesn't say anything about what it's doing. From some issues we got it seemed quite _normal_ to users that paste would paste the underlaying model element as well 😄 .

In the model, the relation is called _subject_, but what to make of that. 

`Paste (link defining elements)` and `Paste (copy defining elements)` looks more reasonable then.

danyeaw:
Ok, that sounds good to me as well 👍

Issue #1264, opened 2022-01-05
Undo of grouped elements positions the grouped elements wrongly


amolenaar:
### To Reproduce
1. Create a C4 System
2. Add a Container to it
3. Select both System and Container and delete them
4. Press Undo
5. See how the grouped elements are located in the top-left corner

### Expected behavior

The items are grouped: drag the system and the nested elements move along.

Only the position is off.

### OS
- [x] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: master

### Analysis

There is a dependency between `Presentation.parent` and `Presentation.matrix`. If `parent` is set, the `matrix` is compensated so it remains on the same place. When undoing, matrix is set first, and then parent is set. This causes the item to move to the right position. If parent was set before matrix, all would be fine.

Undo manager orders the events. ReversibleEvent (a.o. matrix updates) before association updates.

PR #1266, opened 2022-01-06
Fix positions of nested items during undo

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

When undo-ing a nested item, it's positioned wrongly.

Issue Number: #1264

### What is the new behavior?

The nested item appears exactly where it was when it was removed.

- [x] Yes
- [ ] No

To change the parent of an item and keep it visually at the same place, a new method `change_parent()` was introduced.


danyeaw (approved):
This looks great!

PR #1328, opened 2022-01-29
Use Automatic Translation Service to Fill Translation Gaps

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [Gaphor/Gaphor](https://hosted.weblate.org/projects/gaphor/gaphor/).

Current translation status:

Weblate translation status


amolenaar:
I have some doubts. I checked the German translation and many translations are okay, but there are quite some errors in there as well. E.g. "Inhibit Gate" is translated to "Tor sperren", where "inhibit" is used as a verb, instead of a noun ("I am blocking the gate"). This should have been translated to "Sperr Gatter" (I think).

All in all, it could lead to some confusing, maybe funny, translations.

danyeaw:
Ya, I was assuming that 80% is pretty good, and 20% is hilarious. Is it easier for translators to start from there or translate everything from scratch?

Issue #1335, opened 2022-02-02
Dark mode not working on Windows


Viicos:
### Describe the bug
Hey there,
even though dark mode is activated on Windows (`Settings` -> `Personalization` -> `Colors`), Gaphor doesn't seem to detect it and still uses the default clear theme.
Do not hesitate to ask for additional details as I don't really know where it could come from.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [ ] macOS
- [x] Windows

### Version
Version of Gaphor: 2.8.1

amolenaar:
hi @Viicos,

This issue [has come up before](https://matrix.to/#/!maFZnGxzqXqppVZrTd:gitter.im/$96rf2S68ucxdHkLsz5vqY86G0Zr4EL_BeZlsA5mL068?via=matrix.org&via=gitter.im). One way is to set an evironment variable `GTK_THEME` to `Adwaita:dark` (or just `:dark`).

However, I do not think this is a story we can keep up for long: as a user you expect this to *just work* in 2022.

See:
 * https://stackoverflow.com/questions/65294987/detect-os-dark-mode-in-python
 * https://pypi.org/project/darkdetect/
 * https://stackoverflow.com/questions/42881112/pygtk3-how-to-change-application-theme

PR #1336, opened 2022-02-03
Automatically select dark mode for macOS and Windows

amolenaar:
### PR Type

- [ ] Bug fix
- [X] Feature

Dark mode is not picked up for macOS and Windows.

Issue Number: #1335

### What is the new behavior?

Set a dark-mode preference flag on macOS and Windows. Do not do this for any Linux/FreeBSD platform, since it works just fine on GNOME/KDE.

Note that the mode is set on startup, it does not change after that.

- [ ] Yes
- [x] No

Tests:

- [ ] Works on Windows
- [x] Works on macOS

### DIFF
--- a/gaphor/ui/__init__.py
+++ b/gaphor/ui/__init__.py
@@ -7,6 +7,7 @@ import sys
 from pathlib import Path
 from typing import Optional
 
+import darkdetect
 import gi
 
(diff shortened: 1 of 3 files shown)


danyeaw (approved):
Nice, I didn't know about the darkdetect library!

Issue #1009, opened 2021-09-17
Deleting an actor or a usecase shall remove all connecting associations


sz332:
### Describe the solution you'd like

When adding an actor and usecase, and connecting them via an assocation, then removing either the actor or the usecase, it shall remove the connected associations as well.

amolenaar:
Fair point. However, this behavior should be consistent throughout.

At some point we decided to allow association items (the representation) to be coupled with their model elements, so we could support shared and composite associations out of the box.

@danyeaw Any thoughts?

amolenaar:
Associations are removed on model level. This changed in the 2.10 dev cycle. The items on the diagram are not completely removed. Association items are not backed by a model element.

@sz332 Does this fulfill your request?

sz332:
Okay, so my description was probably not on point. A diagram shall be separated from the model. We shall have two different use cases: 

- remove the item from the diagram
- remove the item from the model

I have two items on the diagram. They are also in the model. I draw an association in between the two diagram elements. 

When I remove one of the items **from the diagram** then the associations inbetween any item and the removed item shall also be removed from the diagram (association shall exist on the diagram only with both ends connected to a diagram element). It shall be remain in the model though.

If I remove one of the items from the model then all associations owned by the model element shall be also removed.

amolenaar:
> I have two items on the diagram. They are also in the model. I draw an association in between the two diagram elements.

Now a new association will be created in the model if none existed yet. If an association already exists, it will be drawn in the diagram if its not yet there.

> When I remove one of the items **from the diagram** then the associations inbetween any item and the removed item shall also be removed from the diagram (association shall exist on the diagram only with both ends connected to a diagram element). 

Currently, if one item is removed from a diagram, attached lines remain on the diagram.

> It shall be remain in the model though.

This we have covered by disabling the "Remove Unused Elements" option.

> If I remove one of the items from the model then all associations owned by the model element shall be also removed.

This is how it currently works, right? If I attach a UseCase and an Action with an Association, when I remove the UseCase from the model (e.g. via the tree view), the Association is also removed from the model. But it remains on the diagram. It could be removed here as well -- does make sense, I suppose :smiley:.

PR #1357, opened 2022-02-16
Update connection behavior for Association

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

* Association is the only relation type that holds a model element when disconnected.
* It's not able to find existing associations from the model when a relation is drawn in a diagram.

Issue Number: #1009?

### What is the new behavior?

This PR unifies the behavior of Association with other relation types. Associations can be looked up, it can use the  copy functionality introduced in #1351.

- [ ] Yes
- [x] No

### Other information

Requires #1351 to be merged.

### DIFF
--- a/gaphor/UML/classes/association.py
+++ b/gaphor/UML/classes/association.py
@@ -481,8 +481,7 @@ class AssociationEnd:
             return
 
         cr = context.cairo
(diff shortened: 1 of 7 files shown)


danyeaw (commented):
I like this updated behavior!

Issue #1355, opened 2022-02-15
App crashes when action taken to save


GregJohnStewart:
### Describe the bug
The app completely crashes when you go to save the project.

### To Reproduce
1. Open up a new UML diagram project
2. Go to save the project (`ctrl+s`, `Save` through the ui)

### Expected behavior
App doesn't crash, lets you save the project

### Screenshots

Not a screenshot, but an error from console:

```
(gaphor-exe:25031): GLib-GIO-ERROR **: 21:57:34.821: Settings schema 'org.gtk.Settings.FileChooser' does not contain a key named 'show-type-column'
Trace/breakpoint trap (core dumped)
```
Side note, on program init:
```
Gtk-Message: 21:57:17.430: Failed to load module "canberra-gtk-module"
```

### OS
- [ x] Linux (Ubuntu 20.04.3)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: `2.8.2`

amolenaar:
Hi @GregJohnStewart,

This is an odd error. 

* How did you install Gaphor? Flatpak/appimage/python package?
* What type of desktop environment are you using? GNOME/KDE/XFCE/...?
* Does this issue only happen with Gaphor, or do other (GTK based) applications exhibit the same issue?

A search for the error message on google gives a bunch of hits. You're probably not the first person to run into this issue :wink:.
(e.g. https://unix.stackexchange.com/questions/549741/gtk-filechooser-causes-crashes-in-several-programs).

The `canberra-gtk-module` warning is not a problem.

@all-contributors add @GregJohnStewart for bug.

GregJohnStewart:
Confirmed that `gnome-logs` shows me a file select screen when I hit "Export Logs to File"

danyeaw:
I can reproduce this issue for the segfault, I don't get the missing canberra-gtk-module error though.

GregJohnStewart:
Just checked, this is still an issue with 2.9.2. Is the fix still waiting to make it into a release or does this mean it wasn't quite the fix?

danyeaw:
@GregJohnStewart You are right, I just tested it and I can reproduce the issue again. I am not sure what happened there, I tested PR #1364 and everything was working fine. I'll have to look in to it more.

GregJohnStewart:
Just tested on Fedora 35, getting the same issue.

amolenaar:
@danyeaw @GregJohnStewart Is this still an issue with Gaphor 2.10?

[...]

PR #1364, opened 2022-02-19
Fix AppImage Crashes on Save Command

danyeaw:
This PR sets the environmental variable `XDG_DATA_DIRS` so that Gtk can find the user-specific data files related to GNOME when compiled from source. Also, librsvg was failing to build and this updates the dependencies to fix that issue with the docker image.

### PR Checklist

### PR Type

- [X] Bug fix
- [ ] Feature

AppImage version of Gaphor crashes on save.

Issue Number: #1355 

### What is the new behavior?
No crashes

- [ ] Yes
- [X] No

### DIFF
--- a/.github/workflows/appimage-docker.yml
+++ b/.github/workflows/appimage-docker.yml
@@ -36,5 +36,5 @@ jobs:
           python -m venv .venv
           source .venv/bin/activate
           python -m pip install poethepoet
-          poe appimage-docker-build
+          poe appimage-docker-build-clean
           poe appimage-docker-push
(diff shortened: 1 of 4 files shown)


danyeaw:
I am getting the same error, so this isn't quite right.

amolenaar:
Should we set the `GSETTINGS_SCHEMA_DIR` variable when the Gaphor is started maybe?

danyeaw:
I think the issue is that `XDG_DATA_DIRS` was not set to where jhbuild's share folder was, and then PyInstaller didn't know where to find the GLib schemas.

amolenaar:
Looks good.

PR #1365, opened 2022-02-21
Simulate user behavior with Hypothesis

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

We've seen issues reported by users that we could not properly reproduce.

Issue Number: N/A

### What is the new behavior?

A [Hypothesis Stateful Test](https://hypothesis.readthedocs.io/en/latest/stateful.html) (model-based test) has been added that performs actions a user would do: create/remove items, copy/paste, undo/redo.

So far this test manages to fail in a couple of cases.

- [ ] Yes
- [x] No

, 

### Other information

* [x] Object id's are generated via 1 functions. The implementation can be swapped (required for Hypothesis to work reliably).
* [ ] work with multiple model elements (now only `Class` and `Dependency` are used
* [ ] Does it miss any actions? Grouping is not supported yet.
* [ ] It's currently using only one diagram. Should probably also add/remove diagrams.

To run an in-depth test, run `pytest --hypothesis-profile=ci tests/test_model_consistency.py`. It should raise an error where connections in the diagram do not match connections in the model.

This change also contains some Sourcery updates, not directly related to the functional change.
The implementation can be extended even more. However, this is a good start, and already uncovered an error. Therefore I think it can be merged.

### DIFF
--- a/.github/workflows/hypothesis-test.yml
+++ b/.github/workflows/hypothesis-test.yml
@@ -1,0 +1,43 @@
+name: Test
+on:
+  schedule:
+    # min hour dom month dow
+    - cron:  '0 6 * * *'
+env:
+  python_version: '3.10'
+
+jobs:
+
+  hypothesis:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+    strategy:
+      matrix:
+        python_version: ["3.10"]
+    steps:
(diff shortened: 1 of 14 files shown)


danyeaw (approved):
This looks great @amolenaar!

Issue #1277, opened 2022-01-09
Connectors Won't Connect to Proxy Ports


danyeaw:
### Describe the bug
Connectors won't connect to proxy ports, they always snap to the boundary of a part property instead.

### To Reproduce
1. Create two part properties
2. Add proxy ports to each one
3. Create a connector and try to connect them to the proxy ports

### Expected behavior
Connectors should snap to proxy ports and move when the proxy port is moved.

### OS
- [X] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: master

liavaag95:
I experienced the same on Windows today following the same procedure as specified above.

amolenaar:
Fixed by #1403.

PR #1403, opened 2022-03-16
Proxyport: update ports when proxyport is moved

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

ProxyPort behavior is off when the port is moved around.

Issue Number: #1277 

### What is the new behavior?

The problem was that the port positions were not updated when the port moved. Hence they always stayed at their initial location.

This PR fixes that.

- [ ] Yes
- [x] No


danyeaw (approved):
Nice fix!

PR #1453, opened 2022-04-08
Update permissions for CodeQL GitHub Action

naveensrinivasan:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -3,8 +3,15 @@ name: "CodeQL"
 on:
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   analyze:
+    permissions:
+      actions: read  # for github/codeql-action/init to get workflow details
+      contents: read  # for actions/checkout to fetch code
+      security-events: write  # for github/codeql-action/autobuild to send a status report
     name: Analyze
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,8 +9,6 @@ permissions:
 jobs:
   analyze:
     permissions:
-      actions: read  # for github/codeql-action/init to get workflow details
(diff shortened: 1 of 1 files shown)


amolenaar:
Hi @naveensrinivasan,

Thanks for the improvement. Is this something we should also consider for our other workflows?

naveensrinivasan:
> Hi @naveensrinivasan,
> 
> Thanks for the improvement. Is this something we should also consider for our other workflows?

Yes, For now I was able to do this.

danyeaw:
We could also set the default permissions to be read only:
image

I'm not sure what that would break, but we could try it and do a full build and see. What do you think @amolenaar?

danyeaw:
Ok, I updated the permissions, let's see if things break :crossed_fingers:

naveensrinivasan:
You would still need this https://github.com/gaphor/gaphor/blob/4a1bb2716a1da5dd2a159c6c2b7c3302f319acdb/.github/workflows/codeql-analysis.yml#L14

To write to security events.

danyeaw:
Thanks @naveensrinivasan!

danyeaw:
Some other things we could do to increase our security a bit is to do [signed releases](https://wiki.debian.org/Creating%20signed%20GitHub%20releases), add a security policy from the security tab, and maybe enable fuzz testing through [OSS-Fuzz](https://google.github.io/oss-fuzz).

PR #1462, opened 2022-04-14
Pin support for activity diagram

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Pins are not supported.

Issue Number: N/A

### What is the new behavior?

Support for input and output pins.

- [x] Pins connect to actions
- [x] Object flow connects to pins
- [x] copy/paste
- [x] Pins shall be children of an action, when connected
- [x] What to do with pin names?
- [x] Icons for input and output pin
- [x] ~Pins show arrow when no object flows are connected~ Use stereotype to clarify pin type. There's now enough space for an arrow.
- [x] For input pins, only allow incoming object flow
- [x] For output pins, only allow outgoing object flow
- [x] Pins should be located on the side of the action, not on top of it.

- [ ] Yes
- [x] No

### Other information

- [x] Improve name label placement for _ProxyPort_ and _Pin_.

image

### DIFF
--- a/gaphor/SysML/blocks/proxyport.py
+++ b/gaphor/SysML/blocks/proxyport.py
@@ -15,12 +15,10 @@ from gaphor.UML.recipes import stereotypes_str
 
 def text_position(position):
     return {
-        "text-align": TextAlign.LEFT
-        if position == "left"
-        else (TextAlign.RIGHT if position == "right" else TextAlign.CENTER),
-        "vertical-align": VerticalAlign.TOP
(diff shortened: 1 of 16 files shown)


amolenaar:
Pretty much everything works, except for loading. Then the pins are our of order. This looks a bit like #1467.

Same happens with parameter nodes on activities, btw.

danyeaw (approved):
This looks great! Yay for activity pin support, this is a really nice addition! :tada:

PR #1482, opened 2022-04-22
Fix pin connections when loading a model

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

When loading a model with pins, and object flows connected to those pins, the object flow may be connected in wrongly -> not in a state as when you saved it.

### What is the new behavior?

Fix loading of Attached Presentations. The boundaries of an attached presentation are calculated by constraints. By resolving those constraints it ensures the line is connected exactly as when you saved it.


danyeaw:
Nice one!

Issue #1486, opened 2022-04-24
Allow profiles to be based on all modeling languages


danyeaw:
### Describe the bug
```
Poe => raaml
WARNING:gaphor.storage.storage:file models/RAAML.gaphor could not be loaded (Type Block can not be loaded: no such element 25f62e28-c33b-11ec-819a-f47b099bf663)
Traceback (most recent call last):
  File "<string>", line 2, in <module>
  File "/home/dan/Projects/gaphor/gaphor/codegen/coder.py", line 75, in main
    model = load_model(modelfile)
  File "/home/dan/Projects/gaphor/gaphor/codegen/coder.py", line 408, in load_model
    storage.load(
  File "/home/dan/Projects/gaphor/gaphor/storage/storage.py", line 231, in load
    for status in load_generator(filename, factory, modeling_language):
  File "/home/dan/Projects/gaphor/gaphor/storage/storage.py", line 271, in load_generator
    for percentage in load_elements_generator(
  File "/home/dan/Projects/gaphor/gaphor/storage/storage.py", line 148, in load_elements_generator
    yield from _load_elements_and_canvasitems(
  File "/home/dan/Projects/gaphor/gaphor/storage/storage.py", line 192, in _load_elements_and_canvasitems
    create_element(elem)
  File "/home/dan/Projects/gaphor/gaphor/storage/storage.py", line 181, in create_element
    assert cls, f"Type {elem.type} can not be loaded: no such element {elem.id}"
AssertionError: Type Block can not be loaded: no such element 25f62e28-c33b-11ec-819a-f47b099bf663

```

### To Reproduce
1. Edited the RAAML.gaphor file.
2. Ran `poe raaml`
3. See error above, model file is attached.
[RAAML.gaphor.zip](https://github.com/gaphor/gaphor/files/8550173/RAAML.gaphor.zip)

### OS
- [X] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: main branch

amolenaar:
On `main` `poe raaml` works, did you change/add something?

danyeaw:
If I start adding Blocks that are Stereotyped with Situation, the models become corrupted. See the raaml.gaphor model attached to this issue.

amolenaar:
Ah, the "problem" is in the function `load_model` (https://github.com/gaphor/gaphor/blob/main/gaphor/codegen/coder.py#L403).

[...]

PR #1494, opened 2022-04-28
Codegen use all defined modeling languages

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

For code generation only models can be loaded that use the UML language.

Issue Number: #1486

### What is the new behavior?

* The code generator uses `ModelingLanguage`s to find the class from super models
* In addition to the UML language, additional languages can be loaded. (not sure if we need this)


danyeaw (approved):
Nice one @amolenaar!

PR #1516, opened 2022-05-24
GTK4 list view

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

In our GTK4 port we use the (old) `Gtk.TreeView`. Treeview does not support dnd. Also, it's considered sort of deprecated, since with `Gtk.ListView` we have a nicer API.

Issue Number: #618 

### What is the new behavior?

Use the new `Gtk.ListView` to display the hierarchy. Feature parity with the GTK3 tree view.

### Tasks

- [x] Add/delete elements in tree view
- [x] Create tree when model is loaded (`ModelReady`, `ModelFlushed`)
- [x] Highlight element selected in a diagram in the list view (`DiagramSelectionChanged`)
- [x] Subtree for relationships
- [x] Edit names in tree view (not static ones like `<Relationships>`)
- [x] Elements are ordered by name (case insensitive)
- [x] Formatting. Diagrams are bold, abstract elements italic.
- [x] Pop up menu
- [x] Drag and drop elements within the list view
- [x] Drag and drop elements from list view to diagram
- [x] Double click opens diagram / adds element to current diagram
- [x] Search for elements and filter based on search results (hide non-matching elements?)

### Impediments

- [x] Elements without children (empty child model) still show an expansion arrow.
- [x] The model can't be sorted: [`Gtk.Expression` is not implemented in PyGObject](https://gitlab.gnome.org/GNOME/pygobject/-/issues/457) -> use custom sorter.

### Other

* Fixed deadlock when editing text elements. This case is not possible with GTK3 (there only textbuffers could deadlock).
* The GTK4 tree view is implemented as a separate component.
* Implemented our own element editing (stack + label + text), since the default `EditableLabel` has shortcuts that conflict with the tree view. Also, this allowed me to allow editing for the name only.
* DnD is now communicated with (g)objects. This is simpler than using strings.


danyeaw (approved):
Outstanding @amolenaar! This looks great and gets us one giant leap closer to making GTK4 the default :+1:

Issue #1519, opened 2022-05-29
Create an error report from within Gaphor


amolenaar:
### Is your feature request related to a problem?

Sometimes users raise issues that are hard to reproduce by the developers.

### Describe the solution you'd like

To assist developers in debugging an issue it would be nice to see if there were any exceptions raised by the application. Normally developers would see those, as they run Gaphor from the command line. For people running the Flatpak/Windows/macOS those exceptions are hidden.

I think it would be really helpful if Gaphor can create an error report. This report could include:

* OS
* Gaphor version
* Python version
* Raised exceptions
* Did I miss anything?

### Describe alternatives you've considered

* The report could also include some sanity checks:
  * Are all linked model elements in the repository
  * Do all Presentation elements have a diagram attached
* An option to attach serialized application state? (saved model, maybe a pickled model)

[Python Bytes 270, item 5](https://pythonbytes.fm/episodes/show/270/can-errors-really-be-beautiful).

PR #1529, opened 2022-06-04
Error Reports window

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Users do not see uncaught exceptions. However, they may be really valuable when diagnosing a problem

Issue Number: #1519

### What is the new behavior?

All uncaught exceptions are collected by an Error Reports service. Tools -> Error Reports. These errors can be looked up and attached to a GitHub issue.
A notification is created on the active session window to notify the user such an exception happened.

This allows us to uncover errors, even when the application is ran on Windows or from a Flatpak.

To make this work the Console had to do a little smarter exception handling.

- [ ] Yes
- [x] No

### Other information

image

### DIFF
--- a/gaphor/plugins/errorreports/errorreports.py
+++ b/gaphor/plugins/errorreports/errorreports.py
@@ -1,21 +1,24 @@
 from __future__ import annotations
 
+import datetime
 import logging
+import platform
 import sys
+import textwrap
 import time
 from types import TracebackType
 
-import stackprinter
+from better_exceptions import format_exception
+from exceptiongroup import BaseExceptionGroup
 from gi.repository import Gtk
 
 from gaphor.abc import ActionProvider
 from gaphor.action import action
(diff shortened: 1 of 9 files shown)


gaphor/plugins/errorreports/errorreports.py
```diff
@@ -0,0 +1,167 @@
...
+                Notification(
+                    gettext(
+                        "An unexpected error occurred.\nPlease file an issue on GitHub and include information from Tools → Error Reports."
+                    )
+                )
+            )
+
+
+def os_information():
+    return textwrap.dedent(
```
danyeaw, line 160: Cool, I didn't know about textwrap in the standard libary :+1:

danyeaw (approved):
Really nice enhancement, I think this will make it much easier for everyone to post error reports :+1:

Issue #1535, opened 2022-06-06
Hypothesis test failure caused by a pasting elements


amolenaar:
### Describe the bug

Hypothesis found an issue. The issue can be reproduced by decorating `ModelConsistency` in `tests/test_model_consistency.py`:

```python
from hypothesis import reproduce_failure
...
@reproduce_failure('6.46.11', b'AAABAAEACAABAAEFAAAAAQAAAAEJAAEAAQYAAwABAgA=')
```

Running the test yields the following result:

```
===================================== test session starts ======================================
platform darwin -- Python 3.9.13, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/arjan/Development/gaphor, configfile: pyproject.toml
plugins: mock-3.7.0, xdoctest-1.0.0, hypothesis-6.46.11, cov-3.0.0
collected 1 item                                                                               

tests/test_model_consistency.py F                                                        [100%]

=========================================== FAILURES ===========================================
____________________________________ test_model_consistency ____________________________________

    def test_model_consistency():
>       run_state_machine_as_test(ModelConsistency)

tests/test_model_consistency.py:73: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.9/site-packages/hypothesis/stateful.py:222: in run_state_machine_as_test
    run_state_machine(state_machine_factory)
.venv/lib/python3.9/site-packages/hypothesis/stateful.py:107: in run_state_machine
    @given(st.data())
.venv/lib/python3.9/site-packages/hypothesis/core.py:662: in execute_once
    result = self.test_runner(data, run)
.venv/lib/python3.9/site-packages/hypothesis/executors.py:47: in default_new_style_executor
    return function(data)
.venv/lib/python3.9/site-packages/hypothesis/core.py:658: in run
    return test(*args, **kwargs)
.venv/lib/python3.9/site-packages/hypothesis/stateful.py:180: in run_state_machine
    result = rule.function(machine, **data)
tests/test_model_consistency.py:233: in undo
    undo_manager.undo_transaction()
gaphor/services/undomanager.py:215: in undo_transaction
    transaction.execute()
gaphor/transaction.py:170: in _transactional
    return func(*args, **kwargs)
gaphor/services/undomanager.py:71: in execute
    act()
[...]

PR #1536, opened 2022-06-06
Fix issue with fully pasting a diagram

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

A curious construction of packages containing diagrams can not be fully/deep copied.

Issue Number: #1535 

### What is the new behavior?

When pasting (or loading) a relationships is established. However, if the opposite side is already set, and has a multiplicity of `[0..1]`, the existing relation should not be overwritten.


danyeaw (approved):
I love that Hypothesis is helping to find some of these edge cases!

Issue #1485, opened 2022-04-24
SysML Enumerations Should Also be ValueTypes


danyeaw:
### Is your feature request related to a problem?
In the RAAML profile, we need to be able to make an enumeration show up as values for a Block. For example, Controllability has enumerations (C0, C1, C2, C3) and needs to be a value of AccidentScenario.

![image](https://user-images.githubusercontent.com/10014976/164979943-e881da17-a667-4030-870f-d1d35bae7a1f.png)
![image](https://user-images.githubusercontent.com/10014976/164979971-269fe81f-41bb-4b3e-beb9-725d3159f485.png)

### Describe the solution you'd like

Solution could be in two parts:

1. Allow an Enumeration to select the ValueType stereotype.
2. When an Enumeration is created with the SysML profile, the ValueType stereotype is automatically applied. Although I don't see this explicitly mentioned in the SysML spec, this is how other tools like MagicDraw work.

### Describe alternatives you've considered
I can't think of any alternatives.

danyeaw:
Related to this, our profile parser should also be able to parse Enumerations, maybe as a Python `enum.Enum` type?

amolenaar:
Currently Enumerations are converted to `enumeration` properties based on a class' extension (`-Kind`). It would be nice to use stereotypes instead.

amolenaar:
@danyeaw said in [Matrix](https://matrix.to/#/!maFZnGxzqXqppVZrTd:gitter.im/$4uDCzMxOi5IOUAkJ1UheTk9zQKBCpnjzglJkcS8hvEg?via=gitter.im&via=matrix.org):
> This is an explanation of why MagicDraw automatically applies ValueType to an enumeration: https://community.nomagic.com/correct-stereotype-for-enumerated-value-types-t2607.html

From what I understand, there is no «enumeration» type for SysML. Instead a «ValueType» should be used. Shouldn't we just remove Enumeration from the SysML _blocks_ menu?

danyeaw:
I need a ValueType with Enumeration literals. I guess we could remove the toolbar button for an enumeration and then have a selection on the ValueType property page to turn on enumeration maybe. I'm not sure how that would work in the data model though.

amolenaar:
Hmm.. Both `ValueType` and `Enumeration` inherit from `DataType`. `Enumeration` has a collection of `EnumerationLiteral`'s. Currently the model does not support applying stereotypes on model level. 

[...]

PR #1543, opened 2022-06-08
Make SysML Enumerations also ValueTypes

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

UML's Enumeration is made available in the SysML profile. However, Enumeration is not really supported. It should be a ValueType.

Issue Number: #1485

### What is the new behavior?

When you create an enumeration with the SysML profile selected, it displays a ValueType stereotype to the Enumeration.

### DIFF
--- a/gaphor/SysML/blocks/blockstoolbox.py
+++ b/gaphor/SysML/blocks/blockstoolbox.py
@@ -18,6 +18,12 @@ from gaphor.UML.classes.classestoolbox import (
     shared_association_config,
 )
 
+
+def sysml_enumeration_config(new_item: uml_items.EnumerationItem, name=None):
+    namespace_config(new_item, name)
+    new_item.as_sysml_value_type = True
+
+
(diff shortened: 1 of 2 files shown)


danyeaw:
In other tools there is still an Enumeration object for SysML in the toolbar, it just applies a valueType stereotype to the enumeration when you create it. I don't think we want to get rid of the ability to create an Enumeration.

amolenaar:
Now added an attribute to Enumeration. It will either show with UML or SysML profile, depending if you take the enum from the UML or SysML profile.

image

There's no switch in the property editor, on purpose.

danyeaw:
This PR doesn't remove the Enumeration from the SysML Blocks toolbox now, right? For some reason, I can't get it to show up, I even created a new virtualenv.

image

danyeaw (approved):
This is a great solution @amolenaar :+1:

Issue #274, opened 2020-02-26
Copy/Paste functionality should use Gtk.Clipboard


amolenaar:
### I'm submitting a...
- [X] Bug report  
- [X] Feature request
- [ ] Documentation issue or request

To fix this completely, we have to use [GTK 4](https://github.com/gaphor/gaphor/issues/618).

### Current behavior

* Copy/paste does not work as expected. When a diagram item is copied, it's not possible to paste a text copied from somewhere else.
* It's not possible to copy (part of) a diagram and paste it in another application.

### Expected behavior

* Copying diagram items is only done when the diagram has focus - now it's a "win" action.
* Copy diagram items should allow them to be pasted as PNG/SVG in other applications.

### Minimal reproduction of the problem with instructions

1. Add an item to a diagram.Edit name field in element editor and paste. It won't paste anything in the text field but raise an exception instead.
2. Add an item to a diagram. Copy the item, edit name field in element editor and paste. It'll paste the diagram item instead

### Notes

This issue only concerns copy/pasting within Gaphor.
1. in the same model: copy model elements, relations between elements if they're copies together.
2. Model to model: What should be the rules when copying from one model to another? Model elements should also be copied along. They should get new id's.

We can unite 1. and 2. to some extend. Two options:
1. If both reference the same model, only the Presentation element (diagram item) is copied
2. If elements already exist with the same id, they are reused (this breaks when a model has been copied on disk and edited)

We can also copy/paste to other applications, in this case the selected elements should be rendered to PNG or SVG, in order to be easily copy/pastable in other documents.

### References
* https://lazka.github.io/pgi-docs/index.html#Gtk-3.0/classes/Clipboard.html
* https://python-gtk-3-tutorial.readthedocs.io/en/latest/clipboard.html

amolenaar:
Copy and Paste should be treated as two separate actions. Copy should make a snapshot of
* the selected items
* their connections
* their model elements
* and maybe more (e.g. reference to a stereotype)

Pasting can be done in different ways:
1. paste in the same model
 * add diagram items to the active diagram
 * reuse references to existing model elements
 * recreate connections on diagram side (recreate on model side if removed in between copy and paste)
2. paste in a different model
 * add diagram items to the active diagram
 * Add model elements to the ElementFactory
   * Which elements/references should be created/reused?
   * reuse id or create new id's? Reuse would mean you can copy parts from one model and add them to the new model incrementally (e.g. when multiple diagrams are involved).
 * ...
[...]

PR #1565, opened 2022-06-24
Add Copy/Paste for GTK4

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

No proper copy/paste in our GTK4 version

Issue Number: #274, #618 

### What is the new behavior?

Copy paste works with Gdk.Clipboard.

This will allow us to extend it with copy/paste of images, for example, so you can easily share stuff with other apps. Also we can paste things like text in the application now.

- [x] Yes
- [ ] No

- [x] Fix issue when cutting elements in GTK4. Causes error in tree model. #1567


danyeaw (approved):
Nice update, one more piece to the GTK4 puzzle!

PR #1567, opened 2022-06-25
Fix unlinking elements from the model 

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

I ran into a nice edge case. In gaphor 2.10 we changed the behavior of the EventManager. If an event B is triggered while event A is being handled, B is held back and handled once all handlers for A have been executed.
This fixed some issues in our Undo manager.

However, it seems like it can also introduce some unexpected behavior.

We have a Sanitizer service. This service is responsible for deleting a model element if there are no presentations of that element left. The effect (removed from model and Element.model set to None) takes effect immediately. However the event is added to the queue where a bunch of events still need to be handled. Some of the event handlers expect Element.model to be set.
One way to solve it is to ensure the effect (Element.model set to None) only takes place once that event is to be handled.
Another is to assume that Element.model can be unset (raises exception) and handle this exception accordingly.

Issue Number: N/A

### What is the new behavior?

The actual unlinking happens when the event `UnlinkEvent` is handled. A `ElementDeleted` event is resubmitted after this event has been processed.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/core/modeling/element.py
+++ b/gaphor/core/modeling/element.py
@@ -26,7 +26,11 @@ log = logging.getLogger(__name__)
 
 
 class UnlinkEvent:
-    """Used to tell event handlers this element should be unlinked."""
+    """Used to tell event handlers this element should be unlinked.
+
+    This event is used internally and should not be handled outside
+    `gaphor.core.modeling`.
+    """
 
     def __init__(self, element: Element, diagram: Diagram | None = None):
         self.element = element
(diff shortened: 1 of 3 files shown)


amolenaar:
I think this works fine. Not sure if `Element._model` should be unassigned from `UnlinkEvent` or `ElementFactory`. Now it's `ElementFactory`.

danyeaw (approved):
Looks good!

PR #1578, opened 2022-06-29
Improve build pipelines

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

We have separate pipelines for test (PR's) and the main build.

Issue Number: N/A

### What is the new behavior?

* AppImage, DMG and Windows installer are build is a separate jobs
* One wheel (Linux, Python 3.10) is used as a basis for all installers.

- [ ] Yes
- [x] No

### Other information

image

Installers are working:

- [ ] Linux AppImage
- [x] Windows
- [x] Windows portable
- [x] macOS DMG (no translations!)
- [ ] Are translations also bundled?


danyeaw (commented):
This looks great, I love the refresh! I would like to test the built packages before we merge :+1:

amolenaar:
> I would like to test the built packages before we merge

Do you mean something like a `gaphor --self-test`, which launches an instance and does a couple of basic tests (e.g. can a session be loaded, are Gtk and friends available).

danyeaw:
If we had a way to do more end to end testing that could launch Gaphor as an installed application, that would be awesome. I just meant that we should install Gaphor manually on Windows and macOS to make sure things are still working.

amolenaar:
@danyeaw I think we can mergethis. Can you check the windows installers?

danyeaw:
Windows is working great, unfortunately the AppImage is not. I am getting the following error:

```
(gaphor-exe:15590): GLib-GIO-ERROR **: 13:59:45.502: Settings schema 'org.gnome.settings-daemon.plugins.xsettings' does not contain a key named 'antialiasing'
```

amolenaar:
Is this an issue you do not see with on the main branch?

danyeaw:
> Is this an issue you do not see with on the main branch?

Yup, tested the last main branch build and the AppImage works great.

amolenaar:
I can reproduce the issue. Had a look inside. I see the number of GSettings schema files (xml) is the same in both images. However the file `gschemas.compiled` is 36k on the main branch build and only 3k on the new appimage build. 
The 36k file has the same size and timestamp as the file found in the build container. 

It's peculiar that the main version is not complaining, since the package `gnome-settings-daemon-schemas` is not installed. From what I see that's where the property is coming from.

One more thing to check: are the translations included (.mo files)

PR #1583, opened 2022-07-02
Let Gaphor check for it's own health

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

We (Dan and me) have to manually check if a packaged version of Gaphor will actually start. 

### What is the new behavior?

To be new behavior:

- [x] Application is checked on a clean linux/macos/windows host
- [x] Install the application if needed (windows)
- [x] Checks if all required libraries are present: Gtk, Pango
- [x] Checks if a new session can be started without any exceptions showing up
- [ ] Did I miss something?

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -267,6 +267,8 @@ jobs:
     permissions:
       contents: write
     timeout-minutes: 45
+    outputs:
+      installer: gaphor-${{ steps.setup_and_test.outputs.version }}-installer.exe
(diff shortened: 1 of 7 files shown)


amolenaar:
This is nice I think. Maybe we should direct the logging on windows to a file and print that? It would require for a custom logging config, though.

danyeaw (approved):
This is sooo cool! This adds a ton of confidence that the built app will work, I love it. :1st_place_medal:

PR #1589, opened 2022-07-05
"Full build" pipeline for PR's

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

We have 2 pipeline scripts: one for PR's (test) and one for main (full-build). This means double maintenance and can be harder to debug.

### What is the new behavior?

One full-build pipeline, with conditional steps for the mainline build, mostly signing related.

- [ ] Yes
- [x] No

### Other information

Example non-main build: https://github.com/gaphor/gaphor/actions/runs/2619062283


.github/actions/linux_appimage/action.yml
```diff
@@ -0,0 +1,36 @@
+name: 'Create macOS Application'
```
danyeaw, line 1: The name and description should be updated here

.github/actions/macos_dmg/action.yml
```diff
@@ -0,0 +1,100 @@
...
+        appstore-connect-username: ${{ inputs.ac_username }}
+        appstore-connect-password: ${{ inputs.ac_password }}
+        primary-bundle-id: org.gaphor.gaphor
+    - name: Staple .dmg
+      if: inputs.mainline_build == 'true'
+      uses: devbotsxyz/xcode-staple@v1.0.0
+      with:
+        product-path: "_packaging/dist/Gaphor-${{ inputs.version }}.dmg"
+    - name: Upload Gaphor-${{ inputs.version }}.dmg
+      if: inputs.mainline_build == 'true'
```
danyeaw, line 80: This was a good idea to only build the dmg on release :+1:

danyeaw (approved):
This is a great simplification to the build pipeline, I love it @amolenaar!

danyeaw:
@amolenaar It looks like the Linux .tar.gz and *.whl file aren't found and the `${{ steps.setup_and_test.outputs.version }}` is blank.

.github/workflows/full-build.yml
```diff
@@ -145,63 +189,21 @@ jobs:
...
+          base64_encoded_p12: ${{ secrets.BASE64_ENCODED_P12 }}
+          certpassword_p12: ${{ secrets.CERTPASSWORD_P12 }}
+          ac_username:  ${{ secrets.AC_USERNAME }}
+          ac_password: ${{ secrets.AC_PASSWORD }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
 
   check-macos-dmg:
     needs: macos
     runs-on: macos-latest
+    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/packaging/')
```
danyeaw, line 206: Should we just make this `if: ${{ mainline_build == 'true' }}`?

amolenaar: That doesn't work: mainline_build is an env var and therefore can't be used in those job level if statements. Sadly.

Issue #1585, opened 2022-07-04
Unicode encoding Problems


qsodev:
### Describe the bug
The model files are incorrectly encoded.  

Models created on a Windows machine cannot be opened on Linux.   
Models created on Linux contain wrong characters on Windows.

### To Reproduce

1. On Windows open Gaphor create a new `UML Model`
2. Save the file
3. Try to open this file on Linux
4. Error `This file does not contain a valid Gaphor model`
5.  Terminal output shows `UnicodeDecodeError`...

The problem here is that Gaphor creates some sample diagrams. In `Use Cases` diagram there is a note that contains non standard `-` character. This causes the error on Linux.  
The same problems occur when you write German Umlaute like `üöä ÜÖÄ`. You get the same UnicodeDecodeErrors !

#### The other way:
1. On Linux open Gaphor create a new `UML Model`
2. Write some nonstandard characters like `ü ö ä ÖÄÜ`
3. save
4. open this file on Windows -> the File opens but the characters are wrong.

### Expected behavior
All unicode characters should be encoded correctly and the file should open an all platforms

### OS
- [x] Linux (Archlinux, Flatpak)
- [ ] macOS
- [x] Windows

### Version
Version of Gaphor: 2.10 ( Portable on Windows, flatpak on Linux )

amolenaar:
Hi, thank you for reporting this. All models should have been encoded with utf8.

amolenaar:
I created a little model myself. I can confirm the issue. It looks like the model is written with the default encoding (Windows 1252 on my VM), which is not UTF-8, as it should be.

danyeaw:
@all-contributors please add @qsodev for bug

danyeaw:
Thanks @qsodev for the bug report!

PR #1594, opened 2022-07-06
Ensure all models are saved with UTF-8 encoding

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Models with non-ascii characters created on Windows, can't be loaded on Linux.

Issue Number: #1585 

### What is the new behavior?

Files are explicitly saved with UTF-8 encoding.

- [ ] Yes
- [x] No

### Other information

I expected UTF-8 to be the default encoding, also for files, since `sys.getdefaultencoding()` returns `utf-8`.

### DIFF
--- a/gaphor/storage/parser.py
+++ b/gaphor/storage/parser.py
@@ -311,7 +311,12 @@ def parse_generator(filename, loader):
     parser.setFeature(handler.feature_namespaces, 1)
     parser.setContentHandler(loader)
 
-    yield from parse_file(filename, parser)
+    try:
+        # returns only a progress percentage
+        yield from parse_file(filename, parser)
+    except UnicodeDecodeError:
+        # Fall back on default encoding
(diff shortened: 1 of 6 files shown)


gaphor/storage/parser.py
```diff
@@ -311,7 +311,12 @@ def parse_generator(filename, loader):
     parser.setFeature(handler.feature_namespaces, 1)
     parser.setContentHandler(loader)
 
-    yield from parse_file(filename, parser)
+    try:
+        # returns only a progress percentage
+        yield from parse_file(filename, parser)
+    except UnicodeDecodeError:
+        # Fall back on default encoding
+        yield from parse_file(filename, parser, encoding=None)
```
amolenaar, line 319: NB. This except block should allow old models, saved and loaded on windows, to load.

danyeaw (approved):
Looks great!

Issue #1580, opened 2022-07-02
Gaphor in Windows is Incorrectly Installed to x86


danyeaw:
### Describe the bug
GTK is compiled for 64-bit and Gaphor is running in x64 mode:
![image](https://user-images.githubusercontent.com/10014976/177000387-fb124953-41b4-4727-a5ac-68816363a02c.png)

However, it is installed in the C:\Program Files x86\ directory.

To fix this, we'll need to update the `win_installer.nsi` file and change the INSTALLDIR to `$PROGRAMFILES64`.

### Expected behavior
Gaphor is installed to the correct Program Files directory.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [ ] macOS
- [X] Windows

### Version
Version of Gaphor: Latest

amolenaar:
Will it uninstall the old version of Gaphor still (from the x86 folder)?

danyeaw:
I'm not sure what will happen, we'll have to test it

amolenaar:
Where should it be installed then? In `C:\Program Files\`?

danyeaw:
Yup!

PR #1602, opened 2022-07-09
Move Windows Installation to C:\Program Files

danyeaw:
This PR moves the installation location on Windows to C:\Program Files from C:\Program Files (x86).

If upgrading versions, it will automatically uninstall previous versions, even if they are in C:\Program Files (x86), or any other location. However, it will by default specify the previous location and the user would need to remove the (x86) if they want it installed somewhere else, which I think is OK. For new installs, it always defaults to C:\Program Files.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Gaphor is installed to C:\Program Files (x86)

Issue Number: #1580

### What is the new behavior?
Gaphor is installed to C:\Program Files

- [ ] Yes
- [X] No

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -379,5 +379,5 @@ jobs:
       - name: Perform self-test
         run: |
           start "" /WAIT "${{ needs.windows.outputs.installer }}" /S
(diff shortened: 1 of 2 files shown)


amolenaar:
Nice work 🙂. I think the check job also needs an update.

amolenaar:
A nice, clean PR. :+1:

PR #1604, opened 2022-07-10
Make GTK4 the Default Version and Enable Adwaita

danyeaw:
This PR:

- Switches to making GTK4 the default version
- Enables GtkSourceView for CSS highlighting with GTK4
- Enables libadwaita with GTK4

### PR Checklist

### PR Type

- [ ] Bug fix
- [X] Feature

GTK version 3 is the default

Issue Number: #618 

### What is the new behavior?
GTK version 4 is the default

- [ ] Yes
- [X] No

### DIFF
--- a/.github/actions/linux_dependencies/action.yml
+++ b/.github/actions/linux_dependencies/action.yml
@@ -5,5 +5,5 @@ runs:
   steps:
     - run: >
         sudo apt-get update -qq && sudo apt-get install -qq --no-install-recommends
-        gir1.2-gtk-3.0 libgirepository1.0-dev libgtksourceview-4-dev upx
+        gir1.2-gtk-4.0 libgirepository1.0-dev libgtksourceview-5-dev upx
       shell: bash
--- a/.github/actions/linux_dependencies/action.yml
+++ b/.github/actions/linux_dependencies/action.yml
@@ -4,6 +4,6 @@ runs:
   using: composite
   steps:
     - run: >
-        sudo apt-get update -qq && sudo apt-get install -qq --no-install-recommends
-        gir1.2-gtk-4.0 libgirepository1.0-dev libgtksourceview-5-dev upx
+        sudo apt-get update -qq && sudo apt-get install -qq --no-install-recommends upx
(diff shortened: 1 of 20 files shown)


danyeaw:
I took a swim trying to compile GTK4 on Ubuntu 18.04, and it will be very difficult. It requires gnutls >= 3.6.13, and I tried to rebuild it using dpkg-buildpackage. One of the patches with the 3.5 version was pushed upstream, and even if I don't apply that, I am still getting test errors during the build. I think we should leave the AppImage on GTK3 along with macOS, and upgrade everything else.

amolenaar:
This looks really promising. I like the direction you're taking this.

It's failing on console tests again. I wonder what makes them crash this time.

_packaging/make-script.py
```diff
@@ -32,6 +32,9 @@ def make_gaphor_script():
         # Check for and remove two semicolons in path
         file.write("os.environ['PATH'] = os.environ['PATH'].replace(';;', ';')\n")
 
+        # Load gaphor with GTK3, the frozen gaphor-exe is segfaulting with GTK4
+        file.write("os.environ['GAPHOR_USE_GTK'] = '3'\n")
```
amolenaar, line 36: What is the effect of this line? From what I read from the changes, Windowsa and Linux are fully running GTK4. Or is this only fixing the packaging step?

amolenaar (approved):

amolenaar:
Really nice. It looks really nice with `Adw.ApplicationWindow `.

_packaging/make-script.py
danyeaw: For packaging, I am having PyInstaller use GTK3 for all platforms (macOS, Linux AppImage, and Windows). I am getting a segfault with GTK4, so we'll have to dive in to this more. I think it is still worth switching the default and updating our flatpak builds to use GTK4 though.

amolenaar: Thanks, that was not completely clear to me when I scanned the code.

This is a really nice PR.

Issue #1599, opened 2022-07-07
Sequence Diagram: CreateMessage shows LostMessage circle


Toorero:
### Describe the bug
If you create an *unconnected* CreateMessage in a Sequence Diagram, a circle like on LostMessage appear. If you'd connect the tip to another Lifeline element, it will be converted into a Communication Diagram style message.

![grafik](https://user-images.githubusercontent.com/22551563/177862082-e44b5bf5-1ada-4967-8c2f-11a1b7328a3f.png)

### Expected behavior
![grafik](https://user-images.githubusercontent.com/22551563/177863468-7244aeb4-9876-49a6-a0ea-a3341f5c663c.png)

### OS
- [x] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.10.0

### Work Around
I can currently work around this issue *temporarily* by creating an CreateMessage, connect the tip temporarily to sth, unconnect it and move the other Lifeline to the end of the tip, so they don't connect together. But as soon as I restart the program, the LostMessage circle is present again.

### Suggested Fix
I would favor if the CreateMessage just can't be converted into a Communication Diagram style message, so I can snap it as accustomed.

amolenaar:
The _LostMessage_ on an unconnected create message seems okay to me.

It would be cool if a message is set to a _Create_ message as soon as one end connects to a _Lifeline timeline_ or _Execution Specification_ and the arrow end to a _Lifeline body_.

PR #1612, opened 2022-07-13
Add createMessage to element support with auto convert

Toorero:
### PR Checklist

### PR Type

- [x] Bug fix
- [ ] Feature

 Issue Number:  #1599

### What is the new behavior?
A message is converted automatically to a `createMessage` as soon as it's clear that it can only be a `createMessage`. Also, I relaxed the allowed connections of messages in general quite a bit.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/UML/interactions/interactionsconnect.py
+++ b/gaphor/UML/interactions/interactionsconnect.py
@@ -136,13 +136,23 @@ def convert_create_message(line, convert_back=False):
     def communication_element(elem):
         return isinstance(elem, LifelineItem) and not elem.lifetime.visible
 
-    if tail and communication_element(tail) and head and not communication_element(head):
-        if convert_back and line.subject and line.subject.messageSort == "createMessage":
+    if (
+        tail
+        and communication_element(tail)
+        and head
+        and not communication_element(head)
+    ):
+        if (
+            convert_back
+            and line.subject
+            and line.subject.messageSort == "createMessage"
+        ):
             line.subject.messageSort = "synchCall"
         else:
             line.subject.messageSort = "createMessage"
         return True
     return False
+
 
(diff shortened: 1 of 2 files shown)


Toorero:
I still have to fix the issue that files with the new supported `createMessage`-kind don't load up correctly.

### Before save ###
grafik
### After reopen ###
grafik

amolenaar:
This needs more unit tests. I also expect certain tests to break, because we loosen the requirement that a messages can exist between lifetime lines and object bodies.

Toorero:
> This needs more unit tests. I also expect certain tests to break, because we loosen the requirement that a messages can exist between lifetime lines and object bodies.

Sure thing. I was just working on implementing the actual feature but since I don't really know why the aforementioned issue occurs I was holding back to write tests for something that isn't even implemented yet.

amolenaar:
> > This needs more unit tests. I also expect certain tests to break, because we loosen the requirement that a messages can exist between lifetime lines and object bodies.
> 
> Sure thing. I was just working on implementing the actual feature but since I don't really know why the aforementioned issue occurs, I was holding back to write tests for something that isn't even implemented yet.

Fair point.

There used to be a problem, when after loading lines were connected to the body of the lifeline object, instead of the timeline. This is fixed by a [specific implementation of the connectionSink object for lifeline](https://github.com/gaphor/gaphor/blob/main/gaphor/UML/interactions/interactionsconnect.py#L313). It's basically forcing messages to connect to the lifetime part of the lifeline object. Of course this is no longer true, but if we would just remove this sink, I would expect messages connection almost to the top of the lifetime to connect to body again after loading.

amolenaar:
I created #1715. It allows more liberate connection of messages to lifelines -- body and timeline.

@Toorero Can you test it, to see if it matches your expectations?

PR #1633, opened 2022-07-23
Fix "rename" from popover

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

When you select the "Rename" entry from the popover menu in the tree view, the text is not editable on Wayland.

On X11 it works fine, though :/.

### What is the new behavior?

"Rename" is available fro a shortcut (`F2`). I could not find a way to make the popover work in combination with the edit feature.
Therefore I propose to remove the "Rename" option from the popover.

### DIFF
--- a/gaphor/ui/treecomponent.py
+++ b/gaphor/ui/treecomponent.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from gaphas.decorators import g_async
 from gi.repository import Gdk, GLib, GObject, Gtk
 
 from gaphor import UML
@@ -155,7 +154,6 @@ class TreeComponent(UIComponent, ActionProvider):
         self.event_manager.handle(DiagramOpened(element))
 
     @action(name="tree-view.rename", shortcut="F2")
-    @g_async(single=True)
     def tree_view_rename_selected(self):
         if row_item := self.selection.get_selected_item():
             tree_item: TreeItem = row_item.get_item()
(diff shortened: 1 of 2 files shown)


danyeaw:
@amolenaar Do you think this is the only option? I rather liked rename being in the right click menu.

amolenaar:
@danyeaw It's a bit weird. It looks like focus is (temporary) changing when the popup is removed. Something that does not happen on X11.

But now that you mention it, maybe it works if I delay editing by some time, say 200ms.

amolenaar:
@danyeaw I think you're gonna like those changes.

danyeaw:
That is working much better for me :+1: 

One thing that I noticed which seems somewhat inconsistent on how we do rename:
1. If you rename from a diagram, hitting Escape undoes the editing and hitting enter accepts it
2. If you rename from the tree view, hitting Escape accepts the edit, and hitting enter creates another copy of the element on the diagram.

I think this could be pretty unexpected for a user. What do you think?

danyeaw (approved):

amolenaar:
This is indeed some unexpected behavior. It should just exit, and escape should cancel the edit.

amolenaar:
I addressed the issue with editing. I forgot to provide a proper return value if the key (enter/escape) was handled.

danyeaw:
Thanks @amolenaar, that is much better!

Issue #1635, opened 2022-07-24
Stereotype (de)application is broken


amolenaar:
### To Reproduce
1. Create a stereotype (stereotype --|> `Class`)
2. Apply the stereotype to a newly created Class item
3. Show stereotypes
4. Remove the extension between metaclass and stereotype
5. Stereotypes are still applied to the class

### Expected behavior

The stereotypes should have been removed from the class, once the extension relation is broken.

### Version
Version of Gaphor: 2.11.0

### Additional information

Stereotype cleanup happens in the Sanitizer service. I suspect this is broken ever since the event manager triggers handlers in a different order.

PR #1637, opened 2022-07-24
Fix stereotype application

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

stereotypes were not removed from elements when the extension relation was broken in the profle.

Issue Number: #1635 

### What is the new behavior?

Stereotype unapplication is handled explicitly in the connector. This behavior could no longer be handled by the Sanatizer service, due to the number of properties involved -- and events that get handled much later than the actual action.

- [ ] Yes
- [x] No


amolenaar:
Let's hold this off. There may be a better way.

amolenaar:
#1644 is a better solution

PR #1644, opened 2022-07-27
Fix stereotype unapplication

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Stereotypes are not removed from classes when the (extension) relation has been broken.

Issue Number: #1635 

### What is the new behavior?

Stereotypes do are again removed when extensions are removed.

- [ ] Yes
- [x] No

### Other information

This regression was casued by the new event manager. Now events are queued. So when the event is handled, a lot of propertyies can have changed already. The new algorithm checks for which elements the stereotype application is still valid and removed it from the invalid ones.


danyeaw (approved):
Nice simple way to fix this :+1:

amolenaar:
@danyeaw It only took me a week or so to find out ;)

Issue #1363, opened 2022-02-19
feature request:import python code


ghost:
Hi

 Would like to request a feature: importing python code that is converted into uml automatically by reading the classes.
Also class relations (inheritance, composition) can be created.

Thanks

amolenaar:
Hi @pygamedev,

Indeed this would be a nice feature. Do you know of any (Python) library capable of doing so?
One thing we probably also want then is a way to lay out the elements in a diagram. So far I have not been able to find a proper library for diagram layout.

amolenaar:
Options:

* Pyreverse: https://github.com/PyCQA/pylint/tree/main/pylint/pyreverse
* PyNSource: https://github.com/abulka/pynsource (can do layouting too)

vanillajonathan:
You would build an abstract syntax tree (see the [ast](https://docs.python.org/3/library/ast.html) module), then you iterate over that tree to create a diagram from it.

danyeaw:
Thanks for the feature request. We have updated the README to make it more clear that this isn't one of our goals for the project.

Issue #615, opened 2021-01-19
Describe a diagram in text


amolenaar:
**Is your feature request related to a problem? Please describe.**

Tools like PlantUML provide a simple, textual way to describe a model. The advantage of using text is that it allows for a very straightforward way of describing a model. Text has an additional benefit in that it works well when used with an SCM (e.g. git).

With SysML also providing a (textual) markup language, it may be desirable to support this also in Gaphor.

**Describe the solution you'd like**

I like an optional text box below the diagram that contains a textual representation of the model. 

**Describe alternatives you've considered**

**Additional context**

It may be useful to review the file format, so merge conflicts with Gaphor models using textual representation may be easier to solve. Maybe this issue can help resolve #388.

To make this work, the model has to be automatically drawn and layed out. the built in constraint solver in Gaphas doesn't do that. Maybe a Cassowary constraint solver will do the trick.

* https://mbse4u.com/2020/12/21/sysml-v2-release-whats-inside/
* https://pypi.org/project/cassowary/
* https://kiwisolver.readthedocs.io/en/latest/

danyeaw:
I like this idea, I think showing a text representation would be a great feature.

The issue I always have with PlantUML is that I find it often falls apart for me when trying to model something more complex than a few model elements. Other tools like MagicDraw have a lot of auto layout features as well, but I'm never really happen with them except for a simple tree hierarchy.

We could try to be smarter than the other tools, but I think that this is probably a hard problem to solve.

Is some type of hybrid possible where we have a text representation but still keep the diagram layout from the user? I will look more in to what exactly they are planning for SysML v2.

amolenaar:
Some possible diagramming libs:
* [Diagrams](https://diagrams.mingrammer.com/)
* [Networkx](https://networkx.org/)
* [pydot](https://pypi.org/project/pydot/)

We should be fine as long as we can provide width and height for elements, and read back the coordinates to use to place the elements.

travisfw:
If I could edit mermaid diagrams and export them back into a markdown file, I'm sure a lot of people would find that useful.

amolenaar:
Let's start to break down the problem:

- What should be the role of the textual representation?
- How does the textual representation relate to diagrams? Does one only define the model in text, or also its views?
- How does a textual representation relate to the model (elements) we currently have?
- Do we actually need elements in the model, or can the model be ephemeral? How to deal with object placement then? 

[...]

PR #1653, opened 2022-08-03
Diagram auto-layout with Graphviz

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Currently we have no way to automatically layout a diagram. This is a feature we probably want if we want to allow textual input and diagram generation.

Issue Number: N/A

### What is the new behavior?

I've been playing around with Graphviz as layout engine.

Nested elements: https://stackoverflow.com/questions/2012036/graphviz-how-to-connect-subgraphs

### Other information

I've been looking at PyNSource, but the license (GPL3) does not allow us to include parts of his code.

This feature could ease implementation of #615 and #1363.

Packaging of Graphviz is handled by #1797.


amolenaar:
For bundling:

* https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/86
* https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/357

Install Graphviz via Brew/Choco/Apt. Flatpak should build from source.

amolenaar:
It provides acceptable output (and does not crash).
There are 2 things I think can improve the diagrams:

1. Use digraph and `->` to direct head and tail.
2. provide the existing `pos` to elements to guide placement.

amolenaar:
@danyeaw Still graphviz (`dot[.exe]`) needs to be bundled and made sure to work. But you can give it a try. Since diagrams are generated they look less fancy compared to hand drawn diagrams. On the other hand, this allows us to look at more code-centric workflows. Ergo, it's a nice utility.

gaphor/plugins/autolayout/tests/test_pydot.py
```diff
@@ -0,0 +1,109 @@
...
+    connect(gen, gen.head, subclass)
+
+    auto_layout = AutoLayout(event_manager, None)
+    auto_layout.layout(diagram)
+
+    assert gen.head.pos != (0, 0)
+    assert gen.tail.pos != (0, 0)
+
+
+def test_layout_with_association(diagram, create, event_manager):
```
danyeaw, line 35: Does this test, the comment one, and fork node one,  need an assert?

danyeaw (commented):
This looks great. Having automatic layout in Gaphor is a huge contribution!

gaphor/plugins/autolayout/tests/test_pydot.py
amolenaar: The alternative is that an except would be raised. What I wanted to assert is that it does not crash ;). It's hard to make asserts about the positions the elements are located at after auto-layout.

amolenaar:
Added one more commit: the auto-layout feature will only be exposed if `dot` can be found.

amolenaar:
BTW. Instead of pydot, we can also use [Graphviz' python API](https://graphviz.org/pdf/gv.3python.pdf). That might fix our dependency issue.

One other issue to deal with are the plugins. Those are compiled and installed on a "fixed" location.

danyeaw (approved):

Issue #1654, opened 2022-08-04
State Machine diagrams: composite states and choice states (and other missing elements)


vanem:
The editor has a great look@feel, but I'm mostly interested in UML (Hierarchical) State Machines. The most important elements missing would be Composite States (otherwise you are limited to very simple behaviors ) and to a lesser extent the Choice Pseudostate.
Other elements are missing like: adding multiple activities, custom activities (also called "internal self transitions"),  Regions, Join, Fork

amolenaar:
Hi @vanem,

Thank you for your feedback. For the most part I've been focusing on activity diagrams (Actions). State machines could use some love :smile:.

A quick check makes me assume pseudo states can easily be added to [PseudostateItem](https://github.com/gaphor/gaphor/blob/main/gaphor/UML/states/pseudostates.py) and the [toolbox](https://github.com/gaphor/gaphor/blob/main/gaphor/UML/states/statestoolbox.py). Composite states are probably a bit more work, mostly allow states to be nested.

Can you help with implementing? It will be very much appreciated.

vanem:
Unfortunately I have no UI /Gtk experience (and dealing with 2 jobs atm). https://github.com/DLR-RM/RAFCON builds on gaphas (but their interpretation of HFSMs is very different from UML), perhaps someone could find some useful inspiration in their repo.

amolenaar:
Okay, PR for pseudostates is up.

In order to have a complete story for state machines we still need:

- [ ] Allow multiple regions in a state machine
- [ ] Allow composite states. E.i. a region in a state that can contain states and transitions.

Having multiple regions in a state (machine) poses an issue similar to #1014: when resizing, adding and removing regions the relation between the state and it's (visually) containing region should be maintained.

amolenaar:
State machines:

![image](https://user-images.githubusercontent.com/96249/185649270-94d64421-b395-41ab-b2d1-e87bf4977b1c.png)

amolenaar:
Implemented by #1662 and #1666.

PR #1662, opened 2022-08-09
State diagram: support all pseudostates

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Only initial and shallowHistory pseudostates are supported.

Issue Number: #1654

### What is the new behavior?

All pseudostate kinds are supported.

### To do

- [x] create icons for remaining states
- [x] update connector rules -> removed extra rules for initial pseudostate

image


amolenaar:
Currently there's only one connection rule: an _initial_ pseudostate can only have one outgoing transition, no incoming. The specs only constrain the number of outgoing transitions.

The UML specs describe a few constraints (section 14.5.7 of the UML 2.5.1 specs), mostly on outgoing connections for several pseudostates. Not sure if I should bother implementing those.

Maybe we should just relax and allow any number of incoming and outgoing transitions from all pseudo elements. For example: for initial/junction, and join/fork, the difference if not visible from the diagram.

Should we support a dropdown to change the kind of pseudostate?

danyeaw (approved):
This looks great @amolenaar. I wasn't aware that UML had this many state types, but this a great addition. :+1:

PR #1666, opened 2022-08-11
State machines and regions

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

State machines are only created implicitly to hold states.

Issue Number: #1654 

### What is the new behavior?

- [x] State machines can be created from the toolbox
- [x] (pseudo)states can be moved between state machines
- [x] Region support in Statemachine (state / pseudostate / transition)
- [x] Region support in (composite) State
- [x] property page for regions (create / remove) - how to deal with elements in the region? 
- [x] create state / pseudostate / transition from toolbox places it in the right region

### Other

This PR contains a changes where stereotype watch and compartment code is directly imported from the stereotype module.

### DIFF
--- a/gaphor/UML/states/propertypages.py
+++ b/gaphor/UML/states/propertypages.py
@@ -3,6 +3,9 @@
 To register property pages implemented in this module, it is imported in
 gaphor.adapter package.
 """
+
+from __future__ import annotations
+
 from gaphor import UML
 from gaphor.core import transactional
 from gaphor.diagram.propertypages import (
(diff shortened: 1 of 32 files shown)


gaphor/UML/states/propertypages.py
```diff
@@ -11,6 +14,9 @@
     handler_blocking,
     new_resource_builder,
 )
+from gaphor.UML.states.state import StateItem
+
+from .statemachine import StateMachineItem
```
danyeaw, line 19: To stay consistent with the other import statements, should we use absolute imports here?

danyeaw (commented):
I got about 1/2 way through this, and it looks great so far!

gaphor/UML/states/propertypages.py
amolenaar: Yes it should. I suppose VSCode tried to help me by importing this item

danyeaw (approved):
Wow! Nice improvement to the state machine functionality!

PR #1687, opened 2022-08-23
Assign activities to classifiers

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

### What is the new behavior?

Activities (behaviors) can now be nested in classifier elements. This allows for an ARCADIA style of modeling.

- [ ] Yes
- [x] No

### Other information

Update `vertical-align` in items to `justify-content`. Also checked if all items properly move their name if they get a child item.

### DIFF
--- a/gaphor/C4Model/diagramitems/container.py
+++ b/gaphor/C4Model/diagramitems/container.py
@@ -1,6 +1,6 @@
 from gaphor.C4Model import c4model
 from gaphor.core import gettext
-from gaphor.core.styling import FontWeight, TextAlign, VerticalAlign
+from gaphor.core.styling import FontWeight, JustifyContent, TextAlign
 from gaphor.diagram.presentation import ElementPresentation, Named
 from gaphor.diagram.shapes import Box, Text, draw_border
 from gaphor.diagram.support import represents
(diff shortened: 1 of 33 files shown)


danyeaw:
I am not very familiar with ARCADIA since it is so tightly coupled with Capella. Does UML and SysML allow for an activity to be nested in a Classifier? I am use to this behavior to structure relationship being created through creating swimlanes, assigning a Classifier to each swimlane, and putting the activities in the swimlanes.

amolenaar:
There is an ownership relation between the behavior and the classifier.

image

I made it available as a grouping options, which allows you to assign activities to those elements, either in the tree view or in a diagram. The ownership is indeed valid according to UML.

Most of my exposure to ARCADIA is on the functional level, where you assign activities to certain classifiers. Quite superficial I suppose.

danyeaw:
Not disagreeing, just want to make sure we get this right.

With other tools, if you add an activity to a Class Diagram or Block Definition Diagram it is represented like this with ownership:

image

If you drop an activity on a block, it adds an attribute:
image

Here's the tree view:

image

amolenaar:
I see. It's a bit odd to assign an activity and present it as a property, though.

Containment also works in this case. I updated the containment connection code to use our "group" feature.

image

danyeaw (approved):
Looks good!

PR #1692, opened 2022-08-26
Make model properties strict

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

You can assign any property to a model element. It can be an existing property or a new one (e.g. due to a typo).

### What is the new behavior?

The model element will raise an `AttributeError` if you try to assign anything to a non existant property.

- [x] Yes
- [ ] No

### Other information

An alternative may be to use `__slots__`, but then we should make sure all presentation elements also have slots defined, or the effect of strong typing is sort of gone. It may be an option for a follow-up PR.


danyeaw (approved):
Nice extra error handling and refactoring! :+1:

PR #1697, opened 2022-08-28
Fix Instant editor focus on GTK4

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

When an instant editor closes, focus is not automatically returned to the diagram. This causes users to do an extra click, before (for example) shortcuts can be used again

### What is the new behavior?

Instant editors work in GTK4 the same as GTK3.

- [ ] Yes
- [x] No


danyeaw (approved):
Yay for making GTK4 just a little bit better 👍

PR #1711, opened 2022-09-05
Update poetry to version 1.2.0

danyeaw:
This PR updates Poetry to version 1.2.0. In order to do that, it replaces the legacy way that poetry group dev and non-dev dependencies and creates new groups. An overview of this is in the [poetry 1.2.0 announcement](https://python-poetry.org/blog/announcing-poetry-1.2.0/)

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: N/A

- [ ] Yes
- [X] No


amolenaar:
The group functionality looks like a nice improvement over the current approach :+1:

danyeaw:
image
PyInstaller isn't packaging everything we need for some reason.

gaphor/ui/selftest.py
```diff
@@ -1,14 +1,33 @@
...
+            env_path,
+        ),
+        None,
+    ):
+        with os.add_dll_directory(first_gtk_path):
+            import gi
+
+else:
+    import gi
+
```
amolenaar, line 28: There should be no need to put it here, if it's already defined in the `__init__.py` of the same module.

gaphor/ui/__init__.py
```diff
@@ -1,13 +1,31 @@
...
+            and Path.is_file(Path(path) / "girepository-1.0-1.dll"),
+            env_path,
+        ),
+        None,
+    ):
+        with os.add_dll_directory(first_gtk_path):
+            import gi
+
+else:
+    import gi
```
amolenaar, line 28: Why do we need this? Can't we use the old way instead?

amolenaar (changes requested):
I think this PR is mixing 2 things: updates for Poetry 1.2 and a new way to install stuff on Windows.

Is there a reason to combine these? I like the improvements for Poetry, since it removes a lot of "extras" from the distribution.

I'd rather not have the win32 `add_dll_directory` workaround.

danyeaw:
Unfortunately Poetry 1.2.0 will update installed packages even if I already have the correct version installed. So our previous method of installing the wheels from gvsbuild and then installing the other packages no longer works.

amolenaar:
The win32 patch can also be applied to 'gaphor/__init__.py', since Gaphas also requires this. Or a special module, similar to macos shim module.

danyeaw:
@amolenaar Ok, I refactored the logic and created a windowsshim module. Unfortunately, I can't get PyInstaller to work without https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/211 changes to PyGObject directly.

amolenaar:
This looks good. I think we do not need the same code in `selftest.py`. It's unfortunate we can't see why the windows self-test job is failing :disappointed: .

amolenaar:
Moving the windows fix up doesn't resolve the issue. It's like the import is already handled in the pyimod2- thingy 🤔

amolenaar:
I created 0948b674bcba1478c207ffc862b39fddd15cdb06 (branch poetry-1.2-test). This fixes the windows distribution.

Do we still need the workaround in our code then?

danyeaw:
Nice one @amolenaar, I didn't consider overriding the installed wheel after the poetry run.

pycairo doesn't need the workaround any longer, so we only need to override PyGObject, I updated the PR with just that one.

I think the workaround is still good, otherwise every time we run `poetry install` in Windows we need to then run `pip install --force-reinstall C:\gtk-build\build\x64\release\pygobject\*.whl` which is pretty annoying.

We should also squash and merge this one, somehow the git history got crazy ugly when rebasing - I must have run a git pull.

amolenaar (approved):

amolenaar:
> We should also squash and merge this one, somehow the git history got crazy ugly when rebasing - I must have run a git pull.

I fixed the git history and merged it. A nice improvement :+1:.

Issue #1720, opened 2022-09-07
Some characters like '<' and '>' cannot (reliably) be used in parameter types


JuliusBrueggemann:
### Describe the bug
Some characters (so far I've noticed '<', '>' and '&') lead Gaphor to truncate a parameter type. For example, when trying to add the method `+ test(str: Generic<Type>)`, it gets truncated to `+ test(str: Generic)`.

Strangely, this behavior seems to be somewhat dependant on the return type of the method. If the return type of the method contains one of the special characters, the parameter types can now also use those characters. For example, `+ test(str: Generic<Type>): Generic<Type>` works as expected.

### To Reproduce
1. Create an empty class diagram
2. Create a new class
3. Type in the method `+ test(str: Generic<Type>)` and press `Enter`
4. The method is truncated to `+ test(str: Generic)`
5. Type in the method `+ test(str: Generic<Type>): Generic<Type>` and press `Enter`
6. The method is now displayed as expected

`≡` (top-right menu) → `Tools` → `Error Reports` reports no errors.

### Expected behavior
Entering the operation `+ test(str: Generic<Type>)` always results in the operation `+ test(str: Generic<Type>)` being added.

### OS
- [x] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

Installed as Flatpak from Flathub on Fedora 36.

### Version
Gaphor 2.11

JuliusBrueggemann:
Sorry, I didn't check my GitHub notifications and didn't see your message quick enough. I would have gladly fixed this as my first PR to this project. Maybe next time...

Anyway, thank you for fixing this so quickly. This bug was the only thing that ever made me consider looking for a different UML tool, so I'm very glad its fixed now. Definitely looking forward to the next release :smile:

amolenaar:
@JuliusBrueggemann No problem. I'm glad I could help.

PR #1722, opened 2022-09-08
Support Generic<Type> syntax

amolenaar:
For attributes and operations (parameters).

### PR Type

- [ ] Bug fix
- [x] Feature

Currently syntax like `attr: Generic<Type>` is not supported.

Issue Number: #1720

### What is the new behavior?

The `<Type>` bit is parsed as part of the type value. No special meaning is provided. This is the same as it currently works for union types.


danyeaw (approved):
Looks good!

PR #1742, opened 2022-09-20
Fix appimage docker

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

The AppImage docker container does not build for a while already.

### What is the new behavior?

Currently _jhbuild_ is breaking on _pixman_. Somehow that checkout goes wrong. Not sure why.

### DIFF
--- a/.github/workflows/appimage-docker.yml
+++ b/.github/workflows/appimage-docker.yml
@@ -1,4 +1,4 @@
-name: AppImage Docker Image Build and Push
+name: AppImage Docker Image
 on:
   schedule:
     - cron: '15 1 * * 1' # Weekly at 01:15 UTC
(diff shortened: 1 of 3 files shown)


amolenaar:
@danyeaw Instead of trying to fix the docker image. Would it be a viable alternative to just use Ubuntu 20.04?

danyeaw:
@amolenaar I think if we don't want to try to continue to support AppImage, then we should just drop support for it. The AppImage team doesn't consider support for only newer glibc versions as a working AppImage.

I agree that the support overhead is large and AppImage as a format puts all that burden on the developers.

amolenaar:
@danyeaw I agree with you. I close this PR.

Issue #1747, opened 2022-09-22
A diagram can delete itself by deleting its container from within the diagram


ebousse:
### Describe the bug

1. An expected behavior of Gaphor is to be able to drag and drop any model element into any diagram, to display it in the diagram.
2. It is also possible in Gaphor to "store" a diagram inside a model element, more precisely in a package. 

If we combine (1) and (2), this means it is possible (and valid) to display in a diagram _the package where the diagram is stored_, through a drag and drop.

However, in that case, by default, deleting such package in the diagram will completely delete the package from the model, and thus also delete everything contained in the package, **including the diagram itself.**

In my opinion (1) this is counter-intuitive, as suddenly the diagram disappears while one it only trying to remove an element from it, and (2) removing non-empty container elements from a diagram should not remove the container from the model, as this destroys all enclosed elements.   

### To Reproduce
1. Create an empty generic model — by default it contains a single package containing a single diagram
2. Drag the root package into the diagram to make it appear there
3. Select the root package appearing in the diagram, and press the DELETE key

Then the whole diagram disappears, and everything is deleted from the model.

### Expected behavior

I would expect the package to be removed from the diagram, but not deleted from the model

### Screenshots

Video :
[Capture vidéo du 22-09-2022 11:42:50.webm](https://user-images.githubusercontent.com/5868014/191719606-a10054d7-0c2e-4f1e-a2ba-e8572bb6158b.webm)

### OS
- [x] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

Fedora 36, GNOME 42.4

### Version
Version of Gaphor: 2.11.0 (Flatpak)

ebousse:
> I suppose the rule could be: "if the item is owning (is a parent of) the diagram, remove only the diagram item."

Yes I believe this should work!

It could also help to have a way to explicitly delete either only the diagram item, or the element from the model? Shift-DELETE to only remove the diagram element, for instance?

amolenaar:
We have an option to to remove elements not shown in any diagram in the preferences pane.

![image](https://user-images.githubusercontent.com/96249/191751239-b9d6967b-e591-4a4b-a9e7-a136d58f3ca1.png)

Personally I like the model and diagrams to be in sync, but many like to treat those separately.

ebousse:
> We have an option to to remove elements not shown in any diagram in the preferences pane.

good to know, thanks!

> Personally I like the model and diagrams to be in sync, but many like to treat those separately.

[...]

PR #1751, opened 2022-09-23
Prevent deleting an element in a diagram also deleting the diagram itself

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

A diagram can be automatically deleted when you delete an item that's also an owner of the diagram. Although this is correct, according to Gaphor's inner workings, it's quite surprising for a user to see his diagram gone spontaneously.

Issue Number: #1747

### What is the new behavior?

The diagram is moved to toplevel before the owner element is removed. This avoids surprising removal.

- [ ] Yes
- [x] No

### Other information

I initially tried to fix it in the `SanitizerService`. However, then we should determine what to do while the item is halfway unlinking. Then it's not possible to determine if an element is an owner element reliably.


danyeaw (approved):
Looks good!

Issue #1713, opened 2022-09-05
Invalid Matrix error when adding Use Case to diagram


danyeaw:
### Describe the bug
Adding a Use Case to a digram creates an invalid matrix error.

### To Reproduce
1. Open Gaphor
2. Select a Use Case from the toolbox and add it to the diagram

```
|Traceback (most recent call last):
|  File "/home/dan/Projects/gaphor/gaphor/diagram/painter.py", line 33, in paint_item
|    item.draw(
|  File "/home/dan/Projects/gaphor/gaphor/diagram/presentation.py", line 151, in draw
|    self._shape.draw(
|  File "/home/dan/Projects/gaphor/gaphor/diagram/shapes.py", line 173, in draw
|    self._draw_border(self, new_context, bounding_box)
|    │                 │     │            └ Rectangle()
|    │                 │     └ DrawContext(cairo=<cairo.Context object at 0x7f8d8b9db2f0>, style={'background-color': (0.0, 0.0, 0.0, 0.0), 'color': (0.0, 0.24...
|    │                 └ <gaphor.diagram.shapes.Box object at 0x7f8d72558fd0>
|    └ <gaphor.diagram.shapes.Box object at 0x7f8d72558fd0>
|  File "/home/dan/Projects/gaphor/gaphor/UML/usecases/usecase.py", line 42, in draw_usecase
|    path_ellipse(cr, rx, ry, bounding_box.width, bounding_box.height)
|    │            │   │   │   │                   └ Rectangle()
|    │            │   │   │   └ Rectangle()
|    │            │   │   └ 0.0
|    │            │   └ 0.0
|    │            └ <cairo.Context object at 0x7f8d8b9db2f0>
|    └ <function path_ellipse at 0x7f8db063c5e0>
|  File "/home/dan/Projects/gaphor/.venv/lib/python3.10/site-packages/gaphas/util.py", line 121, in path_ellipse
|    cr.scale(width / 2.0, height / 2.0)
|    │        │            └ 0.0
|    │        └ 0.0
|    └ <cairo.Context object at 0x7f8d8b9db2f0>
|cairo.Error: invalid matrix (not invertible)
```

### Expected behavior
No error

### OS
- [X] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: main

### Additional information
OS: Linux (5.19.2-1-default)
Python version: 3.10.6
GTK version: 4.7.2
[...]

PR #1752, opened 2022-09-24
Fix invalid matrix error when creating use case

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

We use the ellipse function from Gaphas. This one does not work when width or height is 0, something that happens initially for elements.

Issue Number: #1713 

### What is the new behavior?

For Use Case we already used a bezier based ellipse. Now this ellipse algorithm is used for all ellipse and circle shapes.

In addition, the API for this function is closer to the rest of our shape api's. The ellipse algorithm has been tweaked to make a perfectly round circle.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/C4Model/diagramitems/database.py
+++ b/gaphor/C4Model/diagramitems/database.py
@@ -1,10 +1,8 @@
-from gaphas.util import path_ellipse
-
 from gaphor.C4Model import c4model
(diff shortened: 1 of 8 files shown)


danyeaw (approved):
Nice simplification!

Issue #1788, opened 2022-10-10
Unexpected error when deleting an element


DKX47:
### Describe the bug
Trying to remove any element in any diagram with the "del" key doesn't remove the element but produces an error.

### To Reproduce

If possible, add the error report from `≡` (top-right menu) → `Tools` → `Error Reports`.

OS: Linux (5.19.0-76051900-generic)
Python version: 3.10.6
GTK version: 4.8.1
Display: X11Display
Gaphor version: 2.12.1

Errors:

Time since application startup: 0:00:06
|Traceback (most recent call last):
|  File "/app/lib/python3.10/site-packages/gaphor/diagram/tools/shortcut.py", line 25, in on_delete
|    state == 0 or state & Gdk.ModifierType.MOD2_MASK
|    │             │       └ <GdkProxyModule <IntrospectionModule 'Gdk' from '/usr/lib/x86_64-linux-gnu/girepository-1.0/Gdk-4.0.typelib'>>
|    │             └ <flags 16 of type Gdk.ModifierType>
|    └ <flags 16 of type Gdk.ModifierType>
|AttributeError: type object 'ModifierType' has no attribute 'MOD2_MASK'

### Expected behavior
Delete the selected element

### OS
- [x] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

flatpak version

### Version
Version of Gaphor: 2.12.1

amolenaar:
Hi @DKX47, Thank you for reporting. It looks like field `MOD2_MASK` is not defined for GTK 4.0.

Did you use a modifier key when you hit `Delete`?

@all-contributors add @DKX47 for bug reports

DKX47:
Hi @amolenaar, no I didn't use any modifier key

amolenaar:
As a workaround you can start Gaphr with the environment variable `GAPHOR_USE_GTK=3`. This will start Gaphor with GTK3, instead of 4.

amolenaar:
I found this issue can also be triggered if _caps lock_ is enabled (or _scroll lock_).

DKX47:
Thanks @amolenaar for your support, that did the trick :)

To anyone else stumpling upon this bevor the patch is released: Execute in your terminal:
flatpak override --user --env=GAPHOR_USE_GTK=3

lessinge:
Thanks a lot @DKX47  :+1:

PR #1794, opened 2022-10-11
Fix key modifiers for delete element shortcut

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

1. When deleting, it checks if there are no modifiers or the META modifier is pressed. However, keys like _caps lock_ and _scroll lock_ are also modifiers -> deletion will not work.
2. The modifier masdk `MOD2_MASK` is no longer available in GTK4.

Issue Number: #1788 

### What is the new behavior?

The Delete element shortcut works without checking modifier keys.

- [ ] Yes
- [ ] No


danyeaw (approved):
Nice fix!

PR #1823, opened 2022-10-23
Use entry points for pyinstaller builds

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Currenly a Pyinstaller build consists of 3 steps: create `gaphor-script.py`, create a (windows) version file, build the package.

### What is the new behavior?

Packaging is done in one step. The version file step is included in the spec file. Instead of creating a script to import all entry points, use the `collect_entry_point` function from PyInstaller to collect the entry points.

- [ ] Yes
- [x] No


_packaging/gaphor.spec
```diff
@@ -70,14 +87,27 @@ a = Analysis(
...
+
 pyz = PYZ(a.pure, a.zipped_data, cipher=block_cipher)
+
+pyinstaller_versionfile.create_versionfile(
+    output_file="windows/file_version_info.txt",
+    version=get_semver_version(),
+    company_name="Gaphor",
+    file_description="Gaphor",
+    internal_name="Gaphor",
+    legal_copyright="Copyright © 2001-2021 Arjan J. Molenaar and Dan Yeaw.",
```
danyeaw, line 106: While we are at it, what do you think about importing datetime and then:

`legal_copyright=f"Copyright 2011-{datetime.date.today().year}...`

danyeaw (approved):
Really nice simplification!

_packaging/gaphor.spec
amolenaar: I want to make one copyright variable: we also need it for the macOS build.

danyeaw: Great, I like it!

PR #1825, opened 2022-10-24
Pytest now uses importlib

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

`tests` folders have a `__init__.py` file.
According to [Pytest's good practices](https://docs.pytest.org/en/latest/explanation/goodpractices.html), that is not required.

### What is the new behavior?

Pytest now uses `importlib`. 

- [ ] Yes
- [x] No


danyeaw (approved):
Cool, I like that you incorporated this best practice!

Issue #1834, opened 2022-10-27
Delete with CANC


Lucaffo:
When i want to delete a file, i used to press CANC and confirm with enter "Do you want to delete this file?" or without this popup. 

I'd like to add the esc button to delete the current selected fle in the top-left hierarchy. 

Or in another way, add the support for shortcuts? 

![image](https://user-images.githubusercontent.com/55745404/198383473-5738d7dd-80e3-4baf-9fe5-2f2041539a1b.png)

Anyway you (all) made a great work with this project! GG

amolenaar:
Hi,

Thank you for the suggestion.

However, `ESC` is normally not used for deleting elements. When it comes to shortcuts we try to stay close to what's common on Linux/macOS/Windows.

Lucaffo:
I'm dumb, i mean CANC :(

amolenaar:
Ehhmm.. I do not have such a button on my US keyboard layout :thinking:

amolenaar:
I created a small PR that allows you to hit Delete (CANC?) to remove an element from the tree view. However, this only works for GTK4. From your screenshot I see you're using the GTK3 version. Which OS are you using?

Lucaffo:
I use Windows 10 👍🏻

amolenaar:
Cool. The next release will use GTK4 for Windows.

PR #1838, opened 2022-10-30
Support Delete to delete entries from tree view

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

You need to use the popup on the tree view if you want to remove an element.

Issue Number: #1834

### What is the new behavior?

Elements can be  deleted from the tree view directly by hitting Delete.

Only works for GTK 4.

- [ ] Yes
- [x] No


danyeaw (approved):
:+1:

Issue #388, opened 2020-08-07
Support a way to resolve (git) merge conflicts with Gaphor


amolenaar:
### I'm submitting a...
- [ ] Bug report  
- [X] Feature request
- [ ] Documentation issue or request

### Current behavior

If a merge conflicts arises because two people edited a model file at the same time, it's hard to resolve. There are two options:

1. Interpret the XML and merge the right parts :crossed_fingers: 
2. Pick one version and re-apply the changes the other person made (or remembers)

### Expected behavior

We want to be able to resolve merge conflicts in a user friendly way.

We need to find out what a user friendly way is. via CLI or via the GUI? How to present "conflicts"? How to resolve those?

Should Gaphor be able to read a file containing a merge conflict?

NB. In the repo (tag 1.0.0) is a primitive compare tool (`utils/compare.py`). It may be useful for inspiration.

More refinement required, I suppose.

amolenaar:
One way to go about this is to load the before and after model and generate a list of changes in terms of `ElementEvent`'s -- create element, update association, etc. The events can be grouped logically, to give some insight in the changes that have to be applied. E.g. a presentation element should be attached to a diagram and probably needs a subject.

amolenaar:
I have an idea on how to pull this off.

1. A model with merge conflicts can be loaded in both it's _current_ and _incoming_ form. The _current_ model will be loaded in the session, and the _incoming_ model will be side-loaded for comparison.
2. The models are compared, this results in a list of removed, added, and changed objects.
3. Those changes are recorded in the _current_ model in a `ChangeSet` object.
4. Gaphor can show the changes in an ordered fashion in the right side pane. something like:
    a. Created diagrams (with elements as children)
    b. Modifications to diagrams
    c. Modifications to model elements
5. The user can apply changes by checking them. The changes will be marked in the change set (which is part of the model, so you can undo the changes).
6. The user can mark the change set as resolved, which will remove the change set from the model.

This way we can add meaning to the individual changes and group them. The _smarts_ are in grouping related changes. When you create a diagram, the items on the diagram are created too, including their model elements.

Technically, multiple change sets can exist, although I think we need only one.

@danyeaw What do you think?

danyeaw:
That sounds like a great plan! How would the change sets be represented? Visually? XML? Some translation into text?

amolenaar:
I think it'll be a tree view of sorts. You should at least be able to select and unselect changes. Maybe it should replace the left side bar (tree + toolbox).

[...]

PR #1845, opened 2022-11-02
Resolve (git) merge conflict: choose which branch to load

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

When a model has been edited on two branches, there's a big change of merge conflicts. It's not possible to resolve those merge conflicts with a text editor (our file format is xml).

One way would maybe be to figure our a merge-friendly file format, but since the model is littered with object references, this seems like a hard job.

Issue Number: #388 

### What is the new behavior?

Gaphor itself should aid in resolving merge conflicts on a semantic level. This should provide the user with the information needed to update a model with incoming changes.

### DIFF
--- a/gaphor/ui/filemanager.py
+++ b/gaphor/ui/filemanager.py
@@ -3,8 +3,10 @@
 from __future__ import annotations
 
 import logging
+import tempfile
 from pathlib import Path
 from queue import Queue
+from typing import Callable
 
 from gaphas.decorators import g_async
 from gi.repository import Gtk
@@ -24,6 +26,7 @@ from gaphor.event import (
     SessionShutdownRequested,
 )
 from gaphor.storage import storage
+from gaphor.storage.mergeconflict import split
 from gaphor.storage.parser import MergeConflictDetected
 from gaphor.ui.errorhandler import error_handler
 from gaphor.ui.filedialog import GAPHOR_FILTER, save_file_dialog
@@ -77,7 +80,7 @@ class FileManager(Service, ActionProvider):
         self.element_factory = element_factory
         self.modeling_language = modeling_language
         self.main_window = main_window
-        self._filename = None
+        self._filename: Path | None = None
 
         event_manager.subscribe(self._on_session_shutdown_request)
         event_manager.subscribe(self._on_session_created)
@@ -88,7 +91,7 @@ class FileManager(Service, ActionProvider):
         self.event_manager.unsubscribe(self._on_session_created)
 
     @property
-    def filename(self):
+    def filename(self) -> Path | None:
         """Return the current file name.
 
         This method is used by the filename property.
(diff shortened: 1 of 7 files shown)


amolenaar:
This is the solution to merge conflicts at the moment:

image

You can decide to either load the current or the incoming (new) model. 

This means that the change set code (part of this PR) is not actively used. However, we can experiment with it if we want.

gaphor/ui/filemanager.py
```diff
@@ -326,6 +359,31 @@ def response(answer):
             confirm_shutdown()
 
 
+def merge_conflict_dialog(window: Gtk.Window, filename, handler) -> None:
+    dialog = Adw.MessageDialog.new(
+        window,
+        gettext("Resolve merge conflict in model “{filename}”.").format(
```
danyeaw, line 365: Maybe it should say Resolve Merge Conflict?

gaphor/ui/filemanager.py
```diff
@@ -326,6 +359,31 @@ def response(answer):
...
+            filename=filename.name
+        ),
+    )
+    dialog.set_body(
+        gettext(
+            "It looks like the model contains a merge conflict. Do you want to resolve it?"
```
danyeaw, line 371: The model you are opening contains a merge conflict. Do you want to open the current change or the incoming change to the model?

gaphor/ui/filemanager.py
```diff
@@ -326,6 +359,31 @@ def response(answer):
...
+        )
+    )
+    dialog.add_response("cancel", gettext("Cancel"))
+    dialog.add_response("current", gettext("Use current model"))
+    dialog.add_response("incoming", gettext("Use incoming model"))
```
danyeaw, line 376: I think for these two just Current / Incoming, or Current Change / Incoming Change

danyeaw (approved):
The code looks really good, you made some nice simplifications to how we load models.

I was looking at the [HIG Patterns for Dialogs](https://developer.gnome.org/hig/patterns/feedback/dialogs.html), are we sure that showing a confirmation dialog is the best for this? Maybe we can fix this for the next iteration with manual being able to merge the changes.

Also, when I click Cancel, Gaphor is crashing with `terminated by signal SIGSEGV (Address boundary error)`

gaphor/ui/filemanager.py
amolenaar: That's better indeed.

gaphor/ui/filemanager.py
amolenaar: Shouldn't it contain one verb at least? It's an action after all?

gaphor/ui/statuswindow.py
```diff
@@ -76,13 +64,8 @@ def create_gtk3_window_with_progress_bar(self, frame, vbox, label):
         vbox.pack_start(label, True, True, 0)
         vbox.pack_start(self.progress_bar, expand=False, fill=False, padding=0)
 
+    @g_async()
```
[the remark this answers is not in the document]
amolenaar: Thanks :)

amolenaar:
> The code looks really good, you made some nice simplifications to how we load models.
> 
> I was looking at the [HIG Patterns for Dialogs](https://developer.gnome.org/hig/patterns/feedback/dialogs.html), are we sure that showing a confirmation dialog is the best for this? Maybe we can fix this for the next iteration with manual being able to merge the changes.

In the end I want to load the model and show an extra component (e.g. as part of the property editor compartment) with merge conflicts. 
The user should be able to choose which changes to incorporate. I've layed some foundational work in #1961. It still needs a fair amount of work though.

Is that what you're referring to?

> Also, when I click Cancel, Gaphor is crashing with `terminated by signal SIGSEGV (Address boundary error)`

That may be related to #1962. I pushed a fix.

gaphor/ui/filemanager.py
danyeaw: That is good, I think it would look better if each word was capitalized though

amolenaar: Indeed, it should, according to the HIG.

gaphor/ui/filemanager.py
amolenaar: I changed the name to `resolve_merge_conflict_dialog`.

danyeaw: This is what I meant, the actual heading of the message dialog. 
image

amolenaar: Ah. That's a lot nicer!

amolenaar: fixed

PR #1848, opened 2022-11-03
Storage module maintenance

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

The storage module has been stable for quite a while. However, the errors it produces are not very helpful to the user.

Issue Number: #388 

### What is the new behavior?

As a first stage towards #1845, let's make the storage module produce helpful error messages.

- [x] storage functions only accept (seekable) file objects. No file names anymore.
- [x] Add heuristic to detect merge conflicts, raises `MergeConflictDetected`
- [x] inform the user about an invalid model file
- [x] inform the user about a gaphor file with merge conflicts
- [x] Close session if model loading fails

image

### DIFF
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -57,11 +57,12 @@ event_manager = EventManager()
 with mock(["gi.repository.Gtk", "gi.repository.Gdk"]):
     modeling_language = ModelingLanguageService(event_manager=event_manager)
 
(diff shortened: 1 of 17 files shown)


danyeaw (approved):
These are some great looking updates for how we parse models and raise exceptions, nice one!

PR #1857, opened 2022-11-13
Update AppImage build to GTK4 and Python 3.11

danyeaw:
This PR updates the AppImage to use GTK and Python 3.11.

TODO: 
- [x] Push new tagged version of the docker container

One thing I am not really happy with is that JHBuild is building Python files to `/local/lib/python3.11` instead of `/lib/python3.11`. I have worked around it by overriding the `PYTHONPATH` and adding an extra couple of paths to `PATH`. I will submit an issue about this, I can't seem to track down where this local path is coming from.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

AppImage is GTK3 only with Python 3.10

Issue Number: N/A

### What is the new behavior?
AppImage is GTK4 with Python 3.11

- [ ] Yes
- [X] No


danyeaw:
Interesting error on CI `Failed to load shared library 'libadwaita-1.so.0' referenced by the typelib: libadwaita-1.so.0: cannot open shared object file: No such file or directory`. I am able to download the AppImage, and it runs fine for me.

amolenaar:
> Interesting error on CI `Failed to load shared library 'libadwaita-1.so.0' referenced by the typelib: libadwaita-1.so.0: cannot open shared object file: No such file or directory`. I am able to download the AppImage, and it runs fine for me.

It may be that it's taking your system libadwaita instead of the packaged one. Maybe `LD_LIBRARY_PATH` needs to be set?

danyeaw:
It is set, the library is there, and running `from gi.repository import Adw` works with no errors from the Docker container.

danyeaw:
It did end up being the LD_LIBRARY_PATH with this `/install/local/lib` needing to be added because Python libraries are going there right now for some reason. :+1:

amolenaar:
It's a perculiar issue. I tested on a Fedora 36 box, and I see it has loaded the right Python (3.11) and GTK (4.8.3) version. Then why is it not able to find libadwaita?

BTW. It's constantly complaining about GL renderers. It looks like enhanced GL support does not work at least.

danyeaw:
https://github.com/pyinstaller/pyinstaller/pull/7278 to fix the issue with PyInstaller.

What is the GL renderer issue you are seeing?

danyeaw:
@amolenaar Errors are fixed and the AppImage is working great for me, would you mind taking a look when you get a chance?

amolenaar:
@danyeaw Amazing how you managed to track the issue down to PyInstaller :pray:. And fixed it!

I tested on Fedora 35 and Ubuntu 20.04, and on both platforms it _just works_.

PR #1871, opened 2022-11-27
Add some architectural rules to Gaphor

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

We have no way, except for visual code review, to detect unwanted dependencies between parts of Gaphor.

With the implementation of Jupyter notebook support, I want to be able to work with Gaphor from non-ui contexts without a dependency on GTK and friends.

### What is the new behavior?

With my colleague @jwbargsten, we implemented [a little tool](https://github.com/jwbargsten/pytest-archon) that checks for architectural boundaries in the application. With those boundaries we have a better chance to keep the code base clean and organized.

As a result, I moved the help service (about and shortcuts windows) to `gaphor.ui`. Also an unneeded event emission was detected in the copy service.

- [ ] Yes
- [x] No


tests/test_architecture.py
```diff
@@ -0,0 +1,130 @@
...
+    "gaphor.action",
+    "gaphor.entrypoint",
+    "gaphor.i18n",
+    "gaphor.transaction",
+    "gaphor.event",
+]
+
+UI_LIBRARIES = [
+    "gi.repository.Gdk",
+    "gi.repository.Gtk",
```
danyeaw, line 18: Should we add gi.repository.Adw to the UI libraries?

danyeaw (approved):
I love this, adding some guardrails as rules helps us keep things more consistent :+1:

tests/test_architecture.py
amolenaar: Can do that, just to be on the safe side.

Issue #1879, opened 2022-11-28
Improve welcome experience


bertob:
In the current stable version, opening the app initially opens just this dialog:

![image](https://user-images.githubusercontent.com/1908896/204263761-b2e0fecc-3b62-4cf5-a574-749a5f98b018.png)

This is not ideal, because it puts one extra step in front of the actual app opening, which feels like a hurdle. I'd recommend immediately opening the real app window, and either integrating the welcome experience into the main window (like e.g. Builder does on its new tab page), or having the welcome dialog on top of the main window.

The dialog itself could also be improved:
- There is no context about the options other than the names and logos, which doesn't feel approachable. A description for each of the options would be helpful to make the choice easier, perhaps with example use cases.
- The logos are quite prominent, which isn't great since some of them are a bit ugly and they're stylistically inconsistent between each other. It might help to have them all inside white circles or something, to make them look more uniform. Arranging them as a list rather than a grid, with smaller icons, could also help.
- The blue focus state feels off-style, we usually use semi-transparent text color for that
- The dialog feels a bit empty and top-heavy, I'd resize it and, depending on whether you go with list or grid, potentially center the content vertically

amolenaar:
Hi @bertob,

Thanks a lot for your feedback!

I based this loosely on Builder. When I start it, it welcomes me with a "Open a Project" window. Once you have opened a model, it will show the previously opened models, instead of this new-model window. I miss a button to open a model from the file system, though.

The size is based on the recent-models (GNOME builder 3). I suppose we can put them all in one window, in a list and not have separate pages for existing and new models.

amolenaar:
@bertob I did some work on the greeter. Instead of the blocks, it now shows the options for creating a new model from a template as a list, with the icon in front.

See #1892. Does this come close(r) to a visually appealing design?

bertob:
Commented on the MR :)

PR #1892, opened 2022-11-29
Simplify the greeter and provide more info to new users

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

The greeter provides little information for new users

Issue Number: #1879

### What is the new behavior?

This PR proposes to make the greeter simpler. I think we should not drop users in an empty model. This worked fine in the past, but we like to give users a head start by using a template.

image

When no recent files are present:

image


bertob:
Looks good to me structurally, but the spacing inside the rows looks a bit odd, the rows are very tall and there's a ton of spacing between title and subtitle. I'd use standard libadwaita row styling here.

gaphor/ui/greeter.py
```diff
@@ -8,25 +8,55 @@
...
+    ModelTemplate(
+        gettext("UML"),
+        gettext("A Unified Modeling Language template/example."),
+        "UML",
+        "UML",
+        "uml.gaphor",
+    ),
+    ModelTemplate(
+        gettext("SysML"),
+        gettext("Systems Modeling template"),
```
danyeaw, line 41: I think we should either spell them all out completely or abbreviate them all. SysML is Systems Modeling Language and RAAML is Risk Analysis and Assessment Modeling Language. Why does the UML one say template/example and the others just say template?

danyeaw (commented):
I really like where this is heading!

amolenaar:
I updated the GTK4 layout to use `AdwPreferencesGroup`. It looks a bit more compact. I added a little extra padding on the rows.

image

gaphor/ui/greeter.py
amolenaar: I quickly dropped some text.

RAAML spelled out is a bit too much text IMHO. Currently the description does not add a lot. 

Something like "Systems Modeling Language template" does not say anything about what you can do with it. I think that the term "template" is also a bit misplaced. You're creating a new model _based on a template_, not the template.

bertob:
Nice, looks much better! A few more minor things:
- I'd simplify the titles a bit and use title case, e.g. "Recently Opened" and "Create New Model"
- I'd move the Open button to the left side of the headerbar and change the lable to "Open Model..."
- It'd be good to already have access to the app's primary menu in this state, so you can e.g. check the version in the about dialog

amolenaar:
@bertob Thanks a lot for your quick feedback.

Something like this (hamburger will follow)?

image

bertob:
Yup, looks good :+1:

amolenaar:
With hamburger menu:

image

bertob:
Hmm, why is there an extra "Close" item in the menu?

I was thinking of something like this: 

```
Open Model...
---
Keyboard Shortcuts
About Gaphor
```

amolenaar:
I had a look at GNOME Builder. It also had a close menu item 😅. There's not really any keyboard shortcuts in this window.

bertob:
Hmmm, generally menus should have at least 3 items :D  

I guess it'd be fine to show all keyboard shortcuts though since at least Ctrl+W, Ctrl+Q, F10, and the like would work here too.

bertob:
With the keyboard entry this is good to go, I think.

danyeaw (approved):
This is a really great improvement! Thanks @amolenaar for the great implementation and @bertob for the original idea and guidance on the design.

Issue #1839, opened 2022-10-30
Migrate macOS version to GTK4


amolenaar:
### Is your feature request related to a problem?

From our main releases, the macOS build is the still running on GTK 3.

### Describe the solution you'd like

We should decide if and when  we switch to GTK 4.

There are some pro's and cons:

* **pro** The app becomes a _lot_ more responsive.
* **pro** We're one step closer to phasing out GTK 3.
* **con** Drag and drop support does not work (currently drag/drop event controllers are not activated for macOS). DnD can be disabled in our code, but text entry fields also support DnD. ~When you (accidentally) drag some text from a text field, the app freezes.~ DnD is required to allow for dragging and resizing the window, so we can't disable it completely (drag threshold > screen size)

The following GTK4 issues need to be resolved:

- [x] [macos: drag-n-drop is not fully working](https://gitlab.gnome.org/GNOME/gtk/-/issues/3423) - input freezes after drop.
  * MR: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/5291  (_merged_)
- [x] [macos: support for keybinding translations](https://gitlab.gnome.org/GNOME/gtk/-/issues/3421) - currently `Ctrl` is the primary modifier key on macOS, instead of `Cmd`.
  * #1901 fixes it for our own keybindings. 
  * #1911 takes care of (some of) the default key bindings for text entry.
  * https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/5296 (_merged_) shown the command symbol in the Keyboard Shortcuts window.
- [x] Add keybinding: `Cmd`+click for multi-select in model browser. Doesn't work: Ctrl-click is handled in an event controller in [`GtkListItemWidget`](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtklistitemwidget.c#L450).

### Describe alternatives you've considered

We could stay with GTK3 for now, but new features (like HARA table views) will probably not be supported.

### Additional context

~The AppImage build should either be migrated to GTK4 or discontinued.~ AppImage has been ported to GTK4, macOS is the only platform still on GTK3.

amolenaar:
As a side note: the version of Cairo provided by Homebrew is 1.16.0, which is quite old.

amolenaar:
GTK 4.9.3 has been released with DnD on macOS fixed.

danyeaw:
Yay, that's so exciting!

amolenaar:
We could... build GTK ourselves :smile:

[...]

PR #1901, opened 2022-12-05
Fix macOS shortcuts on GTK 4

amolenaar:
### PR Type

- [] Bug fix
- [ ] Feature

On GTK4 no magic mapping is done for us on macOS, where `Command` serves as primary modifier.

Issue Number: #1839

### What is the new behavior?

We remap the `Primary` modifier on macOS to the `Meta` key (`Command`).

image

### DIFF
--- a/gaphor/ui/actiongroup.py
+++ b/gaphor/ui/actiongroup.py
@@ -1,8 +1,12 @@
+from __future__ import annotations
+
+import sys
 from typing import NamedTuple
 
 from gi.repository import Gio, GLib, Gtk
 
 from gaphor.abc import ActionProvider
+from gaphor.action import action
 
 
 def apply_application_actions(component_registry, gtk_app):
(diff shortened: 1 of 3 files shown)


mathiascode:
GTK's built-in shortcuts (e.g. Ctrl+C in textviews) would still be a problem, right?

danyeaw (approved):
Nice improvement for macOS @amolenaar

amolenaar:
> GTK's built-in shortcuts (e.g. Ctrl+C in textviews) would still be a problem, right?

Fair point 🤔. Not sure how easily those can be remapped. It's mainly text views and entry widgets, right? I assume the tab bar has some shortcuts as well.

amolenaar:
I think we can add additional keyboard shortcuts via `Gtk.Widget.add_shortcut()` at class level. The original (`ctrl`-based) keybindings can stay.

amolenaar:
This is the list of supported keybindings on macOS: https://support.apple.com/en-us/HT201236.

PR #1931, opened 2022-12-17
Upgrade pycairo

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

### What is the new behavior?

Upgrade pycairo to 1.23.0 and fix the local flatpak build so it works with an upgraded pycairo.

### DIFF
--- a/org.gaphor.Gaphor.json
+++ b/org.gaphor.Gaphor.json
@@ -44,7 +44,7 @@
       "build-commands": [
         "pip3 install --prefix=${FLATPAK_DEST} poetry",
         "${FLATPAK_DEST}/bin/poetry export --output=requirements.txt",
-        "pip3 install --prefix=${FLATPAK_DEST} -r requirements.txt"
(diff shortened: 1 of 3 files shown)


org.gaphor.Gaphor.json
```diff
@@ -44,7 +44,7 @@
       "build-commands": [
         "pip3 install --prefix=${FLATPAK_DEST} poetry",
         "${FLATPAK_DEST}/bin/poetry export --output=requirements.txt",
-        "pip3 install --prefix=${FLATPAK_DEST} -r requirements.txt"
+        "pip3 install --ignore-installed --prefix=${FLATPAK_DEST} -r requirements.txt"
```
danyeaw, line 47: For my understanding, is ignore-installed overwriting something where force-reinstall couldn't be used?

amolenaar: Force-reinstall tries to remove the old package. That does not work, since there is no list of files to remove.

danyeaw: Got it, thanks!

danyeaw (approved):

PR #1941, opened 2022-12-20
Remove querymixin from modeling lists

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

We could use string based queries, but never used those.

### What is the new behavior?

Just use a list comprehension:

```python
a = instancespec.ownedAttribute["it.name == 'specification'"][0]
a = next(it for it in instancespec.ownedAttribute if it.name == 'specification')
```

- [x] Yes
- [ ] No


danyeaw (approved):
Nice simplification!

Issue #1928, opened 2022-12-16
The files does not contain a valid Gaphor model


mcdigregorio:
### Describe the bug
After upgrading to Gaphor 2.13.0 from 2.11.0, getting a "The files does not contain a valid Gaphor model" message

### To Reproduce
1. Open Gaphor 2.13.0
2. Try to open attached DemandSupply.gaphor file

### Expected behavior
Expected model to open cleanly.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [x] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.13.0

### Additional information
It seems like all of my issues are tied to AssociationItems. I was able to setup a debug environment and grab some exceptions which led me to the below "bad" AssociationItems by id. I believe the core issues were missing "diagram" elements and missing "subject" elements. 

1. 58c9cc3c-62fb-11ed-8ada-acde48001122
    - Missing diagram element
2. 6510dd3c-62fb-11ed-8ada-acde48001122
    - Missing diagram element
3. 5fd55582-62fb-11ed-8ada-acde48001122
    - Missing subject element
4. 764984be-62fb-11ed-8ada-acde48001122
    - Missing subject element
5. 77957e0e-62fb-11ed-8ada-acde48001122
    - Missing subject element
6. 4c2ef7da-62fc-11ed-8ada-acde48001122
    - Missing subject element
8. 51e1b6e0-62fc-11ed-8ada-acde48001122
    - Missing subject element
9. 631ae210-62fc-11ed-8ada-acde48001122
    - Missing subject element
    
 If I remove these elements from the .gaphor file I am able to open the model again. For reference I attached that working version in the .zip as DemandSupply_Working.gaphor.
 
 Does this seem like an issue tied to a corrupt gaphor file? Maybe I had a bad save prior?
 If so, is there any way to recover those "corrupt" elements?

At this point, losing these elements isn't a big deal for me, but just trying to decrease the likelihood it happens again.

[GaphorFiles.zip](https://github.com/gaphor/gaphor/files/10248153/GaphorFiles.zip)

amolenaar:
@mcdigregorio I'll have a look.

amolenaar:
I had a look. It contained 2 associations that were missing a diagram and an association (+ 2 properties/ends) that had no properties at all. I'm curious how that happened, so far our property based tests haven't found this issue.

[...]

PR #1942, opened 2022-12-20
Fix loading error

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

There is still a possibility that a model is saved in a state that can't be loaded.

Issue Number: #1928

### What is the new behavior?

* Add an extra check for presentation elements that lack a diagram reference. We can easily get rid of those *iff* they are not referenced.
* Add an extra check for association connections: do not assume `Association.memberEnd` is always populated.
* Add a priority lane for event handlers. This is mainly to allow Undo Manager to receive all events
    
My hypothesis on why things still can break is because not all events are properly registered at the undo manager. In that case we can no do a full undo, and the model may remain in an inconsistent state. The effect (e.g. a property is set) is applied directly, then we notify by means of an event. That event will eventually be handled, or not, if an exception occurs.

To remedy this, we can basically do two things:

1. delay the (model) change until just before the event is about to be    fired (e.i. change events to commands).
2. Make sure all events are captured by the undo manager.

Option 1. is quite complicated, since a simple relation change canresults in a combination of set and delete events.
Option 2. is a bit hacky as well, since we add a "priotity lane" to the event manager.

### DIFF
--- a/gaphor/UML/classes/classconnect.py
+++ b/gaphor/UML/classes/classconnect.py
@@ -79,12 +79,15 @@ class AssociationConnect(RelationshipConnect):
         subject = line.subject
 
         def member_ends_match(subject):
-            return (
-                head_subject is subject.memberEnd[0].type
-                and tail_subject is subject.memberEnd[1].type
-            ) or (
-                head_subject is subject.memberEnd[1].type
-                and tail_subject is subject.memberEnd[0].type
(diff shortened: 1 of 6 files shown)


danyeaw (approved):
This looks good, nice job continuing to improve model reliability 👍

amolenaar:
I hope it pays off 😅

PR #1951, opened 2022-12-25
Add completion provider for CSS functions

vanillajonathan:
### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/storage/mergeconflict.py
+++ b/gaphor/storage/mergeconflict.py
@@ -1,0 +1,29 @@
+import io
+
+
+def split(fd: io.TextIOBase, current: io.TextIOBase, incoming: io.TextIOBase):
+    CURRENT = 1
+    INCOMING = 2
+    BOTH = CURRENT | INCOMING
+    state = BOTH
+    found_merge_conflict = False
+    for line in fd.readlines():
+        if state == BOTH and line.startswith("<<<<<<<"):
+            state = CURRENT
+            found_merge_conflict = True
+        elif state == CURRENT and line.startswith("======="):
+            state = INCOMING
+        elif state == INCOMING and line.startswith(">>>>>>>"):
+            state = BOTH
+        else:
+            if state & CURRENT:
+                current.write(line)
+            if state & INCOMING:
+                incoming.write(line)
+
+    current.seek(0)
+    incoming.seek(0)
+    if found_merge_conflict:
+        return current, incoming
+    incoming.close()
+    return current, None
--- a/gaphor/storage/tests/test_mergeconflict.py
+++ b/gaphor/storage/tests/test_mergeconflict.py
@@ -1,0 +1,91 @@
+import io
+import textwrap
+
+from gaphor.storage.mergeconflict import split
+
+
+def test_splitter():
+    fd = io.StringIO()
+    fd.write(
+        textwrap.dedent(
+            """\
+        <<<<<<< HEAD
+        current line
+        =======
+        incoming line
+        >>>>>>> 3b251489 (Move tool selection to toolbox)
+        """
+        )
+    )
+    fd.seek(0)
+    current = io.StringIO()
+    incoming = io.StringIO()
+    split(fd, current, incoming)
+
+    assert current.read() == "current line\n"
+    assert incoming.read() == "incoming line\n"
+
+
+def test_splitter_windows_line_endings():
+    fd = io.StringIO()
+    fd.write(
+        textwrap.dedent(
+            """\
+        <<<<<<< HEAD\r
+        current line\r
+        =======\r
+        incoming line\r
+        >>>>>>> 3b251489 (Move tool selection to toolbox)\r
+        """
+        )
+    )
+    fd.seek(0)
+    current = io.StringIO()
+    incoming = io.StringIO()
+    split(fd, current, incoming)
+
+    assert current.read() == "current line\r\n"
+    assert incoming.read() == "incoming line\r\n"
+
+
+def test_split_with_shared_content():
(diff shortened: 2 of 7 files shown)


gaphor/ui/css_function_completion_provider.py
```diff
@@ -0,0 +1,100 @@
...
+        "linear-gradient",
+        "max",
+        "min",
+        "radial-gradient",
+        "repeating-conic-gradient",
+        "repeating-linear-gradient",
+        "repeating-radial-gradient",
+        "rgb",
+        "rgba",
+        "var",
```
amolenaar, line 22: I think we should limit this to the functions that Gaphor supports: `hsl`, `hsla`, `rgb`, and `rgba`.

gaphor/ui/css_function_completion_provider.py
```diff
@@ -0,0 +1,100 @@
+from gi.repository import Gio, GObject, Gtk, GtkSource
```
amolenaar, line 1: In Gaphor we tend to use file names of concatenated words, no undescore (e.g. `treecomponent.py`).

gaphor/ui/css_function_completion_provider.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class CssFunctionCompletionProvider(GObject.GObject, GtkSource.CompletionProvider):
```
amolenaar, line 4: The code for this class is almost a copy of `CssPropertyCompletionProvider`. I think it's pretty simple to create a simple parameterizable version of a completion provider. Then both completion provider can be in the same file (`gaphor/ui/csscompletion.py`).

amolenaar (changes requested):
Really cool to have function completion as well. Thanks a lot, @vanillajonathan,

I think the difference with the current css property completion provider are very small. I think we can DRY this code and have all completion code in the same module.

gaphor/ui/css_function_completion_provider.py
vanillajonathan: Well, I would guess anything that WebKit support would work?
I think the other functions would probably be possible to use too, even though I think they're unlikely to be much used, as many are for advanced uses and seldom used.

gaphor/ui/css_function_completion_provider.py
vanillajonathan: So then this would be `cssfunctioncompletionprovider.py`, yuck!
Quite hard to read.

gaphor/ui/css_function_completion_provider.py
vanillajonathan: Yes, it is, right now, but but GtkSourceView is built to support multiple active completion providers, and we can modify these so they differ from each other.
Example it could append `()` after the function name when inserted. It could append a `:` after the attribute name when inserted. 

They also have different [`do_get_title`](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/vfunc.CompletionProvider.get_title.html) and [`do_get_priority`](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/vfunc.CompletionProvider.get_priority.html). The title isn't used by GtkSourceView today, but in future versions it might be. The priority can be used for ordering, example so we can have attributes ordered before functions.

Later, we can add completion providers for colors (they will have a different priority too), and for units for which we might to present when a number is entered, so it is good to keep completions for different things in different completion providers.

gaphor/ui/css_function_completion_provider.py
amolenaar: > Well, I would guess anything that WebKit support would work? 

We're not using webkit for rendering. Instead we use Cairo and a the items render themselves on the canvas, with some styling.

gaphor/ui/css_function_completion_provider.py
amolenaar: I like the idea of appending `()` and `:`.

> The priority can be used for ordering, example so we can have attributes ordered before functions.

Thinks like title and priority can be past as constructor parameters to a generic base class, where we handle the common boilerplate.

> Later, we can add completion providers for colors (they will have a different priority too), and for units for which we might to present when a number is entered, so it is good to keep completions for different things in different completion providers.

I see your point. Indeed completers could defer. However, I would not focus to much on later :wink: .

gaphor/ui/css_function_completion_provider.py
amolenaar: Both CompletionProvider classes can be added to `csscompletion.py`.

vanillajonathan: True!

gaphor/ui/css_function_completion_provider.py
vanillajonathan: Oh, I am aware which CSS properties and functions can be used in Cairo.

gaphor/ui/css_function_completion_provider.py
vanillajonathan: > I like the idea of appending () and :.

The appending of characters upon insertion would be done in the [`do_activate`](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/vfunc.CompletionProvider.activate.html) method.

Different providers could have different implementations for the [`is_trigger`](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/vfunc.CompletionProvider.is_trigger.html) method, e.g. to show units (such as `px`) when the user types a digit.

PR #1957, opened 2022-12-26
Fix macos build

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Gaphor build crashes on GTK 3 (macOS) because of an unknown import.

### What is the new behavior?

Add lazy annotation resolving.

Also limit the number of supported CSS properties to what Gaphor actually supports.


gaphor/ui/csscompletion.py
```diff
@@ -0,0 +1,86 @@
+from __future__ import annotations
+
+from gi.repository import Gio, GObject, Gtk, GtkSource
+
+from gaphor.core.styling import Style
+
+
+class CssPropertyCompletionProvider(GObject.GObject, GtkSource.CompletionProvider):
+    PROPERTIES = sorted(Style.__optional_keys__)
```
danyeaw, line 9: TIL about `__optional_keys__` from [PEP 655](https://peps.python.org/pep-0655/) :+1:

danyeaw (commented):
This looks good. For my understanding, could you key me in on what makes the loading lazy now?

amolenaar:
It's the future annotation that actually fixed it.

PR #1447, opened 2022-04-03
Add HARA Table Support

danyeaw:
This PR adds initial table support through creation of a Hazard Analysis & Risk Assessment (HARA) table for RAAML.

### PR Checklist
Please check if your PR fulfills the following requirements:

- [X] I have read, and I am following the [Contributor guide](https://github.com/gaphor/gaphor/blob/main/CONTRIBUTING.md)
- [X] I have read, and I understand the [Code of Conduct](https://github.com/gaphor/gaphor/blob/main/CODE_OF_CONDUCT.md)

### PR Type
What kind of change does this PR introduce?

- [ ] Bug fix
- [X] Feature
- [ ] Chore (refactoring, formatting, local variables, other cleanup)
- [ ] Documentation content changes

### What is the current behavior?
It is not possible to create a HARA

Issue Number: #927 

### What is the new behavior?
Support for a HARA diagram and a table view of it

### Does this PR introduce a breaking change?
- [ ] Yes
- [X] No

PR #1961, opened 2022-12-27
Manually resolve merge conflicts

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

When a merge conflict happens, you're in trouble.

Issue Number: #388 

### What is the new behavior?

From the merge conflict a change set is distilled. This change set can be applied to the model by the user. This allows for fine grained control over the changes made to the "resolved" model.

- [ ] Yes
- [x] No

### Other information

Only allow to apply complete diagram changes? Not per element/presentation? If we provide only the option to update a diagram as a whole, we stand a better chance to maintain consistency, I think.

Integrate the model merge feature in the property pane (no separate one).

Deal with composite relations. Those should be created as part of another element.

- [x] Association with its association ends (Property)
- [x] Extension with associated Properties
- [x] Activity with Parameters
- [x] Partition with swim lanes -> This automatically creates an Activity if it doesn't exist yet. Same goes for Actions, etc.
- [x] NamedElement represented by Partition -> normal association
- [x] Class (and similar) with attributes and operations
- [x] Connector with its connector ends and information flow
- [x] State with entry/exit/do activity
- [x] Transition with a guard
- [x] Applied stereotypes
- [x] Deal with Namespace elements (e.g. Package, Activity) that are not visible in (the same) diagram

I hope that the common theme here is composition: we can check if a relation is an composite aggregation; if so, add the element as a nested element.

### DIFF
--- a/gaphor/ui/modelmerge/editor.py
+++ b/gaphor/ui/modelmerge/editor.py
@@ -1,8 +1,14 @@
 from __future__ import annotations
 
 from gi.repository import Gio, Gtk
+from gaphas.decorators import nonrecursive
 
-from gaphor.event import ModelLoaded
+from gaphor.event import (
+    ModelLoaded,
+    TransactionBegin,
+    TransactionCommit,
+    TransactionRollback,
+)
 from gaphor.core.modeling import PendingChange
 from gaphor.core.changeset.apply import apply_change, applicable
 from gaphor.i18n import translated_ui_string
@@ -19,6 +25,7 @@ class ModelMerge:
         self.model = Gio.ListStore.new(Node.__gtype__)
         self.selection = None
         self.tree_view = None
+        self._tx_depth = 0
 
     @property
     def needs_merge(self):
@@ -41,26 +48,8 @@ class ModelMerge:
 
         self.selection = Gtk.SingleSelection.new(tree_model)
 
-        def on_apply(change_node: Node, sync=True):
-            if not change_node.elements:
-                return
-
-            with Transaction(self.event_manager):
-                for element in change_node.elements:
-                    if applicable(element, self.element_factory):
-                        apply_change(
-                            element, self.element_factory, self.modeling_language
-                        )
-                if change_node.children:
-                    for n in change_node.children:
-                        on_apply(n, sync=False)
-
-            if sync:
-                for item in self.model:
-                    item.sync()
-
         factory = Gtk.SignalListItemFactory.new()
-        factory.connect("setup", list_item_factory_setup, on_apply)
+        factory.connect("setup", list_item_factory_setup, self.apply)
 
         self.tree_view = builder.get_object("modelmerge")
         self.tree_view.set_model(self.selection)
@@ -75,6 +64,24 @@ class ModelMerge:
 
     def close(self):
         self.event_manager.unsubscribe(self.on_model_loaded)
+        self.event_manager.unsubscribe(self.on_model_updated)
+
+    @nonrecursive
+    def apply(self, change_node: Node | None):
+        def do_apply(node):
+            for element in node.elements:
+                if applicable(element, self.element_factory):
+                    apply_change(element, self.element_factory, self.modeling_language)
+            if node.children:
+                for n in node.children:
+                    do_apply(n)
+
+        if change_node:
+            with Transaction(self.event_manager):
+                do_apply(change_node)
+
+        for item in self.model:
+            item.sync()
 
     def on_resolve_merge(self, _button):
         pending_changes = self.element_factory.lselect(PendingChange)
@@ -82,10 +89,18 @@ class ModelMerge:
         with Transaction(self.event_manager):
             for change in pending_changes:
                 change.unlink()
+        self.close()
 
     @event_handler(ModelLoaded)
     def on_model_loaded(self, event):
+        self.event_manager.subscribe(self.on_model_updated)
         self.refresh_model()
+
+    @event_handler(TransactionBegin, TransactionCommit, TransactionRollback)
+    def on_model_updated(self, event):
+        self._tx_depth += +1 if isinstance(event, TransactionBegin) else -1
+        if self._tx_depth == 0:
+            self.apply(None)
 
 
 def list_item_factory_setup(_factory, list_item, on_apply):
@@ -101,8 +116,9 @@ def list_item_factory_setup(_factory, list_item, on_apply):
 
     def on_active(button, _gparam):
         node = list_item.get_item().get_item()
-        if button.get_active() and node:
+        if button.get_active() and button.get_sensitive() and not node.applied:
             on_apply(node)
-        node.sync()
 
     apply.connect("notify::active", on_active)
+
+    # TODO: popup
--- a/gaphor/ui/modelmerge/editor.py
+++ b/gaphor/ui/modelmerge/editor.py
@@ -52,7 +52,7 @@ class ModelMerge(UIComponent):
         tree_model = Gtk.TreeListModel.new(
             self.model,
             passthrough=False,
-            autoexpand=True,
+            autoexpand=False,
             create_func=lambda node, _: node.children,
             user_data=None,
         )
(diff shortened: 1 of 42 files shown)


amolenaar:
I'm trying to figure out what the merge window should show. At the top level there is always an element. Preferably a presentation, otherwise the element that's added/removed/updated.

For presentations, the diagram should exist or should otherwise be added separately? Or should the action be added to every presentation that has to be placed on the diagram?

The merge editor component should have a maximum width and labels should wrap.

Every element should allow for a context menu (right click), and allow to show the element in a diagram (similar to model editor).

### Add element

- [ ] Add element of type _{type}_
  - [ ] Update attribute _{name}_ to _{new value}_

### Add presentation

For presentations, make a new subject a nested item:

- [ ] Add element of type _{type}_ to diagram _{diagram name}_
  - [ ] Add subject of type _{type}_
    - [ ] Update attribute _{name}_ to _{new value}_
  - [ ] Update attribute _{name}_ to _{new value}_

For existing subjects:

- [ ] Add element of type _{type}_ to diagram _{diagram name}_
  - [ ] Add relation _subject_ to _{element name}_
    - [ ] Update attribute _{name}_ to _{new value}_
  - [ ] Update attribute _{name}_ to _{new value}_

[optional] Add a presentation to a not (yet) existing diagram:

- [ ] Add element of type _{type}_ to diagram _{diagram name}_
  - [ ] Add element of type _Diagram_
    - [ ] Update attribute _{name}_ to _{new value}_ 
  - [ ] Add subject of type _{type}_
    - [ ] Update attribute _{name}_ to _{new value}_
  - [ ] Update attribute _{name}_ to _{new value}_

### Remove element

- [ ] Remove element _{element qname}_
  - [ ] Remove relation _{property}_ to _{element name}_

### Update attribute

- [ ] Update element _{element qname}_
  - [ ] Update attribute _{name}_ from _{old value}_ to _{new value}_

NB. "update element" is no action in itself, it's only a container for the actual attribute/relation changes.

### Update a reference

- [ ] Update element _{element qname}_
  - [ ] Add relation _{property}_ to _{element name}_
  - [ ] Remove relation _{property}_ to _{element name}_
  - [ ] Update relation _{property}_ to _{element name}_  (for multiplicity [0..1])
- [ ] Remove element _{element qname}_
  - [ ] Remove relation _{property}_ to _{element name}_

amolenaar:
As a test case I tried to resolve merge conflicts for https://github.com/gaphor/gaphor/pull/1447.

What I did not consider is that there may be changes that *have* been merged, which cause the current model to be invalid due to unsolvable references. It may also be that part of the model has been removed, although I expect that to be less of a problem.

What I need instead is the *original* current file and the *original* incoming file. That way I am sure that the models that are loaded are valid at least.

One way would be to require the merge to be done with `--conflict=diff3` ([ref](https://stackoverflow.com/questions/8832201/git-how-to-get-ours-theirs-original-for-merge-conflicts), but who does that really.

amolenaar:
The changes are now shown in a separate pane.

image

It sort of works (changes can be applied), but I think it's not really usable at this point. The amount of data is overwhelming and a lot of things (such as `ValueChange`s) can probably be applied as soon as an element is created.

I used the RAAML model from #1447 as an (acceptance) test case. You can load this model with 

```sh
python tests/run_with_merge_conflict.py
```

@danyeaw Any ideas on how to improve the merge experience?

amolenaar:
Currently every line has a checkbox. This is very overwhelming. The difference between a clickable and a non-clickable checkbox is a bit nuanced. 

I think the  merge window should be [diagram centric](https://docs.gaphor.org/en/latest/design_principles.html#keep-the-model-in-sync), for starters. If need arises we can add different views.

I propose we can drop a lot of the nested check boxes. This leaves us with less options, but is also less confusing. Also there's a smaller likelihood of errors. E.g. an Association needs 2 Property objects. Those should be automatically created as part of the association, given the merge algorithm can figure our which elements need to be created.

For a new diagram with a new model element:

- [ ] Create diagram _{diagram name}_
  - Update attribute _{name}_ to _{new value}_
  - Add presentation for _{type}_
    - Update attribute _{name}_ to _{new value}_
    - Add element _{type}_
      - Update presentation relationship

For an element on an existing diagram:

- Update diagram _{diagram name}_
  - [ ] Add presentation for _{type}_
    - Update attribute _{name}_ to _{new value}_
    - Add element _{type}_
      - Update presentation relationship

For an element not related to a diagram:

- [ ] Add element _{type}_
   - Add _{property}_ relation to _{element name}_

For updates on an existing element:

 - Update element _{element qname}_
    - [ ] Remove relation _{property}_ to _{element name}_

danyeaw:
Hi @amolenaar I agree that things should be diagram centric. This is coming along nicely, even though this is a hard problem to solve.

I would think we would want a merge to be automatic unless two parts of the model are changed at the same time.

1. Newly created objects normally should merge. For example, if a new diagram is created, it shouldn't conflict unless their branch also created a diagram with the same name.
2. Changed object should merge unless they are changed in both branches
3. Deletion of objects should merge unless that object is changed in mine and deleted in theirs

amolenaar:
@danyeaw. Thanks. This makes sense indeed. I think I was too much focused on merging the current and incoming versions. To make this work, I should let Gaphor compare with the common ancestor commit. This is available.

It would make the change list a lot shorter, I suppose :).

amolenaar:
Consider the following case:

A generalization is drawn on a diagram from class `A` to class `B`.
In the current model it's changed from class `A` to class `C`;
in the incoming model it's changed from class `A` to class `D`.

In all three cases the generalization item is backed by a different `Generalization` instance.
Now we can't simply apply all newly created elements to the model. Applying the incoming change would imply that the generalization from `A` to `C` has to be removed, which is normally taken care of by the sanitizer service.

For associations (with their properties at both ends, it's even more complex, I suppose.

amolenaar:
There you go.

image

Merging works now. Elements are grouped in diagrams, where possible and composition is taken into account (diagram has diagram items, has subject, has properties, etc).

The merge window is put in place of the editor windows. This is partly to keep the design simple and partly to nudge people into resolving merge conflicts first.

A model can be saved with merge conflicts (pending changes). That way you can do an actual merge later or in stages. I think that's a neat feature. When you load the model, the merge editor will be populated again and you can continue merging.

You can try yourself by running:

```
python tests/run_with_merge_conflict.py
```

Which will load the model changes from #1447. (this merge conflict is actually a bit to big).

Anyway, I like to keep it like this, since the PR is **huge**.

amolenaar:
One more thing...

It needs some special handling for the StyleSheet class: a model is supposed to have only one.

gaphor/ui/modelmerge/editor.py
```diff
@@ -0,0 +1,124 @@
...
+    apply = builder.get_object("apply")
+
+    def on_active(button, _gparam):
+        node = list_item.get_item().get_item()
+        if button.get_active() and button.get_sensitive() and not node.applied:
+            on_apply(node)
+
+    apply.connect("notify::active", on_active)
+
+    # TODO: popup
```
danyeaw, line 124: Is this TODO still needed?

gaphor/ui/modelmerge/organize.py
```diff
@@ -0,0 +1,350 @@
...
+                ),
+                None,
+            )
+        )
+        else None
+    )
+
+    op = change.op
+    if isinstance(change, ElementChange) and change.element_name.endswith("Item"):
+        # TODO: find subject type
```
danyeaw, line 269: Is this TODO still needed?

danyeaw (approved):
This looks amazing @amolenaar, I can't believe we now have full merge capability!

Possible future enhancements (not blocking this merge):

1. I noticed for the example that if there is a use case to accept almost everything, except for a change or two - it would require a lot of clicking. Maybe we need to allow Shift select of multiple things, or maybe a right click accept everything except?
2. If the property editor isn't open, should we automatically open it?

gaphor/ui/modelmerge/editor.py
amolenaar: Not for this PR probably. I thought it would be nice to open existing diagrams and show elements in diagrams or the model browser.

gaphor/ui/modelmerge/organize.py
amolenaar: Cleaned.

amolenaar:
In addition, it can take a while for the tree of changes to be constructed. This could cause "app not responding" messages.

Alhough this is only a problem for really big change sets.

PR #1967, opened 2022-12-28
Publish to PyPI after all other jobs have passed

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

If part of a release build breaks, it could be that the artifacts have been uploaded to Pypi.org already.

### What is the new behavior?

Only publish to PyPI after all jobs succeed.

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -43,6 +43,9 @@ jobs:
     runs-on: ubuntu-22.04
     timeout-minutes: 30
     if: "!contains(github.event.head_commit.message, 'skip ci')"
+    outputs:
+      targz: gaphor-${{ steps.setup_and_test.outputs.version }}.tar.gz
+      wheel: gaphor-${{ steps.setup_and_test.outputs.version }}-py3-none-any.whl
     steps:
       - uses: actions/checkout@v3.2.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/full-build.yml
```diff
@@ -380,3 +392,21 @@ jobs:
         if: always()
         run: type gaphor-self-test.txt
         shell: cmd
+
+  publish-to-pypi:
+    name: Publish to PyPI (release only)
+    needs: [ linux-gtk4, linux-flatpak-devel, macos-gtk4, check-linux-appimage, check-macos-app, check-windows-installer ]
```
danyeaw, line 398: Since the 3 check-* jobs are the last builds to run in the pipeline, I don't think you need the other jobs here

.github/workflows/full-build.yml
```diff
@@ -39,10 +41,14 @@ jobs:
...
 
   linux-gtk4:
+    name: Linux (GTK4)
     needs: lint
     runs-on: ubuntu-22.04
     timeout-minutes: 30
     if: "!contains(github.event.head_commit.message, 'skip ci')"
+    outputs:
+      targz: gaphor-${{ steps.setup_and_test.outputs.version }}.tar.gz
+      wheel: gaphor-${{ steps.setup_and_test.outputs.version }}-py3-none-any.whl
```
danyeaw, line 51: Do these need to have `dist/` prefixed to the name?

danyeaw (commented):
Thanks for updating this! Maybe we can create a pre-release version once this is merged to fully test out the pipeline.

.github/workflows/full-build.yml
amolenaar: No, since they're uploaded as artifacts without `dist/`. In the Publish to PyPI job it's downloading them in the `dist/` folder.

.github/workflows/full-build.yml
amolenaar: The check-* jobs do not depend on the other jobs. I figured it's best if they also pass before uploading to PyPI, just t be on the safe side. Strictly speaking we only need the check-* jobs.

danyeaw (approved):

PR #1971, opened 2022-12-29
Fix PEP597 Encoding Warnings

danyeaw:
PEP 597 adds encoding warnings and will eventually make UTF-8 the default encoding in Python. This PR fixes encoding warnings to help prevent bugs introduced due to encoding warnings. For example, our `check-babel` module wouldn't run at all in Windows due to an encoding error.

I don't think should be a breaking change, we already try to set encoding to None if opening a model fails due to an encoding error.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

In many places we open and write to files without the encoding specified

Issue Number: N/A

### What is the new behavior?
Encodings are explicitly set to UTF-8

- [ ] Yes
- [X] No

### DIFF
--- a/.github/actions/setup_and_test/action.yml
+++ b/.github/actions/setup_and_test/action.yml
@@ -30,5 +30,7 @@ runs:
       run: poetry run poe gettext-mo
       shell: bash
     - name: Test with Pytest
+      env:
+        PYTHONWARNDEFAULTENCODING: 'true'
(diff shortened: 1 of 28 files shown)


amolenaar:
Wow, that's quite a few places that need encoding explicitly defined.

Great job fixing this!

PR #1991, opened 2023-01-06
Drop pytest-mock

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Pytest-mock was only used in a handful of tests. Everywhere else we use Monkeypatch.

### What is the new behavior?

The few tests that did use pytest-mock have been updated to use monkeypatch instead.
pytest-mock has been dropped as a dependency.

- [ ] Yes
- [x] No


danyeaw (approved):
Nice idea!

Issue #1954, opened 2022-12-25
Ability to create a package


sz332:
### Is your feature request related to a problem?
Add ability to create a package without creating any kind of diagram

### Describe the solution you'd like
I would like to be able to add packages without creating diagrams. 

### Describe alternatives you've considered
-

### Additional context
-

amolenaar:
This is possible by right-clicking on any package. Only it's not possible to create a top level package directly. You'll have to drag it.

danyeaw:
Maybe we should allow it from the top level project?

sz332:
As a user that would be the most intuitive action for me. I honestly never thought about dragging the package to whatever.

danyeaw:
@sz332 I just tested this with 2.14.0, if I right-click on the top level package, I am able to create a Package directly.
![image](https://user-images.githubusercontent.com/10014976/209578046-89f0652e-e933-48ee-8b64-9a7c30792009.png)

danyeaw:
I get it after reading #1952, you are deleting the top level model package, and now you don't have a way to create a new one. Other tools actual call this top level package a Model, put a triangle in the package symbol, and I don't think it can be deleted. I don't think that is specified in the OMG specs, but what if we did something similar?

amolenaar:
I think it's perfectly fine to have multiple top-level packages.

One option is to create a popup menu when not over any elements. 
Another is to generalize the "New diagram" button in the header bar, as discussed in #1927, and allow for a `Create toplevel package` entry.

sz332:
From user point of view the best way to handle this would be definitely to have a top level/root element with a custom icon showing that this represents the model which simply cannot be deleted. The advantage of this concept is that it opens the door to have multi-model solution later, when in a single project you can import multiple models (some of them as read only). It is extremely useful when working on smaller project but when you created reusable UML Profiles (for example for conceptual modeling, REST, etc.).

amolenaar:
I added a popup menu for the empty space of the model browser. It fixes the problem of not being able to create top level packages without having to put them in a diagram. I'm considering adding a "New Package" in the diagrams popup in the header (which then should change it's icon). Not sure if we really need it, though.

Quite some work has to be done before we can support multiple models in one window. Especially if some of them need to be read-only. I think it would be more natural to introduce this concept _when_ we can actually support multiple models.

danyeaw:
@sz332 Thanks again for submitting this issue! Are you happy with us closing this with #1995 merged?

[...]

PR #1995, opened 2023-01-06
Add general popup menu for model browser

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

When you cleaned the whole model, there's no way to easily create new top-level packages without putting them in a diagram as well.

Issue Number: #1954

### What is the new behavior?

When no element is selected, you can add top level diagrams and packages.

This allows to add top-level packages and diagrams from anywhere within the model browser (tree view).

You can now deselect the selected element in the model browser with <kbd>Ctrl</kbd>+mouse click.

image

- [ ] Yes
- [x] No

### Other information

I considered adding it to the "create diagram" button in the header bar as well.

* Renamed "tree component" on layout level as `model_browser`, as it's called in the docs.
* Fixed issue where only new popover menu's were created and none were cleaned up
* Remove GTK4 related code from Namespace component.
* Break dependency from tree component/namespace on function in main window module.


danyeaw (commented):
I like being able to right-click and add a new Package. The refactoring you did also is really nice, I like how the GTK4 and GTK3 code is split apart more for this component.

I noticed when I right-click and go to New Diagram, only the Class Diagram is displayed. Was this on purpose? Do we need diagrams that don't exist within any namespace / package?

amolenaar:
> I noticed when I right-click and go to New Diagram, only the Class Diagram is displayed. Was this on purpose? Do we need diagrams that don't exist within any namespace / package?

Somehow the menu is not resizing to contain all diagram types.

amolenaar:
Not sure why the popup menu is not resizing.

Anyway, it may make sense to add all entries in the main menu anyhow:

image

danyeaw:
Thanks @amolenaar, nice improvement!

Issue #1955, opened 2022-12-25
Don't show tab bar if there's only one tab


bertob:
We generally don't show tab bars when there's a single one (except in Builder, but that's for reasons which don't really apply to other apps, i.e. that it has an in-app tiling system). The current status quo is not great, because there's a useless second bar at all times, which just says "main" when you create a new model:

![image](https://user-images.githubusercontent.com/1908896/209475062-8bf73ca8-14ea-4318-8cdc-d8c03f587872.png)

As far as I understood, the bar was enabled to make it clearer which diagram you're currently working on, so we'd need to find an alternative place to show the title. Would the window title in the headerbar be an option here, e.g. instead of "Gaphor" in the example above?

Independently from that we should find a better default name than "main". Would something like  "Diagram 1" or "Main Diagram" work?

Followup from #1926

danyeaw:
Thanks @bertob for the suggestion!

I am assuming if there were two tabs open, and you close one, that the tab bar would collapse up, and the diagram would take up the space that the tab bar was using. Is that a correct understanding of how it would work?

amolenaar:
Yes. That's actually how it worked for 2.12. However, then the diagram name is no longer visible, so you can't know which diagram you're editing.

The header bae shows the name of the model (file name without suffix). If it's a new model it'll show `Gaphor`.

![image](https://user-images.githubusercontent.com/96249/209583510-95380055-0f1f-4a4b-bff4-fcf9bddcd85c.png)

We could add the diagram name (in this case `sequence-diagram – main`).

I changed the diagram name to _New diagram_ in #1956.

bertob:
> The header bae shows the name of the model (file name without suffix). If it's a new model it'll show Gaphor.

If the file name is visible below anyway, could we put the diagram name there instead? Alternatively I guess we could try "model name • diagram name" in the title?

amolenaar:
I'll give it a try. Alternatively we can also show the qualified name (packages + diagram name) in the top bar. We do not have space for that in the tabs, normally.

amolenaar:
Like this?

![image](https://user-images.githubusercontent.com/96249/211784118-47bd3062-2923-43f0-b545-f82ffaaca364.png)

Or show the full path?

![image](https://user-images.githubusercontent.com/96249/211784263-7d604035-4ba6-482a-9e43-7fb729912875.png)

If the path contains more than 3 elements, I suggest it's shown as **Model :: &hellip; :: Package :: Diagram name**.
[...]

PR #2004, opened 2023-01-11
Show diagram name in header

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

We always show the tab bar. It's better, according to GNOME style guidelines, to hide the tab bar if there's only one tab shown.

However, we want to tell the user which diagram they're editing.

Issue Number: #1955 

### What is the new behavior?

The diagram name is shown in the header bar. The tab bar is hidden if only one diagram is opened.

We can also change this to showing the full path.

Updated window name:
image

## Other

I did some cleanup on some of the events, and reduced the state maintained by `MainWindow`.

### DIFF
--- a/gaphor/event.py
+++ b/gaphor/event.py
@@ -68,9 +68,10 @@ class SessionShutdown(ServiceEvent):
 
 
 class ModelLoaded:
-    def __init__(self, service, filename: Path | None = None):
+    def __init__(self, service, filename: Path | None = None, modified: bool = False):
         self.service = service
         self.filename = filename
+        self.modified = modified
 
 
 class ModelSaved:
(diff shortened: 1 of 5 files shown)


danyeaw (commented):
Hi @amolenaar, I really like this - showing the diagram name in the title looks great along with autohiding the tabs.

I did notice when I rename a diagram that the Title doesn't change unless I select another diagram and go back to the original one.

amolenaar:
Ouch. I thought I tested it :see_no_evil:

amolenaar:
I updated the Diagrams component to send an extra `CurrentDiagramChanged` event, so the main window can update the header accordingly.

Issue #2015, opened 2023-01-18
Application crashes when attempting to save diagram


aunsbjerg:
### Describe the bug
Application crashes when trying to save using `ctrl+s`.

### To Reproduce
1. open program and open a diagram
2. press ctrl+s

This happens regardless of whether using `GDK_BACKEND=x11` or not.

### Expected behavior
I would expect a save dialog

### Screenshots

Log from running the application:

```bash
╰─ GDK_BACKEND=x11 ./Gaphor-2.15.0-x86_64.AppImage                           ─╯
gaphor.storage.storage INFO Read 58 elements from file

(gaphor-exe:20707): GtkSourceView-WARNING **: 10:47:41.343: Failed to load '/tmp/.mount_GaphorrmXmlY/gaphor/ui/language-specs/gaphorcss.lang': could not find the RelaxNG schema file

(gaphor-exe:20707): GLib-GIO-ERROR **: 10:48:34.861: Settings schema 'org.gtk.gtk4.Settings.FileChooser' is not installed
[1]    20707 trace trap (core dumped)  GDK_BACKEND=x11 ./Gaphor-2.15.0-x86_64.AppImage
```

And error report from the tools menu:

```
Gaphor version:    2.15.0
Operating System:  Linux (5.14.0-1055-oem)
Python version:    3.11.1
GTK version:       4.8.4
Adwaita version:   1.2.1
PyGObject version: 3.43.0
Pycairo version:   1.23.1
Cairo version:     1.17.7
Pango version:     1.50.13
Display:           GdkX11Display

No errors have been intercepted. We’re good.
```

### OS
- [x] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: Gaphor-2.15.0-x86_64.AppImage

### Additional information
Seems to be identical to #1355 which was apparently resolved in a later version.

amolenaar:
Hi @aunsbjerg,

Thank you for reporting this issue. What linux distribution are you running? Do you have GTK4 installed on your host machine?

aunsbjerg:
> Thank you for reporting this issue. What linux distribution are you running? 

I am using Ubuntu 20.04.5 LTS

> Do you have GTK4 installed on your host machine?

Not to my knowledge. Is GTK4 a pre-requisite for gaphor?

amolenaar:
It should not have to be. I've created a PR. I think I nailed the issue, since I could make it fail in CI with a similar error.

[...]

PR #2016, opened 2023-01-18
Fix resource lookup for packaged Gaphor

amolenaar:
### PR Checklist

### PR Type

- [x] Bug fix
- [ ] Feature

AppImage crashes when GTK4 is not installed on the system.

This is because it's looking for GSettings schema's only in system folders (e.g. `/usr/share`). Instead it should look in our package (`sys._MEIPASS/share`).

Issue Number: #2015 

### What is the new behavior?

Resources are looked up in the right data directory.

### DIFF
--- a/_packaging/use_gtk_4.py
+++ b/_packaging/use_gtk_4.py
@@ -1,3 +1,5 @@
 import os
+import sys
 
 os.environ["GAPHOR_USE_GTK"] = "4"
+os.environ["XDG_DATA_DIRS"] = os.path.join(sys._MEIPASS, "share")  # type: ignore[attr-defined]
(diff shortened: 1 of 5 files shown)


_packaging/use_gtk_4.py
```diff
@@ -1,3 +1,5 @@
 import os
+import sys
 
 os.environ["GAPHOR_USE_GTK"] = "4"
+os.environ["XDG_DATA_DIRS"] = os.path.join(sys._MEIPASS, "share")  # type: ignore[attr-defined]
```
amolenaar, line 5: @danyeaw Should we upstream this setting to PyInstaller?

danyeaw (approved):
Nice fix!

Issue #2018, opened 2023-01-20
Background is white in dark mode


CorvetteCole:
### Describe the bug
On Linux, version 2.15 installed through Flathub. The application properly detects dark mode and the entire UI is in dark mode except for the background

### To Reproduce
1. Install from Flathub on Linx
2. Open and create new model
3. Enable dark mode
4. Observe white background

Gaphor version:    2.15.0
Operating System:  Linux (6.1.6-arch1-3)
Python version:    3.10.6
GTK version:       4.8.3
Adwaita version:   1.2.1
PyGObject version: 3.42.2
Pycairo version:   1.23.0
Cairo version:     1.16.0
Pango version:     1.50.12
Display:           GdkWaylandDisplay

No errors have been intercepted. We’re good.

### Expected behavior
The background to be black, as shown in screenshots online

### Screenshots
![image](https://user-images.githubusercontent.com/19213965/213592332-a0455ee8-6372-40ee-bf27-d61c50ac5cd6.png)

### OS
-  Arch Linux

### Version
Version of Gaphor: 2.15

danyeaw:
Hi @CorvetteCole, instructions on how to get a dark background with white text are here in the docs:
https://gaphor.readthedocs.io/en/latest/style_sheets.html

CorvetteCole:
Interesting @danyeaw , you're totally right I missed that. This is still pretty darn counter-intuitive though, especially since if you make a new graph it will be light again, and since the rest of the UI switches automatically. Is there a case for some sort of change here?

CorvetteCole:
(I understand why that would be very difficult given the CSS styling built-in already)

amolenaar:
We would basically need a different style sheet for dark mode. The style sheet is also used when you export the diagram as an image. So it seemed reasonable to keep the drawing area off-white for now.

An alternative would be a `@media` query in the CSS :wink:.

danyeaw:
What if we introduce a default color palette depending on light/dark mode?

Then the CSS only overrides the default. For example, in dark mode, the background of the diagram is always black, unless you override the CSS attribute for the background of the diagram.

When exporting images, it could default to the light pallette, and maybe could be configurable to export dark as well.

CorvetteCole:
I like that idea personally, it feels more intuitive from a new user perspective

amolenaar:
[...]

PR #2027, opened 2023-01-22
Add support for CSS variables

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

If we want to use the same color/line width/... in multiple places, we have to define those explicitly.

Issue Number: #2018

### What is the new behavior?

It is now possible to define variables:

```css

* {
  --bg-color: rgb(12, 12,12);
  --line-width: 2.4;
}

diagram {
  background-color: var(--bg-color);
}

association {
  line-width: var(--line-width);
}
```

See also https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties

Support for default/fallback values has not been implemented. Given our context is quite limited (e.i. we're not making a CSS framework), we can probably do without fallback values.

If variables do not resolve a value from the parent layer is used. E.g. if you provide an invalid variable for color, the default color is used.

### DIFF
--- a/docs/style_sheets.md
+++ b/docs/style_sheets.md
@@ -212,6 +212,33 @@ be provided to set the level of line wobbliness. 0.5 is default, 0.0 is a
 straight line. The value should be between -2.0 and 2.0. Values between 0.0 and
 0.5 make for a subtle effect.
 
+### Variables
+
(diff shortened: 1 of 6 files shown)


danyeaw (approved):
Nice one @amolenaar!

PR #1973, opened 2022-12-30
Fix Translations on Windows, macOS, and AppImage

danyeaw:
This PR fixes two major issues with translations:

1. PyInstaller wasn't collecting any .mo translation files, so they weren't available in any of the packaged apps built using PyInstaller.
2. On Windows, Gaphor wasn't automatically setting the language based on the locale Windows was running with, so translations weren't working without manually setting the LANG environmental variable

### PR Checklist
Please check if your PR fulfills the following requirements:

- [X] I have read, and I am following the [Contributor guide](https://github.com/gaphor/gaphor/blob/main/CONTRIBUTING.md)
- [X] I have read, and I understand the [Code of Conduct](https://github.com/gaphor/gaphor/blob/main/CODE_OF_CONDUCT.md)

### PR Type
What kind of change does this PR introduce?

- [X] Bug fix
- [ ] Feature
- [ ] Chore (refactoring, formatting, local variables, other cleanup)
- [ ] Documentation content changes

### What is the current behavior?
App translations aren't working with packaged apps on Windows, macOS, and Linux with AppImage

Issue Number: #1969 

### What is the new behavior?
Translations are working on all platforms

### Does this PR introduce a breaking change?
- [ ] Yes
- [X] No

PR #2045, opened 2023-01-28
Win and macOS: Wrong language selected when region not default

danyeaw:
@mathiascode pointed out (thanks!) in https://github.com/gaphor/gaphor/pull/1973 that `locale.getlocale()` returns the wrong language if the region in Windows or macOS is set different from the display language.

This PR uses the native language defaults in Windows and macOS to get the correct language.

I feel like this should just be built in to Python, maybe we should attempt to create a PR to CPython itself?

### PR Checklist

### PR Type

- [X] Bug fix
- [ ] Feature

The wrong language is selected under certain circumstances.

Issue Number: N/A

### What is the new behavior?
The correct language should always be selected.

- [ ] Yes
- [X] No


gaphor/i18n.py
```diff
@@ -15,18 +17,47 @@
...
+            return language
+    elif sys.platform == "darwin":
+        import subprocess
+
+        try:
+            lang_out = subprocess.check_output(
+                ("defaults", "read", "-g", "AppleLanguages")
+            )
+            if languages := lang_out.decode("utf-8").strip('()\n" ').split(","):
+                return languages[0][:2]
```
mathiascode, line 42: This output works with the LANGUAGE environment variable, but I don't remember if LANG supports it. Might be worth confirming this.

amolenaar:
What settings for region (=country?) and language should be set differently to make locale.getlocale() produce the wrong locale?

gaphor/i18n.py
danyeaw: I think LANGUAGE allows for a list of languages, but I expect here we are just getting the first one. Maybe @amolenaar would be able to help test on macOS that this works.

mathiascode:
> I feel like this should just be built in to Python, maybe we should attempt to create a PR to CPython itself?

I'm not sure if there are other cross-platform Python projects that handle this better, but it would be nice to know how others do translations. The macOS hack in particular is very ugly, but at the time I didn't find another way without bringing in third-party dependencies.

danyeaw:
> What settings for region (=country?) and language should be set differently to make locale.getlocale() produce the wrong locale?

In Windows, if I set the language to Spanish and the region to Ingles with any country, Gaphor starts in English.

gaphor/i18n.py
amolenaar: Maybe https://stackoverflow.com/questions/17679651/how-to-figure-out-what-the-current-system-preferred-language-is? It isn't spawning any processes at least.

danyeaw: Nice one, I'll update the PR!

amolenaar:
I can confirm that on macOS, I always get the message on the `main` branch

```
No translations were found: [Errno 2] No translation file found for domain: 'gaphor'
```

With the changes on this PR it's taking my preferred language, which I set to Dutch for this test. I set my region (country) to Spain. The language I see is Dutch, as expected.

I added a commit, so only the Cocoa framework code is installed.

amolenaar:
I think it's good to merge.

danyeaw:
Thanks for the teamwork on this one!

PR #2050, opened 2023-01-30
Replace flake8, pyupgrade, docformatter, and isort with ruff

danyeaw:
[Ruff](https://github.com/charliermarsh/ruff) is a fast linter that is able to replace a few of the tools we have been using, fix more automatically, and uses the pyproject.toml for configuration.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Linting uses flake8, isort, pyupgrade, docformatter

Issue Number: N/A

### What is the new behavior?
Linting uses ruff

- [ ] Yes
- [X] No


amolenaar:
Nice. This reduces the pre-commit checks on my machine from 15 to 7 seconds.

gaphor/storage/storage.py
```diff
@@ -139,7 +139,9 @@ def load_elements_generator(elements, factory, modeling_language, gaphor_version
     # The elements are iterated three times:
     size = len(elements) * 3
 
-    def update_status_queue(_n=[0]):
+    def update_status_queue(_n=None):
+        if _n is None:
+            _n = [0]
```
amolenaar, line 144: This change causes the progress bar to stop increasing at around 50%. Not sure why, yet.

I agree that the list should be removed as argument.

To see it in action: open gaphor, from the greeter load `UML.gaphor` model. You'll see the progressbar stop  around the middle. On `main` it works fine.

amolenaar (changes requested):
Nice simplification. There's one small regression AFAICS, when loading a model.

Issue #1920, opened 2022-12-14
AppImage crashes on Fedora 37 / Wayland


amolenaar:
### Describe the bug

Starting the latest AppImage on Fedora 37 causes a crash

### To Reproduce
1. Start AppImage from a terminal
2. See it crash:

```
 $ ./Gaphor-2.13.0.dev1937+6761c206-x86_64.AppImage 

(process:596852): GLib-GIO-ERROR **: 13:51:39.573: Settings schema 'org.gnome.settings-daemon.plugins.xsettings' does not contain a key named 'antialiasing'
[1]    596852 trace trap (core dumped)  ./Gaphor-2.13.0.dev1937+6761c206-x86_64.AppImage
```

### OS
- [x] Linux / Wayland
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.13.0

### Additional information

On X11 / XWayland it works okay. As a workaround the application can be launched with:

```sh
GDK_BACKEND=x11 ./Gaphor-*.AppImage
```

Not sure if we want to invest time in fixing this. It's a never ending story to get AppImage working on all platforms.

I added a comment on our [Download page](https://gaphor.org/download/).

Not sure if it should be raised as an issue with GTK.

danyeaw:
It works great for me in Wayland on latest OpenSUSE, we'll have to figure out what's unique about Fedora 37.

amolenaar:
Something DBus-y IIRC.

danyeaw:
I just ran Fedora 37 from the Desktop LiveCD, downloaded the AppImage and it launched just fine.

danyeaw:
I am now able to reproduce this on the latest openSUSE with the version 2.14.2 AppImage

amolenaar:
Should be fixed by #2020.

PR #2077, opened 2023-02-08
Package GSettings daemon schemas for AppImage

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

The AppImage crashes when run on Wayland

Issue Number: #1920

### What is the new behavior?

GSettings daemon schemas are also included. Now our Wayland environment also starts.

- [ ] Yes
- [x] No

### Other information

I'm not sure how I can build and upload a new images to ghrc.io. The docker image used to build the AppImage should be updated as well.


danyeaw:
@amolenaar I think this is good to go now.

Pip was throwing some weird setuptools not found error when not using a virtualenv with the Docker container (even with setuptools installed), so I switched the CI to use pipx instead to keep things isolated when installing poetry.

danyeaw (approved):

amolenaar:
Looks good. It's a bit awkward that poetry needs to be installed via pipx, though.

danyeaw:
I submitted an enhancement request to pipx since pipx is already installed by default on the GitHub Actions runners, and we shouldn't have to reinstall it for Python 3.11:
https://github.com/pypa/pipx/issues/940

Issue #803, opened 2021-05-14
Diagram metadata


vanillajonathan:
I would like to set metadata.

- A short description of what the diagram represents.
- A URL to a website.
- Author name and email address.
- Revision, version, or last updated at.
- Created on date.
- Updated on date.
- License, maybe in the form of a SPDX short identifier such as [CC-PDDC](https://spdx.org/licenses/CC-PDDC.html).

danyeaw:
UML/SysML species that you can attach a Comment element to a diagram itself.

![image](https://user-images.githubusercontent.com/10014976/118344637-38f15e00-b4fd-11eb-9e4e-de58c7c15784.png)

Gaphor currently doesn't support the diagram frame, or attaching comments to the diagram itself. If we could attach a comment to a diagram would this provide what you are looking for? Then you can define any information you want to show on each diagram.

vanillajonathan:
> UML/SysML species that you can attach a Comment element to a diagram itself.

I wasn't are of that.

> If we could attach a comment to a diagram would this provide what you are looking for?

I don't know. Maybe. I just noticed there were no way to set any metadata and thought there ought to be a way to so. I had in mind a dialog with predefined fields but maybe this could work too.

amolenaar:
@vanillajonathan What do you want do with the metadata?

vanillajonathan:
Nothing in particular in this point of time, I just figured there ought to be some sort of metadata support.

So if someone comes across my file, this person can know what version it is, find a URL where he can find the latest version if there is any newer version, and if I made any mistakes it should be possible for someone to contact me over email to tell me.

amolenaar:
@vanillajonathan I see. I would expect this file to be in a (git) repository or something, so you should be able to obtain all meta data from there as well.

vanillajonathan:
If it gets printed on a paper then it might be useful to have the metadata printed on the paper.

amolenaar:
We have an really nice Sphinx extension in Gaphor 2.7.

vanillajonathan:
Sphinx is not what I want, because the image may be printed or it may be embedded on a website or redistributed. I want a metadata block in the corner of the image.

https://www.google.com/search?q=engineer+drawing+title+block&tbm=isch
https://en.wikipedia.org/wiki/Engineering_drawing#Title_block

amolenaar:
[...]

PR #2082, opened 2023-02-11
Diagram metadata

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: #803

### What is the new behavior?

Add a diagram item that can provide diagram meta data.

- [ ] Should this diagram item be backed by a model element, or is this data really diagram specific?
- [x] Add Property page to edit fields in metadata block
- [x] Hide unused fields?


danyeaw:
How do I add metadata to a diagram?

danyeaw:
How do I add metadata to a diagram?

danyeaw:
How do I add metadata to a diagram?

amolenaar:
I forgot: it still needs a proper icon.

amolenaar:
@danyeaw @vanillajonathan Added an icon.

image

NB. All fields are saved in the diagram item. There's no model element with "shared" data.

gaphor/diagram/general/metadata.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class MetadataItem(ElementPresentation):
+    createdBy: attribute[str] = attribute("createdBy", str, "")
+    description: attribute[str] = attribute("description", str, "")
+    website: attribute[str] = attribute("website", str, "")
+    revision: attribute[str] = attribute("revision", str, "")
+    license: attribute[str] = attribute("license", str, "")
+    createdAt: attribute[str] = attribute("createdAt", str, "")
+    updatedAt: attribute[str] = attribute("updatedAt", str, "")
```
danyeaw, line 28: I would expect that Created/Updated At would be for a location, I suggest Created On and Updated On would be more clear for a date.

gaphor/diagram/general/metadata.py
```diff
@@ -0,0 +1,206 @@
...
+
+        self.shape = Box(
+            *(
+                [
+                    Box(
+                        *(
+                            [
+                                Box(
+                                    Text(
+                                        text=gettext("created by:"),
```
danyeaw, line 66: Kind of nit, but I think the metadata would look better if it was in Title Case with each work capitalized.

danyeaw (approved):
This looks fantastic, I really like how it turned out! It is simple to add metadata to a diagram, and I think it was the right approach to save it to the diagram instead of creating a separate element.

I have two minor comments about naming.

gaphor/diagram/general/metadata.py
vanillajonathan: Yeah, or all caps but in a small font like it was in my proposal.
https://github.com/gaphor/gaphor/issues/803#issuecomment-986680585

gaphor/diagram/general/metadata.py
amolenaar: fixed

gaphor/diagram/general/metadata.py
amolenaar: Let's stick with Title Case indeed. It's more consistent with the rest of the app.

amolenaar:
> Looks good, maybe the Gtk.Entry for the license could have proposals for some SPDX license identifiers such as `CC-PDDC`. `CC0-1.0`, `CC-BY-NC-4.0`, `CC-BY-NC-SA-4.0`, `CC-BY-SA-4.0`, `GFDL-1.3-or-later`.

I'd like to postpone this until we're rid of the GTK3 code. GTK4 does not have a nice editable dropdown by itself (`GtkComboBoxText` is deprecated).

vanillajonathan:
> > Looks good, maybe the Gtk.Entry for the license could have proposals for some SPDX license identifiers such as `CC-PDDC`. `CC0-1.0`, `CC-BY-NC-4.0`, `CC-BY-NC-SA-4.0`, `CC-BY-SA-4.0`, `GFDL-1.3-or-later`.
> 
> I'd like to postpone this until we get rid of the GTK3 code. GTK4 does not have a nice editable dropdown by itself (`GtkComboBoxText` is deprecated).

GTK 3 have [`EntryCompletion`](https://docs.gtk.org/gtk3/class.EntryCompletion.html) for the `Gtk.Entry` widget. GTK 4 also have [`EntryCompletion`](https://docs.gtk.org/gtk3/class.EntryCompletion.html)but it is deprecated since 4.1.0.

https://discourse.gnome.org/t/replacement-for-entrycompletion/13505

PR #2091, opened 2023-02-15
Properly unsubscribe when property page is removed

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Event handlers "leak" in GTK 4.

Issue Number: N/A

### What is the new behavior?

Add a function to unsubscribe watchers. It appeared handlers were not properly removed in GTK4, since in GTK4 widgets are not "destroyed", but only use reference counting.

### Information

Added a function to `gaphor.diagram.propertypages` to unsubscribe automatically on widget `destroy` (gtk3) or `unparent` (gtk4).

### DIFF
--- a/gaphor/C4Model/propertypages.glade
+++ b/gaphor/C4Model/propertypages.glade
@@ -21,7 +21,6 @@ renderers and such.</property>
     <property name="orientation">vertical</property>
     <property name="spacing">6</property>
     <property name="baseline-position">top</property>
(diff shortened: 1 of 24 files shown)


danyeaw (approved):
This looks great! How did you figure out that GTK4 event handlers were leaking?

amolenaar:
I know something changed in the way widgets are destroyed. Initially I just made it work (not break), but forgot to come back and fix it for real.

PR #2112, opened 2023-03-01
Proper loading of current/incoming model

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Currently, Gaphor is notifying a user in case a model contains a merge conflict. However, apart from the merge conflicts, in bigger models there are also changes already in the combined model that did not cause an error.

This approach does not work, since we have a lot of objects serialized in the model and a lot of object references. So it's very likely that, with a bigger model, parts are merged and parts are not. Hence we can not just base the current and incoming model on the partly merged state.

Issue Number: #388

### What is the new behavior?

Instead of reading the partly merged model, Gaphor should read the original current and incoming model from Git. That way we have a greater assurance that the model is actually valid.

This has a downside that we need to know about the VCS, in our case Git.

- [ ] Yes
- [x] No

### Other information

Added [pygit2](https://www.pygit2.org/) as a dependency to deal with the Git models.

The code has changed to only show the merge conflict dialog in case we can perform the merge.


danyeaw (approved):
This looks good!

PR #2143, opened 2023-03-10
macOS: Bundle the application with GTK4

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

With GTK 4.10 enabled we can change the macOS build to use GTK4 by default.

Issue Number: #1839 

### What is the new behavior?

Build macOS app with GTK 4.

In addition, I did a bit of cleanup: removed `linux|macos_dependencies` actions and renamed the linux gtk4 build to `linux-wheel`.

Now, always a DMG will be built. The .zip distribution caused problems ("application is damaged"). Also, this is more in line with the final distribution.


_packaging/gaphor.spec
```diff
@@ -87,14 +83,14 @@ a = Analysis(  # type: ignore
...
-                "GtkSource": "4" if os.getenv("GAPHOR_PKG_GTK") == "3" else "5",
+                "Gtk": "4.0",
+                "GtkSource": "5",
             },
         },
     },
     hookspath=["."],
     runtime_hooks=[
-        f"use_gtk_{os.getenv('GAPHOR_PKG_GTK', '4')}.py",
+        f"use_gtk_4.py",
```
danyeaw, line 93: I think we can remove this whole runtime hook since `GAPHOR_USE_GTK=4` is the default.

danyeaw (approved):
This look great, one additional opportunity to remove an extra runtime hook I think, or you can include it with #2146.

amolenaar:
I removed that in the follow-up PR where we get rid of the 'GAPHOR_USE_GTK' environment variable.

Issue #2138, opened 2023-03-08
NotInTransactionException when adding swimlanes


danyeaw:
### Describe the bug
I am consistently getting a NotInTransactionException when adding a couple of additional swimlanes.

### To Reproduce
1. Open the SysML starting model
2. Logical Level -> Behavior -> Functional Boundary Behavior
3. Click on the Swimlane, go to the Property Page and increase the number of swimlanes from 2 -> 4

```
gaphor.plugins.errorreports.errorreports ERROR Uncaught exception
  + Exception Group Traceback (most recent call last):
  |   File "/home/dan/Projects/gaphor/.venv/lib/python3.11/site-packages/gaphas/decorators.py", line 86, in async_wrapper
  |     func(*args, **kwargs)
  |   File "/home/dan/Projects/gaphor/.venv/lib/python3.11/site-packages/gaphas/view/gtkview.py", line 313, in update
  |     model.update_now(dirty_items)
  |   File "/home/dan/Projects/gaphor/.venv/lib/python3.11/site-packages/gaphas/decorators.py", line 126, in wrapper
  |     return func(*args, **kwargs)
  |            ^^^^^^^^^^^^^^^^^^^^^
  |   File "/home/dan/Projects/gaphor/gaphor/core/modeling/diagram.py", line 394, in update_now
  |     self._connections.solve()
  |   File "/home/dan/Projects/gaphor/.venv/lib/python3.11/site-packages/gaphas/connections.py", line 74, in solve
  |     self._solver.solve()
  |   File "/home/dan/Projects/gaphor/.venv/lib/python3.11/site-packages/gaphas/solver/solver.py", line 185, in solve
  |     c.solve()
  |   File "/home/dan/Projects/gaphor/.venv/lib/python3.11/site-packages/gaphas/solver/constraint.py", line 105, in solve
  |     self.solve_for(wvar)
  |   File "/home/dan/Projects/gaphor/.venv/lib/python3.11/site-packages/gaphas/constraint.py", line 211, in solve_for
  |     self.bigger.value = self.smaller.value + self.delta
  |     ^^^^^^^^^^^^^^^^^
  |   File "/home/dan/Projects/gaphor/.venv/lib/python3.11/site-packages/gaphas/solver/variable.py", line 111, in set_value
  |     self.notify(oldval)
  |   File "/home/dan/Projects/gaphor/.venv/lib/python3.11/site-packages/gaphas/solver/variable.py", line 90, in notify
  |     handler(self, old)
[...]

PR #2164, opened 2023-03-20
Update diagram directly when partitions change

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Exceptions are thrown when partitions with actions when more swimlanes are added.

Issue Number: #2138 

### What is the new behavior?

Changing partitions may cause actions to be moved around as well, since they should remain in the rigth swimlane.
Now the diagram items are updated as part of changing the swimlanes.

- [ ] Yes
- [x] No

### Other information

Also fixed the `Diagram.update_now()` interface: matrix-only updates are no longer used.

### DIFF
--- a/gaphor/UML/actions/partitionpage.py
+++ b/gaphor/UML/actions/partitionpage.py
@@ -111,3 +111,5 @@ class PartitionPropertyPage(PropertyPageBase):
 
             last_child = self.partitions.get_last_child()
             self.partitions.remove(last_child)
+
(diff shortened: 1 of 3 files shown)


danyeaw (approved):

danyeaw:
Thanks @amolenaar, this fix is a big improvement to the activity partitions! I did notice if I make the swim lanes very narrow that activities will jump out of them and get lost in the diagram, but we can make that as part of a future fix.

PR #2171, opened 2023-03-23
Make main window always available

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

LibAdwaita is complaining sometimes a dialog is opened without a parent window reference.

Issue Number: N/A

### What is the new behavior?

The main window is always available, so it can be used as a parent for any kind of popup.

- [ ] Yes
- [x] No


danyeaw (approved):
Nice one, I had seen some of those warnings as well :+1:

Issue #2162, opened 2023-03-19
Export Language overwrite


cloud-erik:
### Is your feature request related to a problem?
At the moment laguage for GUI and export is derived from system laguage. Even if the model itself is language independet, it is only possible to export drawings in system laguage.
So I could not use gaphor for projects or publications with language =! system laguage.

### Describe the solution you'd like
For publication or interaction in international projects, it would be great to set (at least export-) laguage independet from system language.

### Describe alternatives you've considered

- Change of system language to project language. But this is not convinient if you would like to work in system in local laguage or if system language is set by administrator.
Also switching back and forward could happen if working for projects with different languages.

- Changing source code of e.g. https://github.com/gaphor/gaphor/blob/a16d18c33609c4b1c7544da5daa1758bef53de7e/gaphor/i18n.py local and overwrite language.

### Additional context
When working in a international environment, it would be helpfull, if its possible to set laguage of exported drawings different from loacal laguage.
E.g., you work in a project with aggreed project laguage is English but team is globaly distributed, it would be nice if language (at least for export) could be set independend from system language.

Other used case would be, if translation in local laguage is not suficient.

See also #1176

amolenaar:
Hi @cloud-erik, thank you for your report.

I do not think I understand. The system language is only used for the GUI. The models/diagrams do not have any internationalization support. You can use whatever language you agree upon, independent of the system language.

cloud-erik:
Hi @amolenaar thanks for pointing me to that point. Maybe I am just misunderstading the model or structure.
Let me try to explaine a little more on following exmaple.
If I have a PC with local laguage German and I create e.g. a requirement it is created with type "Anforderung", which is the germand translation of requirement. I found no way to translate or rename that when I export to pdf, png or eps.
Same also for other elements.
So If would like to create a English drawing, only way is to change system language (if I have it installed and I am allowed to).
Please see attached example where marked parts which I could not translate so far.
![Requirement](https://user-images.githubusercontent.com/67910530/226382488-9427d952-b449-442f-8212-4ca46c48d903.jpg)

And I would be happy, if it would be possible to set language (for everything or just export) independent from system. In my example to English.

amolenaar:
Ah! I totally forgot about those. Indeed, some default stereotypes and keywords are translated. My bad.

It sounds very reasonable to change the translation for those terms. Maybe introduce a "model language", so the terms match the language used in the model?

amolenaar:
[...]

PR #2178, opened 2023-03-25
Change the spoken language in a model

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

The diagram language s the same as the system language. This can be a problem if the model is supposed to be made in a particular language.

Issue Number: #2162 

### What is the new behavior?

Add a setting for the language used in the model. This can be different from the system language. The language setting is saved along with the model.

The translated text is taken from `Diagram.gettext()`.

- [ ] Yes
- [x] No

### Other information

Question: should we set the default language to English, as @peter88213 suggests in #2162, or should it stick to the system default language? For me personally, having the diagrams in English would be my default.

image

### DIFF
--- a/gaphor/ui/elementeditor.py
+++ b/gaphor/ui/elementeditor.py
@@ -302,7 +302,7 @@ class PreferencesStack:
             (None, "English"),
             *sorted(
                 (
-                    (d.name, Locale.parse(d.name).get_language_name())
+                    (d.name, Locale.parse(d.name).get_language_name().title())
                     for d in localedir.iterdir()
                     if d.is_dir()
                 ),
--- a/gaphor/ui/elementeditor.py
+++ b/gaphor/ui/elementeditor.py
@@ -16,6 +16,7 @@ from gaphor.core.modeling.event import (
     ModelReady,
 )
 from gaphor.diagram.propertypages import PropertyPages, new_resource_builder
+from gaphor.i18n import gettext
 from gaphor.ui.abc import UIComponent
 from gaphor.ui.csscompletion import (
     CssFunctionCompletionProvider,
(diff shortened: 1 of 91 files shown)


gaphor/ui/elementeditor.py
```diff
@@ -34,6 +35,32 @@
...
+    ("it", gettext("Italian")),
+    ("ja", gettext("Japanese")),
+    ("ko", gettext("Korean")),
+    ("pl", gettext("Polish")),
+    ("pt_BR", gettext("Portuguese (Brazil)")),
+    ("ru", gettext("Russian")),
+    ("sl", gettext("Slovenian")),
+    ("es", gettext("Spanish")),
+    ("sv", gettext("Swedish")),
+    ("tr", gettext("Turkish")),
```
amolenaar, line 60: We could also get those language names from the Babel database. That would safe us some translation work.

Only, how do we know what the actual language is?

danyeaw: You mean what the system language is set to?

I like the idea to get the names from Babel. Could we parse the name of each folder in gaphor/locale to get the list of supported languages? Then `babel.Locale.get_language_name()` should be able to translate the name, I think.

amolenaar:
The merge conflict resolver already proved useful in this PR :)

amolenaar:
I think the natural/spoken language is working fine. I changed the app, so it's using English by default. I think it makes sense, since English is *the* language of Open Source software and most software used internationally is written in English. The language can be changed and the language preferences will be saved as part of the model.

danyeaw (approved):
This is a great feature @amolenaar!

gaphor/ui/elementeditor.py
```diff
@@ -295,7 +298,20 @@ def tx_update_style_sheet(style_sheet, text):
         self._style_sheet_update = DelayedFunction(800, tx_update_style_sheet)
         self._in_update = 0
 
+        self._languages = [
+            (None, "English"),
+            *sorted(
+                (
+                    (d.name, Locale.parse(d.name).get_language_name())
```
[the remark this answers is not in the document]
amolenaar: Done. Looks better indeed.

Issue #2214, opened 2023-04-07
Add General Support for Code Import to generate a single UML map of a codebase.


duaneking:
### Is your feature request related to a problem?
I'm trying to map a very large code base; I cannot find a single tool that is open source and free that allows me to simply scan the directory support the model as UML.  I'm finding a lot of proprietary software that claims to be open source that is completely neutered and ineffective and unable to do its job without a third party plugin or component that you have to buy for an excessive amount of money.

### Describe the solution you'd like
I would like to be able to point the tool at a directory containing source code in C#/Java and export a UML.

### Describe alternatives you've considered
I have literally tried everything I can think of. The prior open source solutions have all been taken off of the market by their authors and made closed source/expensive. In addition, many non-functional abandoned solutions seem to litter the landscape and make it that much harder to find a solution.

### Additional context
Adding code import to UML support for languages in general would be amazing and actively put this project on the map as a project worth investing time and energy in.  I cant easily add languages without the initial support that doesn't currently exist.

danyeaw:
Hey @duaneking, sorry you are having trouble finding an open source tool to your liking, and thanks for the suggestion. This doesn't sound like our focus right now for this project. This is also a similar request for importing Python code at https://github.com/gaphor/gaphor/issues/1363, but it sounds like you are looking for something more general for a C# or Java project.

Are you interested in contributing support for this?

duaneking:
I responded in the other thread.

danyeaw:
Thanks again @duaneking for the feature request. We have updated the README to make it more clear that this isn't one of our goals for the project.

PR #2232, opened 2023-04-14
Add non-goals to README

danyeaw:
We have received a couple of requests for things that I don't think should be the focus of Gaphor in https://github.com/gaphor/gaphor/issues/2214 and https://github.com/gaphor/gaphor/issues/1363.

Sometimes it helps clarify the focus of things when we are also explicit about the things that aren't the focus of the project.

I know at one point, there was a lot of interest in MDA and creating executable models that could generate source code. I think the power of modeling is for the descriptive and analytical analysis that is possible. If you want an actual implementation, then just write the code. Do others agree?

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature
- [X] Documentation content changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,13 +23,14 @@ It is designed to be easy to use, while still being powerful. Gaphor implements 
 
 ## 📑 Table of Contents
 
-- [Background](#-background)
-- [Install](#-install)
-- [Usage](#-usage)
-- [Contributing](#-contributing)
-    - [Translations](#-translations)
-    - [Code of Conduct](#%EF%B8%8F-code-of-conduct)
-- [License](#%EF%B8%8F-license)
+- [📑 Table of Contents](#-table-of-contents)
+- [📜 Background](#-background)
+- [💾 Install](#-install)
+- [🔦 Usage](#-usage)
+- [♥ Contributing](#-contributing)
+  - [🌍 Translations](#-translations)
+  - [♿️ Code of Conduct](#️-code-of-conduct)
+- [©️ License](#️-license)
 
 ## 📜 Background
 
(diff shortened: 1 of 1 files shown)


amolenaar:
> I think the power of modeling is for the descriptive and analytical analysis that is possible.

I completely agree.

Nice addition! I'm wondering if it should also include (web-) service like OSLC. These can be done in a different project too. (I'm happy to host it in the gaphor org, though).

PS. I have a small app (done at a client) that generates [workflows from models](https://github.com/amolenaar/cr2.zaakbegeleider). It created the configuration for the actual application.

amolenaar (approved):

duaneking:
You mentioned my request so I wanted to respond.

I have no interest in code generation from UML myself; my current focus is on mapping legacy software that already exists in order to better understand it and maintain it. In that goes,  I'm able to find a lot of Projects like this that want you to start a new project using their solution, but none that want you to be able to maintain a project that you may have or may not have started with their UML solution.

Effectively it really looks to me as an outsider, like you're trying to limit yourself to a small problem that's already been solved multiple times by many other open source projects, because the open source landscape is littered with projects for UML that only allow you to use their UI to generate an image but do not allow you to import or scan code that is already written; Making this project yet another one of them makes me struggle to understand the value of this project if all you want to do is effectively copy the functionality of other open source project that doesn't move the needle, so respectfully, I'm deeply confused by this.

If this project is not not interested in helping developers write maintainable code or deep dive legacy code to update it and support upgrade paths by making the code more discoverable by presenting an open source tool that leverages python to be cross platform and allow developers to create UML models,  then you should understand that from my perspective, that's exactly what I would expect from this sort of project. It's the perfect fit for this project.

The ability to manually go in with a UI and create a UML diagram is by itself possibly useful for small projects for the sighted, but only if you have a V0 idea that you have not already committed to, or you do not have a legacy code base that somebody else has already generated based on their idea that you now have to map.  And you're completely missing the ability to convert and deal with the resulting UML in a way that allows you to make it a bindle.

The core value of this tool is not in starting new projects; It's in reviving old ones, and allowing the community to reuse old code and investment without having to start from scratch every time.

danyeaw:
Hi @duaneking, thanks for the feedback! I absolutely understand your use case, but that's not our goals for the project. If you want to scratch your itch, we would be glad to provide guidance while you build a plugin for Gaphor to do what you are looking for.

PR #2299, opened 2023-05-11
Fix undo loops

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

If code in an event handler raises an error, this can cause an rollback loop.

### What is the new behavior?

Check if we're already performing a rollback from the undo manager. If so, do not try to roll back again.

- [ ] Yes
- [x] No


danyeaw (approved):
Hey @amolenaar, this looks great!

amolenaar:
Thanks.

Issue #389, opened 2020-08-07
Support plugins 


amolenaar:
### I'm submitting a...
- [ ] Bug report  
- [X] Feature request
- [ ] Documentation issue or request

### Current behavior

It's not really possible to install plugins for the Flatpak, Windows and Mac builds.

This is a shame, since Gaphor is built to ver *very* extensible.

### Expected behavior

It would be nice if we can provide plugins (e.g. requirements reporting, ONTO-UML, etc.) for the pre-packaged builds, so users can easily extend Gaphor with new functionality.

NOTES:

* How to deal with plugins that provide "Modeling Languages"?
* How to test compatibility with Gaphor versions for the plugins?
* Should we host a repository somewhere, that contains all plugins? (like FlatHub does)

TiemenSch:
The package still contains python, right? Would it then be possible for users to "drag-and-drop" (or git clone) packages into a predetermined folder? E.g. a folder that Gaphor's built-in Python scouts for plugins to include.

Another option would be recipes for offering plugins on the different distribution methods. Although this may grow unwieldy rather quickly with many manuals to maintain. Given how tricky packaging is, I personally wouldn't recommend this.

With respect to plugin-definition on the Python side of things, I can recommend taking a look at [pluggy](https://pluggy.readthedocs.io/en/latest/) which lies at the core of pytest. Although I think the current services system shows some overlap with this.

With respect to compilation, you may get [Nuitka](https://nuitka.net/pages/overview.html) to compile "just" the plugin/module you want to add such that a Gaphor install (also compiled using Nuitka?) might pick it up. Nuitka is a Python compiler written in Python.

amolenaar:
My idea is to use Pypi.org as a distribution mechanism. With an index available on Github maybe (it would be cool if we can query PyPI directly with a gaphor-plugin specific filter). We can install in a predefined folder, like you say. E.g. `~/.local/lib/gaphor2` (or `gaphor2.2`, but something with a version, so we can introduce breaking changes at some point. The advantage of using pip is that it also takes transitive dependencies into account.


I looked at Pluggy before, but its extension mechanism is different from how Gaphor works. In Gaphor, entry points are considered services. A service can take arguments: other services. 
This is how extensions can make use of existing functionality. 
In pluggy the possible arguments for a plugin are pre-defined.

amolenaar:
Whichever system is used for plugins, it would be good to sandbox them in some way, so the core functionality of Gaphor stays safe.

[...]

PR #2301, opened 2023-05-12
Execute scripts from the command line

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Currently command line parsing is done in the UI module. This limits the extensibility (and usability of Gaphor in non-UI environments).

Issue Number: #389

### What is the new behavior?

Command line arguments are parsed in `gaphor.main`. The arguments are nicely grouped.

It was also pretty easy to add script execution (I recall it being discussed somewhere). It also opens the door for plugin support: we can provide options to install plugins and list plugins.

```
usage: gaphor [-h] [-v] [-d] [-q] [--exec script] [-p] [--self-test] [--gapplication-service] [model ...]

Gaphor is the simple modeling tool.

options:
  -h, --help            show this help message and exit
  -v, --version         print version and exit
  -d, --debug           enable debug logging
  -q, --quiet           only show warning and error logging

scripting options:
  --exec script         execute a script file and exit

interactive (GUI) options:
  -p, --profiler        run in profiler (cProfile)
  --self-test           run self test and exit
  model                 model file(s) to load

GApplication settings (use from D-Bus service files):
  --gapplication-service
```
- [x] Yes
- [ ] No

### DIFF
--- a/gaphor/__main__.py
+++ b/gaphor/__main__.py
@@ -1,6 +1,7 @@
 import sys
 
-from gaphor.ui import main
+from gaphor.main import main
+
 
 if __name__ == "__main__":
     sys.exit(main(sys.argv))
--- a/gaphor/main.py
+++ b/gaphor/main.py
@@ -7,7 +7,7 @@ from gaphor.application import distribution
 LOG_FORMAT = "%(name)s %(levelname)s %(message)s"
 
 
-def main(argv) -> int:
+def main(argv=sys.argv) -> int:
     """Start Gaphor from the command line."""
     args = parse_args(argv)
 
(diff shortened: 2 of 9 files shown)


danyeaw (approved):
I like this idea to better separate the CLI from the UI code, nice one!

If using Gaphor from the command line becomes much more common, it might be nice to update the interface with Typer or Click, but I think not having those extra dependencies for now is good.

amolenaar:
I considered it (shortly). It may be an option when we get plugins that also want to expose CLI commands.

Initially I wanted something like sub-parsers (as is used in poetry and pip) for executing scripts:

```
gaphor exec script.py
```

However, then we should do something like `gaphor gui` to launch the gui, and that will not work well. The gui should be the default.

PR #2302, opened 2023-05-13
Diagram export as gaphor CLI subcommand

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

in #2301 I added an option to execute a script from the command line. However, this approach does not scale.
What I wanted is to say something like `gaphor exec script.py`. This is different from an option, since now we have more control over what arguments are supported.

Issue Number: N/A

### What is the new behavior?

In the new behavior you can execute `gaphor exec script.py`. This can be extended by `gaphor export` for exporting diagrams, replacing `gaphorconvert`.

```
$ gaphor --help
usage: gaphor [-h] [-V] {exec,export,gui} ...

Gaphor is the simple modeling tool.

options:
  -h, --help         show this help message and exit
  -V, --version      print version and exit

subcommands (default: gui):
  Get help for commands with gaphor COMMAND --help.

  {exec,export,gui}
    exec             execute a python script from within gaphor
    export           export diagrams from a gaphor model
    gui              launch the gui

Thank you for using Gaphor <https://gaphor.org>.
```
```
$ gaphor export --help
usage: gaphor export [-v | -q] [--profiler] [-h] [-u] [-o directory] [-f format]
                     [-r regex]
                     model [model ...]

Export diagrams from a Gaphor model.

positional arguments:
  model

options:
  -v, --verbose         enable verbose logging
  -q, --quiet           only show warning and error logging
  --profiler            run in profiler (cProfile)
  -h, --help            show this help message and exit
  -u, --use-underscores
                        use underscores instead of spaces for output filenames
  -o directory, --dir directory
                        output to directory
  -f format, --format format
                        output file format, default pdf
  -r regex, --regex regex
                        process diagrams which name matches given regular expression;
                        name includes package name; regular expressions are case
                        insensitive
```

There is a bit of smartness involved: a subcommand should be provided as first argument. Default options (`-v/-q/-p/-V`) are supported for all (sub)commands. If no known subcommand is provided, the default (gui) parser is used.

A new entry point has been introduced: `gaphor.argparsers`. Here we can define parsers for commands we want to support via the CLI.

### Other information

I tried to make it work with Click first, but it did not work with default options (including multiple model arguments) and subcommands. Therefore I went back to argparse.

### DIFF
--- a/_packaging/gaphor.spec
+++ b/_packaging/gaphor.spec
@@ -69,6 +69,7 @@ a = Analysis(  # type: ignore
     + copy_metadata("gaphor")
     + copy_metadata("gaphas"),
     hiddenimports=collect_entry_points(
+        "gaphor.argparsers",
         "gaphor.services",
         "gaphor.appservices",
         "gaphor.modelinglanguages",
--- a/gaphor/entrypoint.py
+++ b/gaphor/entrypoint.py
@@ -28,7 +28,7 @@ def load_entry_points(scope, services=None) -> Dict[str, type]:
     for ep in list_entry_points(scope):
         cls = ep.load()
         if not services or ep.name in services:
-            logger.debug(f'found service entry point "{ep.name}"')
+            logger.debug(f'found entry point "{scope}.{ep.name}"')
             uninitialized_services[ep.name] = cls
     return uninitialized_services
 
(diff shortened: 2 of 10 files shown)


danyeaw (approved):
Nice addition, and wow, argparse wizardry :wink:

amolenaar:
The composition feature ('parents') is quite powerful. 🧙‍♂️🪄

PR #2303, opened 2023-05-14
Ensure Linux CI uses latest Python

danyeaw:
I noticed that the Linux Wheel build was using Python 3.10, even though Python 3.11 was installed by the setup-python action. This PR forces the Linux build to use of the right version when installing poetry, which then sets the rest of the steps up to use the right version instead.

### PR Checklist

### PR Type

- [X] Bug fix
- [ ] Feature

Linux build is using an older Python Version

Issue Number: N/A

### What is the new behavior?
Python 3.11.3 is being used

- [ ] Yes
- [X] No


amolenaar:
The python version used in Flatpak is still 3.10, so I think we need to test it with 3.10 at least.

danyeaw:
@amolenaar Ya, that's a good point, at least until org.freedesktop.Sdk 23.08 is released. Should we setup a matrix build in Linux with both versions of Python? The current method was by accident, we were telling it to use Python 3.11 using setup-python, and we were getting 3.10 instead with just Linux.

amolenaar:
Hmm. Then it may be better to set up a matrix build, since we need to support both Python version on Linux at least.

danyeaw:
@amolenaar I think this is ready to go now!

amolenaar:
Nice! We can rely on our CI even more now.

It's a pitty the build script has become a little more complex, but that's inevitable. It will be pretty easy to test with Python 3.12 this way.

Issue #2304, opened 2023-05-15
Double-clicking on a diagram element bug


Linuxiness:
### Describe the bug
Double-clicking on a diagram element does not show in the Diagram section.

### OS
- [x] Linux (Manjaro, Appimage)

### Version
Version of Gaphor: 2.18.0

### Additional information
```
Gaphor version:    2.18.0
Operating System:  Linux (5.10.174-1-MANJARO)
Python version:    3.10.9
GTK version:       4.8.3
Adwaita version:   1.2.2
PyGObject version: 3.42.2
Pycairo version:   1.23.0
Cairo version:     1.17.8
Pango version:     1.50.12
Display:           X11Display


Errors:

Time since application startup: 9:48:05
|Traceback (most recent call last):
|  File "/home/igor/.local/lib/python3.10/site-packages/gaphor/ui/modelbrowser.py", line 84, in list_view_activate
|    self.open_element(element)
|    │                 └ <gaphor.UML.uml.Message element 7c46ea76-f2ae-11ed-8705-10f0059ea926>
|    └ <gaphor.ui.modelbrowser.ModelBrowser object at 0x7f39bb1074c0>
|  File "/home/igor/.local/lib/python3.10/site-packages/gaphor/ui/modelbrowser.py", line 139, in open_element
|    self.event_manager.handle(ElementOpened(element))
|    │                         │             └ <gaphor.UML.uml.Message element 7c46ea76-f2ae-11ed-8705-10f0059ea926>
|    │                         └ <class 'gaphor.ui.event.ElementOpened'>
|    └ <gaphor.ui.modelbrowser.ModelBrowser object at 0x7f39bb1074c0>
|  File "/home/igor/.local/lib/python3.10/site-packages/gaphor/core/eventmanager.py", line 83, in handle
|    self._events.handle(queue.pop())
|    │                   └ deque([])
|    └ <gaphor.core.eventmanager.EventManager object at 0x7f39d2befdc0>
|  File "/home/igor/.local/lib/python3.10/site-packages/generic/event.py", line 61, in handle
|    raise ExceptionGroup("Error while handling events", exceptions)
|          │                                             └ [AssertionError()]
|          └ <class 'exceptiongroup.ExceptionGroup'>
|exceptiongroup.ExceptionGroup: Error while handling events (1 sub-exception)
└─┬──────────────────────────────╌┄┈
  |Traceback (most recent call last):
[...]

PR #2309, opened 2023-05-17
Fix drop for messages with one end connected

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Messages with one end connected can't be dropped on a diagram

Issue Number: #2304

### What is the new behavior?

Dropping of (interaction) messages with only one end connected works.


danyeaw (approved):
Nice fix @amolenaar!

PR #2313, opened 2023-05-18
Allow grouping of property with current owner element

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

I was not able to let the properties of "Espresso Machine" (in the coffee machine example) be grouped in an espresso machine. Problem seemed to be that there was an association related to those properties.

Issue Number: N/A

### What is the new behavior?

Return `True` from the `group()` dispatch function when the property is already owned by the requested parent. This allows the properties to be grouped on their existing owner.
 
- [ ] Yes
- [x] No


danyeaw (approved):
Thanks @amolenaar!

Issue #2325, opened 2023-05-26
"Show type" option of the "Decision" node breaks project


Fell-x27:
### Describe the bug
If you check the "show type" option and then save the project, you can't open the file. There will be an error about "invalid model".

### To Reproduce
1. Create a new empty model;
2. Add a "merge/decision node", hosted on "G" hotkey by default;
3. Click on it;
4. Open attributes panel;
5. Check "show type" switcher;
6. Save the model;
7. Try to reopen it...and see the error;

The internal report has no errors, only the common information:
```
Gaphor version:    2.18.0
Operating System:  Linux (5.19.0-40-generic)
Python version:    3.10.6
GTK version:       4.10.1
Adwaita version:   1.3.1
PyGObject version: 3.44.0
Pycairo version:   1.23.0
Cairo version:     1.16.0
Pango version:     1.50.14
Display:           X11Display
```

### Expected behavior
The file should open without errors.

### File
Here is an example of the simplest broken project:
[broken.txt](https://github.com/gaphor/gaphor/files/11573715/broken.txt)

### OS
- [x] Linux ( Linux Mint 21.1 Cinnamon)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.18.0

### Additional information
You can **_fix it by deleting <show_type> sections_** from your file;

danyeaw:
Thanks @Fell-x27 for the nice clear bug report, I am able to reproduce this :+1:

danyeaw:
@all-contributors please add @Fell-x27 for bug

PR #2327, opened 2023-05-28
Fix decision node loading

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Models can not load if "show-type" is selected on decision nodes.

Issue Number: #2325 

### What is the new behavior?

* Fix loading of decision nodes with "show-type" enabled
* Add extra test to validate properties on all diagram items.

- [ ] Yes
- [x] No


danyeaw (approved):
This looks really good, nice fix!

This is kind of a nit, but underlaying isn't a word, but underlying is. Would this be an opportunity to rename this to show_underlying_type?

amolenaar:
@danyeaw Thanks for correcting this. For sure this is *the* opportunity to fix it once and for all.

Fell-x27:
Additionally, I suggest that if the utility encounters an unknown node or subnode in the XML, it should just ignore it, write a relevant message to the log, and continue operating, instead of crashing.

amolenaar:
> Additionally, I suggest that if the utility encounters an unknown node or subnode in the XML, it should just ignore it, write a relevant message to the log, and continue operating, instead of crashing.

Doing so has the potential side effect of loading a completely invalid model. However, in this case it would've helped a lot as well.

It can be done, but Gaphor should make it really clear to users that something is not okay.

PR #2329, opened 2023-05-28
Replace deprecated Gtk.TreeView with ListView: Activity Parameter Nodes

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

`Gtk.TreeView` is deprecated. GTK 4 has a much nicer tree widget: `Gtk.ListVIew`. It's already used for the model browser. We should also use it in the property editor, for both consistency and ease of use.

- [x] Yes
- [ ] No

### Other information

`Gtk.ListView` has no header. This will be [introduced in GTK 4.12](https://docs.gtk.org/gtk4/property.ListView.header-factory.html). So we may want to postpone changing to `ListView`s.

### DIFF
--- a/gaphor/UML/actions/tests/test_activitypropertypage.py
+++ b/gaphor/UML/actions/tests/test_activitypropertypage.py
@@ -1,3 +1,4 @@
+import pytest
 from gi.repository import Gtk
 
 from gaphor import UML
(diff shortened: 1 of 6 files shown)


amolenaar:
This PR now has replaced the activity parameter property page with a new style list view.

I think it reached feature parity, with keybindings and easy-adding features in place.

Let's do the other TreeView's via different PR's, so we can keep the change manageable.

gaphor/UML/actions/tests/test_activitypropertypage.py
```diff
@@ -1,53 +1,185 @@
...
+    property_page.construct()
+
+    del activity_item.subject.node[node]
+
+    assert not activity_item.subject.node
+    assert property_page.model.get_n_items() == 1
+    assert property_page.model.get_item(0).node is None
+
+
+def test_update_model_with_multiple_noes_when_parameter_deleted(
```
danyeaw, line 107: Super nit, but should this be nodes instead of noes?

danyeaw (approved):
Really nice update to use ListView @amolenaar!

gaphor/UML/actions/tests/test_activitypropertypage.py
amolenaar: fixed

PR #2335, opened 2023-05-30
Be more lenient when loading models

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: N/A

### What is the new behavior?

Do not crash if a property or reference cannot be found.

NB. This was already the case for single references.

If an element type is not found, an exception is raised.

- [ ] Yes
- [x] No

### Other information

- [ ] Display a warning or notification when errors have occurred.

### DIFF
--- a/gaphor/storage/storage.py
+++ b/gaphor/storage/storage.py
@@ -216,7 +216,10 @@ def _load_attributes_and_references(elements, update_status_queue):
 
         # load attributes and references:
         for name, value in list(elem.values.items()):
-            elem.element.load(name, value)
+            try:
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
This looks great, I think this will really reduce occurrences of models failing to load!

danyeaw:
@amolenaar I see in the other info section there is a checkbox, anything else needed for this before we merge?

amolenaar:
Not for now. I haven't figured out yet how to communicate this to the user. It's in the logging anyway.

PR #2347, opened 2023-06-07
UI: Mark translatable strings yes

sabriunal:
It appeares that using "yes" is a common practice for translatable strings in GNOME apps.

### PR Checklist

### PR Type

- [x] No


amolenaar:
Hi @sabriunal, Thank you for your addition. Indeed "yes" is the term we should use for translatable strings.

@all-contributors Add @sabriunal for code.

PR #2353, opened 2023-06-12
Add a simple pip-based plugin installer

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Gaphor is only extensible if installed manually via pip (e.i. set up a dev environment).

Issue Number: N/A

### What is the new behavior?

Allow to install extra packages from pypi and make them available when running a packaged version of Gaphor.

- [x] Yes
- [ ] No

### Other information

- [x] Add plugin path before any entrypoints are loaded
- [ ] Add documentation on how to use the plugin commands
- [x] Add a self-check (ensure pip is bundled)
- [ ] Once packaged, the `sys.executable` is no longer `python`, but `gaphor-exe`. I suppose pip should be run from inside our Gaphor instance, instead of spawning a subprocess, although that's discouraged.

https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -75,7 +75,7 @@ jobs:
         env:
           CC_TEST_REPORTER_ID: 05f6288b94a87daa172d3e96a33ec331a4374be7d01eb9a42b3b21c4c550a8ff
         with:
-          coverageCommand: poetry run coverage xml
+          coverageCommand: poetry run coverage xml --omit="/tmp/*"
       - name: Create Source Dist and Wheel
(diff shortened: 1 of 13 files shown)


amolenaar:
I get a bit stuck on this one: `pip` is a command line tool at it's core, and it's hard to make it work embedded in a PyInstaller distribution. Mostly because it's spawning subprocesses.

Maybe we should limit our install efforts to use [resolvelib](https://pypi.org/project/resolvelib/), and [installer](https://pypi.org/project/installer/).

BTW. The flatpak and mac entitlements should probably have network access in order to install packages from PyPI.

PR #2371, opened 2023-06-19
Do no longer package TkInter

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

I noticed Tk(Inter) is bundled with our app. That makes no sense.

### What is the new behavior?

Nothing changed. TkInter has been excluded.

See https://github.com/pyinstaller/pyinstaller/wiki/Recipe-remove-tkinter-tcl

Also, there's no need for Gaphas dist-info, so that's also removed.

- [ ] Yes
- [x] No


danyeaw (approved):
Great idea!

PR #2376, opened 2023-06-21
Load plugins installed in plugin folder

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

When Gaphor is packaged as Window exe or in a macOS app, there's no way to add extra functionality to the app.

Issue Number: #389

### What is the new behavior?

Gaphor can load and initialize services defined in a separate plugins folder (`$HOME/.local/gaphor/plugins-2` by default). This location can be overridden by an environment variable `GAPHOR_PLUGIN_PATH`.

### Other information

This PR contains the working parts -- loading of plugins -- from #2353.

Plugin installation and maintenance is a lot harder can cannot be done from within Gaphor at the moment. This should be fixed in a future PR.


danyeaw (approved):
This looks great @amolenaar!

Issue #2366, opened 2023-06-17
Apply undo to swimlane delete recreates swimlanes in reverse order


mikekidner:
### Describe the bug
When you apply Undo to the deletion of swimlanes, the lanes return in reverse order. The enclosed actions do not relocate.

### To Reproduce
1. Create three swimlanes in a diagram, name uniquely. Add action to each lane 
2. Select and delete the swimlanes
3. apply ctrl-Z / command-Z to undo the delate
4. See the order of lanes is reversed. And actions have not stayed in lanes.

### Expected behavior
Upon undo, the swimlane object should return as it was prior to delete.

### Screenshots
<img width="516" alt="Screen Shot 2023-06-17 at 10 40 55 am" src="https://github.com/gaphor/gaphor/assets/50702106/7e6ab36d-4a84-4f94-a920-0f2b6bf329dc">
Before deletion.

<img width="509" alt="Screen Shot 2023-06-17 at 10 41 06 am" src="https://github.com/gaphor/gaphor/assets/50702106/81e01e58-d5f1-4072-8283-d7e003549d4b">
After undo delete.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [x] macOS
- [ ] Windows

### Version
Version of Gaphor:2.18.1

danyeaw:
Hi @mikekidner, thanks for submitting the issue! I can reproduce the names of the swimlanes getting reordered.

amolenaar:
The "problem" here is that we treat collections as sets. Delete events should record the position, so we can add them to the same place when undoing.

danyeaw:
@all-contributors please add @mikekidner for bug

PR #2388, opened 2023-06-26
Fix order when undoing deletion from association

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: #2366

### What is the new behavior?

The index of an element deleted from an association is recorded, so it can be placed
back on the same index when it was deleted.

- [ ] Yes
- [x] No


danyeaw (approved):
Nice fix @amolenaar!

Issue #2368, opened 2023-06-17
Ports on parts do not group below owning block


mikekidner:
### Describe the bug
As [discussed](https://github.com/gaphor/gaphor/discussions/2339) ports do not group beneath the owning block, if created on the part. Also, ports created on a block are not available to use on the part.

### To Reproduce
1. Create a blocks B1 , B2, B3, with ports on B2 and B3. See how they are grouped beneath the blocks in the model.

<img width="843" alt="Screen Shot 2023-06-18 at 8 10 32 am" src="https://github.com/gaphor/gaphor/assets/50702106/149e15a5-5fab-4bc6-a84a-ff03709ee125">

2. Create an IBD for B1

<img width="536" alt="Screen Shot 2023-06-18 at 8 12 08 am" src="https://github.com/gaphor/gaphor/assets/50702106/00c04cf9-0707-4f25-a0ea-16bc217fd821">

See how the ports are not available to use on the diagram

3. Add a port to a part in the IBD. See how it appears at the top level of the model, not under the owning block

<img width="601" alt="Screen Shot 2023-06-18 at 8 12 46 am" src="https://github.com/gaphor/gaphor/assets/50702106/8d8af152-5724-4329-8590-34338d36532a">

### Expected behavior
When adding ports to a block or it's corresponding part, it is expected that the port nests beneath the block. 
If a port exists on a block, it should exist on the corresponding part.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [x] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.18.1

amolenaar:
Thanks @mikekidner, I can reproduce this issue.

amolenaar:
The missing piece is here: https://github.com/gaphor/gaphor/blob/main/gaphor/SysML/blocks/connectors.py#L39-L40.

Is a `Property` actually allowed to have `Port`s? [`Port` is a subtype of `Property`](https://docs.gaphor.org/en/latest/models/uml/05._Structured_Classifiers.html#encapsulated-classifiers).

If so, what would the relation be? For a `ProxyPort` on a `Property`? `Property.datatype`?

mikekidner:
I think that placing a port on a part should be _interpreted_ as adding a port property to the owning block, 
So when dropping a port onto a part in a diagram, it would have to reference the owning block. 

To do nested ports, a port would have to have _port properties_.  So ports, may need to be realised as properties that can own properties, as you suggest.

[...]

PR #2394, opened 2023-06-27
Fix ports on properties

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Ports can be attached to properies but end up at toplevel in the model browser.

Issue Number: #2368 

### What is the new behavior?

* Properties can be assigned a proper type, like we do for `ItemFlow` and `Partition`.
* Ports can be attached to properties with a type assigned.

- [ ] Yes
- [x] No


danyeaw:
I checked the functionality, I like this! I'll try to finish looking at the code tomorrow.

danyeaw (approved):
Nice one @amolenaar!

PR #2399, opened 2023-06-28
Add a fullscreen mode to Gaphor

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Gaphor can be maximized, but not put in full screen mode.

### What is the new behavior?

Full screen mode.

Also updated the Shortcuts window and added a shortcut to show the Shortcuts window.

Updated translations too, since the terminology is not _that_ much different.


danyeaw (approved):
This is a nice addition!

Issue #2421, opened 2023-07-06
After using merge tool model becomes invalid


mikekidner:
### Describe the bug
Using merge tool (2.19.0) to resolve merge conflicts between edits made on 2.19.0 and 2.19.1. Merge was completed, but after closing file it no longer opened succesfully. On either version.

Running from the debugger showed:

```
miyomoto:gaphor mikekidner$ poetry run gaphor
No translations were found for language en_AU.UTF-8: [Errno 2] No translation file found for domain: 'gaphor'
gaphor.storage.storage INFO Read 4181 elements from file
> /Users/mikekidner/BitBucketRepos/gaphor/gaphor/core/modeling/properties.py(639)postload()
-> len(u) <= 1
(Pdb) l
634  	    def postload(self, obj):
635  	        self.version += 1
636  	        if self.upper == 1:
637  	            u = self.filter(obj)
638  	            assert (
639  ->	                len(u) <= 1
640  	            ), f"Derived union {self.name} of item {obj.id} should have length 1 {tuple(u)}"
641  	
642  	    def save(self, obj, save_func):
643  	        pass
644  	
(Pdb) p u
[<gaphor.SysML.sysml.Block element bdb9c7e6-0bca-11ee-8816-ba28606445aa>, <gaphor.UML.uml.Package element a24bd3b4-0b02-11ee-8936-ba28606445aa>]
(Pdb) p obj
<gaphor.SysML.sysml.Block element 19fad756-0b0e-11ee-8936-ba28606445aa>
(Pdb) p self.name
'namespace'
(Pdb) w
  /Users/mikekidner/BitBucketRepos/gaphor/gaphor/ui/filemanager.py(213)_load_async()
-> for percentage in storage.load_generator(
  /Users/mikekidner/BitBucketRepos/gaphor/gaphor/storage/storage.py(292)load_generator()
-> for percentage in load_elements_generator(
  /Users/mikekidner/BitBucketRepos/gaphor/gaphor/storage/storage.py(166)load_elements_generator()
-> elem.element.postload()
  /Users/mikekidner/BitBucketRepos/gaphor/gaphor/core/modeling/element.py(146)postload()
-> prop.postload(self)
> /Users/mikekidner/BitBucketRepos/gaphor/gaphor/core/modeling/properties.py(639)postload()
-> len(u) <= 1
(Pdb) 
```

### Expected behavior
Not make the file invalid for both versions. 
[...]

PR #2423, opened 2023-07-06
Lenient derived unions

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

`[0..1]` derived unions raise errors when there's multiple elements that.

Issue Number: #2421 

### What is the new behavior?

Just take the first element for a `[0..1]` derived union.

- [ ] Yes
- [x] No

### Other information

Cleaned up the collection code.

If we have an element with multiple owners, does it break the GUI? [Should be fine](https://github.com/gaphor/gaphor/blob/main/gaphor/ui/treemodel.py#L172-L176).


danyeaw (approved):
These are nice simplifications @amolenaar! 🤩

Issue #2433, opened 2023-07-07
Crash when adding parameter nodes to activity nodes


Bambofy:
### Describe the bug
When i try to add a parameter node to activities on an activity diagram, sometimes it causes the editor to crash. I have tried gaphor on both linux and OS X and it is happening on both OSes. I have captured the error logs for the first time the crash happened and is linked below.

[parameter-node-crash-report.txt](https://github.com/gaphor/gaphor/files/11985508/parameter-node-crash-report.txt)

Now after the editor has crashed, the editor seemingly crashes every time i try to add a action to an activity diagram. This functioned properly before and the previous crash due to the parameter node seems to have caused this too. I have added 2 error reports which occured after that first crash due to the parameter node.

[activity-node-crash-report-1.txt](https://github.com/gaphor/gaphor/files/11985505/activity-node-crash-report-1.txt)
[activity-node-crash-report-2.txt](https://github.com/gaphor/gaphor/files/11985507/activity-node-crash-report-2.txt)

### To Reproduce
1. Create a blank diagram.
2. Add an activity node to the diagram.
3. Add an action node to the diagram outside of the activity node
4. Add a parameter node to the activity node.

### Expected behavior
gaphor does not crash when adding parameter nodes.

### OS
- [X] Linux (Please put in notes the specific distro)
- [X] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.19.1

### Additional information
Linux distro: Ubuntu

amolenaar:
Hi @Bambofy, Thanks for reporting.

Indeed I can reproduce this locally with the Flatpak build (linux).

I suspect this error was introduced by https://github.com/gaphor/gaphor/pull/2329.

amolenaar:
I created a MR on GTK for this: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/6174.

amolenaar:
The issue is fixed in Gaphor 2.19.2.

However, I'd like to keep this issue open until a fix is proveded upstream.

PR #2436, opened 2023-07-08
Reapply #2329 - Gtk.ListView for Activity Parameters

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

**This PR should only be merged if we're sure it does not cause any crashes.**

The Activity Parameter property page had been reverted to `Gtk.TreeView`, since the `Gtk.EditableLabel` caused the app to crash.

Issue Number: #2433 

### What is the new behavior?

We want to get rid of deprecated API's.

`Gtk.EditableLabel` exposes some behavior that we currently have differently (e.g. the text should not change to editable state when a row is selected, only after a double click).

We have a new widget: `TextField`, that contains the text edit logic we use in the model browser. This widget looks more stable and works for the Parameter ListView too.

- [ ] Yes
- [x] No

### Other information

Added a placeholder text to the Parameter table, so it's more clear what a user can do in the box:

image

### DIFF
--- a/gaphor/UML/actions/activitypropertypage.py
+++ b/gaphor/UML/actions/activitypropertypage.py
@@ -1,62 +1,81 @@
-from gi.repository import Gtk
+from __future__ import annotations
+
+from gi.repository import Gdk, Gio, GObject, Gtk
 
 from gaphor import UML
-from gaphor.core import transactional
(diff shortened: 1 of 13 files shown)


danyeaw (approved):
This new Parameter adder looks great @amolenaar!

Issue #2452, opened 2023-07-10
Interactions: adding a comment line results in an error


TitouanReal:
### Describe the bug
Trying to add a comment line can fail on specific conditions.

### To Reproduce
1. Create a Generic empty model.
2. Add an interaction.
3. Add a lifeline to the interaction.
4. Add an execution specification to the lifeline.
5. Add a comment.
6. Try to add a comment line.
7. Trying to add the comment line immediately fails on the first click of the drag.

I can't add the Gaphor error report in this issue due to the character limit. Please let me know how I can provide this information if necessary.

### Expected behavior
I expect the comment line to draw.

### OS
- [x] Linux (Fedora 38)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.19.2

amolenaar:
I can trigger an issue. I hope it's the same that you see in the Error Reporter:


```
Time since application startup: 0:00:48
|Traceback (most recent call last):
|  File "/home/arjan/Development/gaphor/.venv/lib/python3.11/site-packages/gaphas/view/gtkview.py", line 506, in do_snapshot
|    self.painter.paint(list(items), cr)
|    │                       │       └ <cairo.Context object at 0x7f8de995a590>
|    │                       └ <generator object Diagram.sort.<locals>.<genexpr> at 0x7f8de9976f80>
|    └ <gtkview.GtkView object at 0x7f8dc1bedf40 (GaphasView at 0x55837025b7d0)>
|  File "/home/arjan/Development/gaphor/.venv/lib/python3.11/site-packages/gaphas/painter/chain.py", line 30, in paint
|    painter.paint(items, cairo)
|    │             │      └ <cairo.Context object at 0x7f8de995a590>
|    │             └ [<gaphor.UML.actions.activity.ActivityItem element 8c3b4530-1ff9-11ee-a057-0456e5e540ed>, <gaphor.UML.interactions.interaction.I...
|    └ <gaphor.diagram.painter.ItemPainter object at 0x7f8dc1bc3a50>
|  File "/home/arjan/Development/gaphor/gaphor/diagram/painter.py", line 53, in paint
|    self.paint_item(item, cr)
|    │               │     └ <cairo.Context object at 0x7f8de995a590>
|    │               └ <gaphor.UML.interactions.executionspecification.ExecutionSpecificationItem element 969c15ae-1ff9-11ee-a057-0456e5e540ed>
[...]

PR #2439, opened 2023-07-08
Fix Parameter Node and Execution Specification with Dark mode

amolenaar:
### PR Type
What kind of change does this PR introduce?

- [x] Bug fix
- [ ] Feature
- [ ] Chore (refactoring, formatting, local variables, other cleanup)
- [ ] Documentation content changes

### What is the current behavior?

Elements such as Parameters and Execution Specifications have a white background when used with dark mode.

Issue Number: #2437

### What is the new behavior?

The background color of overlaying elements changes with the color scheme.

Light mode:

image

Dark mode:

image

### Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #2460, opened 2023-07-11
Fix: resolve CSS style variables before using the values

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Rendering throws exception when an element with a solid background is on the diagram

In #2439 we introduced variables for the background color, so solid items could change the background color in dark mode.

Issue Number: #2452

### What is the new behavior?

CSS variables are evaluated before font-size and color (with opacity) is finally calculated.

- [ ] Yes
- [x] No


danyeaw (approved):
Looks good!

PR #2469, opened 2023-07-13
Fix loading of ProxyPorts with informationFlow attached

### LINKED ISSUE #2466: Can not load model: Proxyport.informationFlow
### Describe the bug

@2old4this reported an issue on Matrix. His saved models (Gaphor 2.13) could not be loaded.

This is caused by a `ProxyPort.informationFlow` attribute. This attribute was present in Gaphor 2.13, however
ProxyPort has no informationFlow attribute in Gaphor 1.19.

How did it get there? Was it a change in the model without proper migration step in the loader?

### Version
Version of Gaphor: 2.19.1

### Additional information

[Matrix discussion](https://matrix.to/#/!maFZnGxzqXqppVZrTd:gitter.im/$d-ojU0iqtd7vBMNexYRFPsCefNvEOElMWfiq6NAT1sQ?via=gitter.im&via=matrix.org&via=kompot.si)

I have been able to reproduce this issue:

1. Open Gaphor 2.13
2. Create 2 Blocks, connect a ProxyPort on both
3. Create a Connector between both ports
4. Enable Information Flow on the connector and define a name and a type
5. Save the model
6. Load the model with Gaphor 2.19

https://github.com/gaphor/gaphor/blob/main/gaphor/storage/storage.py#L428-L431 contains an upgrade function that deletes `informationFlow` from `Property`, however, it does not take into account `ProxyPort` (which is a subtype of `Property`).

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

The current upgrade function did only cover Property, no sub-types.

Fixes #2466.

### What is the new behavior?

ProxyPort is upgraded, similar to Property.

### DIFF
--- a/gaphor/storage/storage.py
+++ b/gaphor/storage/storage.py
@@ -426,6 +426,9 @@ def upgrade_flow_item_to_control_flow_item(elem, elements):
 
 # since 2.19.0
 def upgrade_delete_property_information_flow(elem):
-    if elem.type == "Property" and "informationFlow" in elem.references:
+    if (
+        elem.type in ("Property", "Port", "ProxyPort")
+        and "informationFlow" in elem.references
+    ):
         del elem.references["informationFlow"]
     return elem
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
Nice fix @amolenaar!

PR #2479, opened 2023-07-14
Toggle visibility of ProxyPort.type

### LINKED ISSUE #2424: Port type in label
### Describe the bug
Diagram label and tree label are inconsistent for typed ports. 

### To Reproduce
1. create block definition diagram
2. create port definition block
3. create component block with port
4. set port type to definition
5. Observe port correct

### Expected behavior
The port label should be similar to the internal block diagram parts (`Name:Type`) when a definition is provided. This is correct in the tree but not in the diagrams. Figure 9-7 in the [SysML 1.6 spec](https://sysml.org/.res/docs/specs/OMGSysML-v1.6-19-11-01.pdf) shows this for proxy and full ports 

### Screenshots
Current
image

Corrected Visualization
image

### OS
- [X] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.19.1

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Type of proxy ports is not shown

Issue Number: Fixes #2424

### What is the new behavior?

ProxyPort type is now shown if the "show-type" property is set.

### DIFF
--- a/gaphor/SysML/blocks/proxyport.py
+++ b/gaphor/SysML/blocks/proxyport.py
@@ -1,3 +1,4 @@
+from gaphor.core.modeling.properties import attribute
 from gaphor.diagram.presentation import AttachedPresentation, Named
 from gaphor.diagram.shapes import (
     Box,
@@ -25,7 +26,11 @@ def text_position(position):
 class ProxyPortItem(Named, AttachedPresentation[sysml.ProxyPort]):
     def __init__(self, diagram, id=None):
         super().__init__(diagram, id, width=16, height=16)
-        self.watch("subject[NamedElement].name")
+        self.watch("subject[NamedElement].name").watch(
+            "subject[TypedElement].type.name"
+        ).watch("show_type")
+
+    show_type: attribute[int] = attribute("show_type", int, default=False)
 
     def update_shapes(self):
         self.shape = IconBox(
@@ -35,6 +40,15 @@ class ProxyPortItem(Named, AttachedPresentation[sysml.ProxyPort]):
                     self.subject, [self.diagram.gettext("proxy")]
                 )
             ),
-            Text(text=lambda: self.subject and self.subject.name or ""),
+            Text(text=self._format_name),
             style=text_position(self.connected_side()),
         )
+
+    def _format_name(self):
+        if not self.subject:
+            return ""
+
+        name = self.subject.name or ""
+        if self.show_type and self.subject.type:
+            return f"{name}: {self.subject.type.name or ''}"
+        return name
--- a/gaphor/SysML/propertypages.py
+++ b/gaphor/SysML/propertypages.py
@@ -14,6 +14,7 @@ from gaphor.diagram.propertypages import (
 )
 from gaphor.SysML import sysml
 from gaphor.SysML.blocks.block import BlockItem
+from gaphor.SysML.blocks.proxyport import ProxyPortItem
 from gaphor.SysML.requirements.requirement import RequirementItem
 from gaphor.UML.classes.classespropertypages import AttributesPage, OperationsPage
 
@@ -87,7 +88,6 @@ class CompartmentPage(PropertyPageBase):
     def __init__(self, item):
         super().__init__()
         self.item = item
-        self.watcher = item.subject and item.subject.watcher()
 
     def construct(self):
         if not self.item.subject:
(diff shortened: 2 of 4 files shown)


sz332:
Hello, I checked out the feature and it works, but maybe we can make it even better.

Suggestions:

- Remove the aggregate dropdown because it has no meaning for a port in my opinion @danyeaw 
- Instead of show type checkbox have a dropdown where user can select: Display name, Display type, Display name and type
- If a Type is selected in the dropdown then the  the second dropdown should automatically set it's value to Display name and type

amolenaar:
> Remove the aggregate dropdown because it has no meaning for a port in my opinion @danyeaw

Fair point, it's inherited from `Property`, the base type of ports.

> Instead of show type checkbox have a dropdown where user can select: Display name, Display type, Display name and type

Not sure about this. Now the behavior is consistent with the _Show ..._ switches for classes and blocks.

Another question: do we also want to  show/hide the type for Properties in SysML?
At the moment I coded it as a special thingy for ProxyPorts only.

gaphor/SysML/blocks/proxyport.py
```diff
@@ -35,6 +40,15 @@ def update_shapes(self):
...
             style=text_position(self.connected_side()),
         )
+
+    def _format_name(self):
+        if not self.subject:
+            return ""
+
+        name = self.subject.name or ""
+        if self.show_type and self.subject.type:
+            return f"{name}: {self.subject.type.name or ''}"
```
danyeaw, line 53: I think here and in the Model Navigation, it should be `f"{name} : {self.subject.type.name}`, with a space before the colon.

image

danyeaw (approved):
This looks great! I agree that we don't want 3 options. If you don't want to name your port, then well don't name your port :)

Just one comment, I think there should be a space before the colon when we separate names from types.

gaphor/SysML/blocks/proxyport.py
amolenaar: Really? We do not do it for attributes and operations in classes.

sz332: I think we are mixing the model element and the presentation of the model element.

- The model element shall have a name because it is part of the model and we want to know what it is
- Whether we want to display it on a view (diagram) and in what way (only name, only type, both) is a totally different question

So, I would say that the user shall be able to configure those things **for the presentation element**.

danyeaw: > Really? We do not do it for attributes and operations in classes.

I think that is how it is specified in both UML / SysML for all types including in the presentation.

> I think we are mixing the model element and the presentation of the model element.

There is no requirement in UML and SysML that things should have names. I don't understand the use case for why you would want to name something but then hide the name. I realize that MagicDraw allows that, but we aren't building a MagicDraw clone.

sz332: My reasoning is to enable the user to be able to use a diagram really as a view. On a view I would like to display a representation of a model element a way it supports the communication I would like to achieve. Sometimes it makes sense to hide name or type, simply because it helps the communication by removing details that are deemed not important at that point when a diagram is being presented to the users. I find it totally valuable and use it day to day. It is not just magicdraw, you can have it in each and every modeling tool (EA, Visual Paradigm, and so on).

danyeaw: @sz332 We understand that diagrams are views. The point is adding more complexity to configuration makes the tool harder to use, sometimes that's a good tradeoff, in this case I don't understand why you would want to show types and hide names. If you don't care what things are named on the diagrams, then why would you care that they have a name at all? I don't get it. It is also consistent, if something has a name, you know that it is automatically displayed all the time. With those other tools, you see something only has a type displayed, and you don't know if that is because the Element has no name in the model or because the name is hidden.

sz332: But the question is: do you really want to automatically display names all the time? For example in case of pins, based on the abstraction level I am working on, I sometimes only display name, then both name and type, and in the final step I hide names where it is not important in order to decrease the noise on the diagram. So it might be a way of doing the modeling process, maybe other people are not doing it that way. 

I would say that we will have this problem in general for multiple type of elements, and I suggest to create a common solution and understanding for it (and probably adding some common code that can be easily used across all those components might make sense).

amolenaar: Let's stick with names. If that has to change, it has to change throughout the app, which is a fair bit of work.

image

In this image there's no space between `p1` and the colon, IMHO.

danyeaw: I can open a separate issue about the colons since it is outside this PR. I agree that the text in the PDF is hard to tell.

danyeaw (approved):

Issue #2181, opened 2023-03-26
Adding model root element to diagram then deleting element from diagram deletes whole model


skalemkiewicz:
### Describe the bug
If the model root element is added to a diagram and then the element is deleted from the diagram, the application deletes the entire model.

### To Reproduce
1. Open Gaphor.
2. Create New Model -> Generic. Gaphor creates a new model.
3. Open settings, ensure that "Remove Unused Elements" is enabled.
4. In the left pane, expand the model's root node to expand the default diagram.
5. Open the diagram.
6. Drag the model's root node from the left pane onto the diagram. Gaphor creates a package element on the diagram.
7. Select the package element that is on the diagram and delete it. Gaphor will remove the element from the diagram but will also remove the root node from the model in the left pane leaving only the diagram.

### Expected behavior
Gaphor should only remove the element from the diagram, not delete the entire model.

### Screenshots
![Before](https://user-images.githubusercontent.com/938809/227799184-e1062851-eea5-488e-80f6-2c9d3cb0de04.png)

![After](https://user-images.githubusercontent.com/938809/227799201-27ea7c9d-a687-4779-b4c1-ea510d5acd04.png)

### OS
Linux Pop!_OS 22.04 LTS

### Version
Version of Gaphor: 2.17.0

### Additional information
Output from error report:

Gaphor version:    2.17.0
Operating System:  Linux (6.2.6-76060206-generic)
Python version:    3.10.6
GTK version:       4.10.1
Adwaita version:   1.3.1
PyGObject version: 3.44.0
Pycairo version:   1.23.0
Cairo version:     1.16.0
Pango version:     1.50.14
Display:           X11Display

No errors have been intercepted. We’re good.

danyeaw:
Hi @skalemkiewicz, thanks for the bug report! It sounds like we need to check that an item doesn't have any nested or children items before removing it.

skalemkiewicz:
Hi @danyeaw, you're welcome! That seems reasonable -- or even to walk all of the child elements to see if they are being used before deleting. I guess it depends on how this feature was envisioned to be used. 

In Sparx Systems's Enterprise Architect, deleting an element from a diagram using the delete key will only remove the element from the diagram but keep it in the model. If you really want to remove the element from the diagram and the model, it requires a user to SHIFT+DELETE.

skalemkiewicz:
Sorry, clicked the wrong button and accidentally closed the ticket instead of just posting the comment.

[...]

PR #2491, opened 2023-07-15
Do not remove unused Packages with childen

danyeaw:
This PR fixes a bug that would allow for large portions of the model to be deleted if the last presentation of the owning Package was deleted.

### PR Checklist

### PR Type

- [X] Bug fix
- [ ] Feature

If the Remove Unused Elements option is selected, and you delete the last usage of a Package that has children like a package or activity, all the children are also deleted.

Issue Number: #2181 

### What is the new behavior?
Packages with children aren't removed automatically.

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/UML/sanitizerservice.py
+++ b/gaphor/UML/sanitizerservice.py
@@ -63,7 +63,12 @@ class SanitizerService(Service):
             return
 
         old_subject = event.old_value
-        if old_subject and not old_subject.presentation and deletable(old_subject):
+        if (
+            old_subject
+            and deletable(old_subject)
+            and not old_subject.presentation
+            and not old_subject.ownedElement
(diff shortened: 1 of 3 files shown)


amolenaar:
I do not think this is what we want: now every element that has a child as shown in the model browser, is not removed anymore.

I would expect that id I delete the last presentation of a class (with attribute), that the class will be removed. However, it's not.

amolenaar:
Maybe an explicit exception for Packages would do?

danyeaw:
@amolenaar Good catch, I forgot that ownedElement was that broad. I updated the PR for just Packages.

amolenaar (approved):
I think this change will reduce the level of surprise for many users.

PR #2493, opened 2023-07-15
Tree view should not collapse when an element is deleted

### LINKED ISSUE #2488: Do not close the tree when (last) element is removed
### Describe the bug
When removing an element from a diagram that is not referenced it will be also removed from the model. This is 
an expected behavior. However, it also closes the tree, which is pretty annoying. In my opinion it might refresh the
tree, that's fine, but should not close it.

### To Reproduce
1. Create a diagram
2. Open the tree
3. Add an element
4. Select the element
5. Press delete button
6. Element will be removed, but also tree is closing

It seems that when also deleting an element in the tree it will close the tree as well. The two cases might or might not related.

### Expected behavior
Do not close the tree.

### OS
- [ ] Linux
- [ ] macOS
- [X] Windows

### Version
Version of Gaphor: main branch

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

When you delete an element, the parent node in the model browser collapses (closes).

Issue Number: fixes #2488

### What is the new behavior?

Tree view does not collapse the node when an element changes inside.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/ui/tests/test_modelbrowser.py
+++ b/gaphor/ui/tests/test_modelbrowser.py
@@ -288,6 +288,28 @@ def test_drop_multiple_elements(model_browser, element_factory, event_manager):
     assert class_a not in model_browser.get_selected_elements()
 
 
+def test_unlink_element_should_not_collapse_branch(
+    model_browser, element_factory, event_manager
+):
+    package = element_factory.create(UML.Package)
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
@amolenaar This looks great!

Issue #2464, opened 2023-07-12
Support for call behaviour actions


Bambofy:
### Is your feature request related to a problem?
Without the call activities each nested activity forces the higher level activity diagrams to expand making them impossible to view.

### Describe the solution you'd like
Call actions which let us link between activity diagrams.

### Describe alternatives you've considered
1. Putting all of the hierarchical activities in one diagram makes it really difficult to see the high-level activity flows.
2. Putting the name of the activity in an action node doesn't convey the idea that it is linked to another activity.
3. Adding an inverted '⑃' symbol to the end of the action's name is ok but it isn't exactly the same thing as having the symbol standardized in the bottom right corner.

### Additional context
1. Reference in the UML specification (https://www.omg.org/spec/UML/2.5.1/PDF) heading "16.3.4.1 Call Behaviour Actions".

<img width="161" alt="Screenshot 2023-07-12 at 07 47 48" src="https://github.com/gaphor/gaphor/assets/1304898/88644f09-03a8-4fac-bf4c-19e0b4c29121">

2. Example of decomposed action (slide 24 "Rate and buffer reduction") https://www.i3s.unice.fr/~map/Cours/MASTER_TSM/Cours5_Activity.pdf

Bambofy:
Shortened the problem description.

marek-piirikivi:
I would like to work on this.

This is what I have currently.
![image](https://github.com/gaphor/gaphor/assets/45357878/298d765a-de91-4ef0-97c4-a61c4dc2f457)

What kind of connection should be established between call behavior action and an activity?

amolenaar:
Fixed by #2495.

PR #2495, opened 2023-07-16
Add CallBehaviorAction

marek-piirikivi:
### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: #2464 

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/UML/actions/action.py
+++ b/gaphor/UML/actions/action.py
@@ -5,6 +5,7 @@ from gaphor.diagram.presentation import ElementPresentation, Named
 from gaphor.diagram.shapes import Box, Text, draw_border, stroke
 from gaphor.diagram.support import represents
 from gaphor.UML.recipes import stereotypes_str
+from gaphor.UML.umlfmt import format_call_behavior_action_name
 
 
 @represents(UML.Action)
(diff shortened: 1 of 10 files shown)


danyeaw:
This is an awesome contribution, thanks @marek-piirikivi! This could be a follow-up PR, but do you think there should be a way to go to the selected Behavior from the CallBehavior?

amolenaar (approved):
Excellent!

The [Invocation Actions diagram](https://gaphor--2495.org.readthedocs.build/en/2495/models/uml/10._Actions.html#invocation-actions) also looks really nice.

gaphor/UML/actions/action.py
```diff
@@ -32,6 +32,64 @@ def __init__(self, diagram, id=None):
...
+            draw=self.draw_border_with_fork,
+        )
+
+        self.watch("subject[CallBehaviorAction].behavior.name")
+
+    def name(self):
+        """Name conforms to UML2.5.1 16.3.4.1 naming description"""
+        if self.subject.behavior and not self.subject.name:
+            return self.subject.behavior.name
+        return self.subject.name or ""
```
amolenaar, line 61: One thing to improve it even more is to move this code to a `format` function in `gaphor/UML/umlfmt.py`. Then the activity name will also be used as fallback in the model browser.

Just noticed this feature is not mentioned in the documentation :see_no_evil:

marek-piirikivi:
> Excellent!
> 
> The [Invocation Actions diagram](https://gaphor--2495.org.readthedocs.build/en/2495/models/uml/10._Actions.html#invocation-actions) also looks really nice.

@amolenaar I tried to be as authentic to the UML2.5.1 document and capture everything that seemed relevant. I noticed that Action itself already defines relationship to the InputPin and OutputPin; and so does CallAction+InvocationAction. So, the references are now in uml.py duplicated. :thinking: I am not sure how to address this; or what to do in this instance. Seems as if it is just redefinition of the same thing to use different names for the same thing - in CallAction context, all input pins should be called arguments. What do you think?

marek-piirikivi:
> This is an awesome contribution, thanks @marek-piirikivi! This could be a follow-up PR, but do you think there should be a way to go to the selected Behavior from the CallBehavior?

Yes. I think so. It would be a good follow-up.

amolenaar:
> > Excellent!
> > The [Invocation Actions diagram](https://gaphor--2495.org.readthedocs.build/en/2495/models/uml/10._Actions.html#invocation-actions) also looks really nice.
> 
> @amolenaar I tried to be as authentic to the UML2.5.1 document and capture everything that seemed relevant. I noticed that Action itself already defines relationship to the InputPin and OutputPin; and so does CallAction+InvocationAction. So, the references are now in uml.py duplicated. thinking I am not sure how to address this; or what to do in this instance. Seems as if it is just redefinition of the same thing to use different names for the same thing - in CallAction context, all input pins should be called arguments. What do you think?

I have no clear strategy for those. As long as a relation is not actively used by the application I nowadays tend to leave it out.

The UML meta model is not always consistent. Often a value subsets a value that is not a derived unions; something that our Gaphor model cannot deal with and will ignore.

gaphor/UML/actions/action.py
marek-piirikivi: I just noticed that there is a slight conflict with the name formatting when looking at SysMLv1.6. SysML requires activity name, and optionally you can also include the action name. How should we proceed?

image

amolenaar: We can keep is as is for now.

amolenaar (approved):

amolenaar:
Let's call it a day and merge this. A really well executed PR.

Thanks @marek-piirikivi!

PR #2513, opened 2023-07-18
Separate build and test

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

### What is the new behavior?

* Separated install and test steps - this makes it more clear to the tests.
* Move coverage config to `.coveragerc`. It looks like pytest-cov is not picking up the config from `pyproject.toml` (many warnings on not-found source files).

## ...

We're currently testing with X11 only. It may be worthwhile to test with a Wayland backend as well.


danyeaw (approved):
Nice refactoring of the CI pipeline!

Issue #2504, opened 2023-07-17
Dragging elements from one model to another results in errors.


Bambofy:
### Describe the bug
When i try to copy an element from one model to aother model (so that i can reuse it) gaphor reports an error and the copying doesn't work.

### To Reproduce
1. Open a  model with gaphor.
2. Open a different model with gaphor so you have 2 seperate windows. 
3. Create a class element in one model.
4. Try to drag the class element from one model's element tree to the others.

Cannot open the error reporting tool once the error occurs.

### Expected behavior
A copy of the element to be created in the other (dragged onto) model.
### OS
- [ ] Linux (Please put in notes the specific distro)
- [x] macOS
- [ ] Windows

### Version
Version of Gaphor:
2.19.3

Issue #2422, opened 2023-07-06
Package import from file/clipboard


mikekidner:
### Is your feature request related to a problem?
To help with re-use, collaboration and consistency it would be great to be able to import packages from other models. 

### Describe the solution you'd like
Import packages from other models. Either via: 
drag & drop into left hand menu. 
Copy-Paste into left hand menu, or
Reference imported package in a diagram as a URL

### Describe alternatives you've considered
I've looked at appending the XML itself, but it appears that element ID's are not unique between two models.

### Additional context
As per sysML diagrams, which would possibly keep a 'live' connection to the source package.
<img width="791" alt="Screen Shot 2023-07-06 at 1 12 34 pm" src="https://github.com/gaphor/gaphor/assets/50702106/0ccf73e1-c429-4b75-82f9-08b0a61b3796">

However the drag/drop copy/paste maybe be more intuitive for users

amolenaar:
Hi @mikekidner,

You can copy element from one diagram to another, also in another model. You should select elements in a diagram and copy them, then paste in another diagram.

It's not possible to copy directly from the model browser (top left).

danyeaw:
I think this use case is a bit different than copy and paste. It would be useful to allow projects to use other projects, for things like common elements and customizations used across a bunch of models.

This would be very powerful, but would also take some considerable work I think. The git equivalent to this would be a submodule.

mikekidner:
@amolenaar thanks for the note about copy/paste - I hadn't realised, and that does solve the immediate issue. @danyeaw I agree that linking to other models/repos will be a large amount of work - but it may also be relevant for sysMLv2 implementation,  where there appears to be have emphasis on API's to remote models, (which makes it a longer term feature).

mikekidner:
When using copy/paste from I noticed that the content of a package is not brought across, if the package is copied from one diagram to a diagram in another model. Same goes for blocks, the component blocks do not come into the new model file.

amolenaar:
That's correct. It copies the model element with the minimal set of content. For a class operations and attributes are copied along.

For copying to other models (deep copy) it sounds reasonable to copy the contents of a package as well.

I may be misunderstanding you. Blocks are copied with their parts and references, right?

mikekidner:
Hi @amolenaar, I couldn't get blocks to bring their parts over.
I'm copying the parent block from a diagram, into a new model. Only the parent block comes across.

[...]

PR #2519, opened 2023-07-19
Allow only changing element ownership in the same model

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Currently an exception is raised when you drag elements from one session and drop them in the model browser of another session

Issue Number: #2504

### What is the new behavior?

This PR fixes the exception by explcitly checking if elements are part of the same model.

- [ ] Yes
- [x] No

### Other information

To support DnD between models we need to use the internal copy functionality. But for that the copy functionality should be improved, as described in #2422.


danyeaw (approved):
Looks good!

Issue #2444, opened 2023-07-09
Support diagram model element type in SysML


marek-piirikivi:
### Is your feature request related to a problem?

Diagram header only supports diagram name. It is not possible to explicitly know what modelElementType and model element (modelElementName) the diagram corresponds to. This is important information in SysML to properly navigate the model.

### Describe the solution you'd like

1. I would like to be able to properly select diagram modelElementType and corresponding modelElementName (if diagram kind requires) according to the rules described by SysML v1.6 standard specification. Below is a screenshot of the specification describing which diagram kind options are available for each diagram.

![image](https://github.com/gaphor/gaphor/assets/45357878/1e2c639e-2f49-4b4e-83d2-905630fdd0cd)

2. I would like to see diagram header generation automated and contain the information in SysML standard proposed way: <diagramKind> [modelElementType] <modelElementName> [diagramName].

This solution would promote correct SysML usage and improve user experience: a) by guiding users to use diagrams more correctly reducing the learning curve; and b) by simplifying model navigation and understanding of the model.

### Describe alternatives you've considered

Currently, I have manually inserted necessary information into the diagram name. See the following example:
![image](https://github.com/gaphor/gaphor/assets/45357878/a441d72a-8cba-46d0-a4ab-33c26223d10c)
This serves the purpose of the simplification of the model understanding.

### Additional context

I would like to propose more details about the diagram modelElementType and modelElementName selection.
* For the new diagrams, default modelElementType should be package which would take the composing package element as diagram modelElementName, if the diagram type is bdd, pkg, req, uc.
* For other new diagrams - act, ibd, par (not supported by gaphor, but hopefully soon), seq, sm - new corresponding element should be created. 
* For the existing elements, corresponding diagram can be created by right-clicking the element and then selecting "New diagram" which would list only the diagrams where the element type can be valid element type.

I would consider this issue complete with a minimum of having an option to select diagram kind and corresponding diagram kind elements from the dropdown; and auto-updating rendering of the diagram header.

danyeaw:
@marek-piirikivi Thanks for the idea. I think #2213 is related to this, since it includes the idea that the boundary of a diagram represents the diagram kind.

amolenaar:
It may be more friendly to use the abbreviated diagram type name.

E.g. 

```
[cls] My Class Diagram
[act] My Activity
[sd] My Sequence Diagram
```

marek-piirikivi:
[...]

PR #2520, opened 2023-07-19
Support diagram model element type in SysML

marek-piirikivi:
### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: #2444

- [ ] Yes
- [x] No

### Other information

From #2444 description, technically, you can have all kinds of diagram element types, even those that are illegal according to the SysML, with these changes. I would like to do a follow-up pull request that would additionally constrain diagram creation to avoid creating illegal diagrams. The current solution works by taking the diagram owning element as diagram model element. This seems to be the common way to define the model element based on how Cameo, and Enco SOX handle it. For example, activity diagram should always be within activity, not outside for it to be considered valid SysML (unenforced at the moment).

Most of the changes are related to providing necessary context to the diagram header rendering in various use contexts. I am not very satisfied with the solution; so feel free to offer alternative suggestions on how to make diagram header modeling language customizable.

### DIFF
--- a/gaphor/C4Model/modelinglanguage.py
+++ b/gaphor/C4Model/modelinglanguage.py
@@ -24,3 +24,6 @@ class C4ModelLanguage(ModelingLanguage):
 
     def lookup_element(self, name):
         return getattr(c4model, name, None) or getattr(diagramitems, name, None)
+
+    def format_diagram_label(self, diagram) -> str:
+        return str(diagram.name)
--- a/gaphor/RAAML/modelinglanguage.py
+++ b/gaphor/RAAML/modelinglanguage.py
@@ -25,3 +25,6 @@ class RAAMLModelingLanguage(ModelingLanguage):
 
     def lookup_element(self, name):
         return getattr(raaml, name, None) or getattr(diagramitems, name, None)
+
+    def format_diagram_label(self, diagram) -> str:
+        return str(diagram.name)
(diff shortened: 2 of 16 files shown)


marek-piirikivi:
@danyeaw I am not yet entirely sure whether we want to do it like that; because in the long run I see that it would help us if we differentiated between SysML diagrams and UML diagrams. First hint of that came when I refactored exports to accommodate modeling language to get proper rendering for the SysML diagrams - in some places I couldn't get it correctly.

You may, however, still review it, but it is kinda "mess" in terms of exporting.

marek-piirikivi:
@danyeaw I converted this to draft. I'll try with a "better" approach first that would be more forward compatible, because the current solution encourages "hacks" and populates many locations, which I would prefer to avoid. :smile:

danyeaw:
@marek-piirikivi Cool, I like the new diagram headers, very nice!

amolenaar:
I like the functionality.
I like that the format can be changed depending on the profile.

Downside is that we have to pass the modeling language around. Instead it would be nice if we can make the diagram render it's own title. Rendering the title should not be a responsibility of the `ModelingLanguage`. Or maybe it is?

marek-piirikivi:
Closing this because an alternative PR was opened https://github.com/gaphor/gaphor/pull/2540

Issue #2522, opened 2023-07-20
Note or Remark box is different between diagrams for the same block


tompkins-ct:
### Describe the bug
Notes missing from blocks across different diagrams. 

Writing a note in the note or remark box will show up on a given block in one diagram but the note or remark is not visible in other diagrams (the box is blank).

### To Reproduce
1. Create a diagram A
2. Add a block A
3. Create a new diagram B
4. drag the previous block A onto the diagram
5. Add a comment in the `note or remark` box for block A
6. check block A on diagram A
7. Notice text is missing

### Expected behavior
I expected notes attached to a block to travel with the block.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [x] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.19.2

sz332:
This is even true for a simple class diagram, creating a class.

sz332:
@danyeaw @amolenaar it looks like that Note/Remark is associated to a diagram element and not a model element. Is it a bug or a feature?

amolenaar:
The note should be on the model element. This is a bug for sure.

sz332:
The problem is around propertypages.py. 

It seems that self.subject is pointing to gaphor.UML.classes.klass.ClassItem for class NotePropertyPage which is a visual representation of the model element and not the model element itself.

Is this a problem? (without knowing the internals I would say that any property setting in general should work on the model element itself and not the visual representation UNLESS we are really want to modify the visual representation)

It also seems that when we drag an existing model element into a new diagram then no new uuid is generated for the new ClassItem. That in my opinion is a bug, but might be unrelated.

@amolenaar any idea how to progress?

amolenaar:
> (without knowing the internals I would say that any property setting in general should work on the model element itself and not the visual representation UNLESS we are really want to modify the visual representation)

I would expect that too.

> It also seems that when we drag an existing model element into a new diagram then no new uuid is generated for the new ClassItem. That in my opinion is a bug, but might be unrelated.

That would be a serious issue, since elements (also ClassItem) are managed by their id. I'm not able to reproduce it. Can you create an issue for this with steps to reproduce?

> @amolenaar any idea how to progress?

For the property editor we check the model element and then the presentation. I think the presentation is overriding the model element at some point. However, it would also need a migration (upgrade) so the remark is moved from the presentation to the model element.

sz332:
[...]

PR #2525, opened 2023-07-21
Fix notes: should be applied to model elements

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Notes are applied on diagram items, so they do are not visible in other diagrams

Issue Number: #2522

### What is the new behavior?

Notes are applied to model elements.
During loading, notes are transferred to the model elements.

- [ ] Yes
- [x] No


danyeaw (approved):
Nice fix @amolenaar!

PR #2529, opened 2023-07-22
Fix weird pin rendering

### LINKED ISSUE #2428: Appearance/placing of ports on activity diagram is unusual
### Describe the bug
When placing an input/output pin or proxy port on an activity or block diagram, the pin does not automatically attach to a nearby action or block, and it doesn't appear as a pin. The pin can be dragged onto an activity and seems to function as expected, but the appearance looks strange. 

### To Reproduce
1. Add an action to an activity diagram and place an input/output pin
2. Attempt to attach the pin to the action
3. Same steps for adding a block and proxy pin to a block diagram

### Expected behavior
The pin to appear as expected (small square on the outside of the action). Notably, the pin behaviour seems to be correct on the coffee machine example model, but it is only on my own model that the pins behave this way.

### Screenshots
Straight after adding the pin to the activity diagram:
image
After adding the pin to the action and connecting an object flow arrow:
image

### OS
- [ ] macOS
- [x ] Windows

### Version
Version of Gaphor: 2.19.1

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Pins are rendered weirdly.

Issue Number: fixes #2428 

### What is the new behavior?

Border rendering van handle positions where top/left is not (0, 0).

### DIFF
--- a/gaphor/diagram/shapes.py
+++ b/gaphor/diagram/shapes.py
@@ -52,21 +52,21 @@ def stroke(context: DrawContext, fill=True, dash=True):
 
 def draw_border(box, context: DrawContext, bounding_box: Rectangle):
     cr = context.cairo
-    d = context.style.get("border-radius", 0)
     x, y, width, height = bounding_box
+    d = min(context.style.get("border-radius", 0), width / 2.0, height / 2.0)
 
-    cr.move_to(x, d)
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
Nice!

PR #2540, opened 2023-07-23
Support SysMLDiagram type and diagram type specific header formatting

marek-piirikivi:
* Adds SysMLDiagram which extends regular Diagram
* Adds diagram type specific header formatting
* All SysML diagrams created under SysML profile (modeling language) will be created as SysMLDiagram and treated as such throughout the model.

### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: #2444

- [ ] Yes
- [x] No

### Other information

As a follow-up, it may be good to also have "refactor" option for regular diagrams which should be SysML diagrams.

This change and the ability to differentiate SysML diagrams opens up a large possibility to address SysML specific constraints and defaults more gracefully.

### DIFF
--- a/gaphor/SysML/diagramlabel.py
+++ b/gaphor/SysML/diagramlabel.py
@@ -1,0 +1,33 @@
+from gaphor.diagram.diagramlabel import diagram_label
+from gaphor.SysML.sysml import SysMLDiagram, Block, ConstraintBlock, Requirement
+from gaphor.UML.uml import Activity, Package, Interaction, Profile, StateMachine
+
+
+@diagram_label.register(SysMLDiagram)
+def sysml_diagram_label(diagram):
+    def element_type(el):
+        if isinstance(el, Activity):
+            return "activity"
+        if isinstance(el, ConstraintBlock):
+            return "constraint block"
+        if isinstance(el, Block):
+            return "block"
+        if isinstance(el, Interaction):
+            return "interaction"
+        if isinstance(el, Profile):
+            return "profile"
+        if isinstance(el, Package):
+            return "package"
+        if isinstance(el, Requirement):
+            return "requirement"
+        if isinstance(el, StateMachine):
+            return "state machine"
+        return ""
+
+    if diagram.element:
+        return (
+            f"[{element_type(diagram.element)}] {diagram.element.name} [{diagram.name}]"
+        )
+    else:
+        # TODO: SysML specification does not allow parentless elements, but since it is not constrained (yet), it may happen.
+        return diagram.name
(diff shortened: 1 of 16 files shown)


gaphor/diagram/headerlabel.py
```diff
@@ -0,0 +1,8 @@
+from functools import singledispatch
+
+
+def get_diagram_name(diagram):
+    return diagram.name
+
+
+get_header_label = singledispatch(get_diagram_name)
```
amolenaar, line 8: I would call it `diagram_label`, instead of `get_header_label`.

```
@singledispatch
diagram_label(diagram: Diagram):
    return diagram.name
```

amolenaar (approved):
I like this approach better, indeed.

Having the `SysMLDiagram`,opens the door to other types of diagrams (like tabular diagrams).

I would call the function differently, but apart from that I'm okay with this solution.

(Note to self: `get_icon_name` should be renamed to `icon_name`)

gaphor/diagram/headerlabel.py
marek-piirikivi: > (Note to self: get_icon_name should be renamed to icon_name)

I might as well do that myself with a following commit.

gaphor/SysML/diagramlabel.py
```diff
@@ -0,0 +1,27 @@
...
+            return "profile"
+        if isinstance(el, Package):
+            return "package"
+        if isinstance(el, Requirement):
+            return "requirement"
+        if isinstance(el, StateMachine):
+            return "state machine"
+        return ""
+
+    return f"[{element_type(diagram.element)}] {diagram.element.name} [{diagram.name}]"
```
amolenaar, line 27: I ran into one issue: when a diagram does not have a parent (`diagram.element`), this raises an error.

```suggestion
    if diagram.element:
        return f"[{element_type(diagram.element)}] {diagram.element.name} [{diagram.name}]"
    else:
        return diagram.name
```

gaphor/diagram/headerlabel.py
amolenaar: Thanks!

gaphor/SysML/diagramlabel.py
marek-piirikivi: Good catch! According to the SysML spec, diagrams have to have a parent. At the moment, we do not (yet?) have these constraints, so we can create diagrams wherever we like (technically we have few constraints on this, but still, very loose). Your proposed default makes sense.

amolenaar:
Thanks @marek-piirikivi !

Issue #2484, opened 2023-07-15
Input and output pins shall not be detachable from an action


sz332:
### Describe the bug
Input and output pins are belonging to an action. Free floating pins do not make any sense. It should be not allowed.

### To Reproduce

- Create an activity diagram
- Add input or output pin
- Drag and drop pin
- Pin is free floating

### Expected behavior
When pin is added to an activity it cannot be moved out from the action but only deleted

### Screenshots
![image](https://github.com/gaphor/gaphor/assets/8182138/1e1cfbb5-565b-4456-94e6-028180678d48)

### OS
- [X] Linux (latest)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: main branch

marek-piirikivi:
Would this be an acceptable solution? If "yes", I'll prepare a pull request (if someone is not already working on it. And I suppose I'll also apply similar behavior on the ports as well.



https://github.com/gaphor/gaphor/assets/45357878/85e9e54e-79d7-46c5-8e1e-de49246f3c62

sz332:
Yes, absolutely, it works totally as expected. Please prepare a PR.

sz332:
Solved in #2542, closing the issue.

Issue #2450, opened 2023-07-10
Disallow free-floating ports for SysML


sz332:
### Describe the solution you'd like
1, Ports cannot exist without a container, so creating a port not as part of a "block" should be disallowed.
2, When a port is added to a block, it shall be added to the outside line.
3, When a port as part of the block is being moved, it shall move only around the border, not inside or outside.

amolenaar:
The way you describe it, it looks a lot like Parameters on Activities: edit delete parameters in the Property Editor. Are you proposing a similar approach for ports?

sz332:
I would simplify the possible actions a user can do to the following:

Adding a new port:

- Select a port icon on the toolbar
- Click outside a non-block element's boundary: nothing happens
- Moves the mouse to the boundary of the block: show rectangle. Clicks: display port on the edge.
- Moves the mouse inside the block and clicks: add ports to the edge.

To move a port around:

- Drags a port at the edge: can be moved only around the edges. Not inside, not outside.

To remove a port:

- Click on the port and press delete or anything alike.

amolenaar:
> Click outside a non-block element's boundary: nothing happens

I think it's weird to let nothing happen. We should at least provide feedback to the user.

A ProxyPort is only created in the model when a Port (item) is connected to a block.

sz332:
I think we just found another problem.

Create a block. Select proxy port. Click on the diagram outside the block. I will see a port being displayed free floating. It is although not associated to a block, that is true, but I would argue that a port cannot exist (displayed) on a diagram as NOT part of a block.

sz332:
I tried out the parameters on activities and it wasn't intuitive to be honest. I started looking for the activity parameter on the toolbar, but did not find it, and then I by mistake clicked on the activity and then I saw, okay, this is how I can add a new parameter. 

The current functionality is great, the only thing you need to do is to disallow adding a port to a diagram alone.

amolenaar:
We could also do something with color. For example make the port red if it is not connected.

(just thinking through some options)

amolenaar:
The underlaying mechanism is similar to the issues you describe in #2484, and #2481.

The elements in the model are only created when the pin/port/relationship is properly connected. The underlaying model is always consistent.

I do not think Gaphor should be very restrictive/prescriptive in that respect. A user should be able to move a relationship aside and reconnect it at a later point. The model is in an "invalid state" for a while.

Also, what is valid and not valid depends on the methodology you follow. SysML is only a generic language.

[...]

PR #2542, opened 2023-07-24
Disallow disconnection by dragging for proxy port, activity pin after connection; ensure that activity parameter node is always stuck to the activity when moving.

marek-piirikivi:
I added StickyAttachedHandleMove which can be registered to the items that once connected, can not be disconnected via dragging and can only be moved along the border.

I added this behavior to activity parameters (only dragging functionality; otherwise was already there inability to disconnect); pins (both input and output); proxyport.

### PR Checklist

### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: #2484, #2450 (only point 3 from the initially proposed solution)

- [ ] Yes
- [x] No

### Other information

From the both related issues, not having free floating port/pin at all is not yet addressed - it is still possible to have free floating port/pin if it has never been attached to anything, but at least it is not possible to detach it anymore other than explicitly deleting.

### DIFF
--- a/gaphor/SysML/blocks/__init__.py
+++ b/gaphor/SysML/blocks/__init__.py
@@ -1,3 +1,4 @@
+import gaphor.SysML.blocks.blockmove
 import gaphor.SysML.blocks.connectors
 import gaphor.SysML.blocks.datatype
 import gaphor.SysML.blocks.group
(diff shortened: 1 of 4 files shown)


amolenaar:
Works like a charm. Thanks @marek-piirikivi!

Issue #2598, opened 2023-08-02
Opening multiple models cause changing of background color of the diagrams


shobeira:
### Describe the bug
The background of the recent model opened is applied to all model opened before the recent one.

### To Reproduce
Open multiple models with Gaphor.

### Expected behavior
Each model has its own style sheet and is expected to be based on that. Regardless of what other model style sheets are.

### OS
- [ Ubuntu 22.04] Linux

### Version
Version of Gaphor: 2.19.3

amolenaar:
@shobeira Thank you very much for report.

danyeaw:
@all-contributors please add @shobeira for bug

shobeira:
> @all-contributors please add @shobeira for bug

That's nice of you mate! didn't expect it. Cheers.

shobeira:
@danyeaw that's nice of you mate! didn't expect it. Cheers.

PR #2602, opened 2023-08-04
Fix diagram background shouldn't be shared between open models

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: #2598 

### What is the new behavior?

Diagram background is properly set on a per-diagram basis.

- [ ] Yes
- [x] No

### Other information

Additionally the style context is properly removed when the diagram is closed.


danyeaw (approved):
This fix looks great, and I tested it and this fixes the original issue :+1:

Issue #2584, opened 2023-07-30
Add Support for KerML


danyeaw:
### Is your feature request related to a problem?
This is the first step with #1399, to add support for KerML

### Describe the solution you'd like
The [SysML v2 Pilot Implementation](https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation) uses [Xtext](https://eclipse.dev/Xtext/) to define the KerML language, since SysML is no longer built on top of UML. In Python, we have TextX, which can use Xtext like definitions, with some changes that are documented here: https://textx.github.io/textX/3.1/about/comparison/#difference-to-xtext-grammar-language. Xtext is a more traditional grammar parser and TextX uses a PEG parser.

One of the tricky things that we need to figure out is that the Xtext definitions use ECore, and import an EPackage of the SysML metamodel schema to define KerML.

There is a JSON version of that schema definition here: https://www.omg.org/spec/SysML/20230201/SysML.json. I think that using the JSON as its own metamodel that the TextX metamodel can reference would be ideal.

sz332:
Can we use some python ecore implementation like https://github.com/pyecore/pyecore or is it totally not relevant to the problem?

danyeaw:
There is an old fork of TextX that tried to add pyecore support to TextX: https://github.com/aranega/textX
It isn't documented though. I'm not sure what the advantage would be to use ECore though in this case if we can parse the JSON metamodel.

sz332:
The idea is not really just having ECore, because it is a meta-metamodel but to rely on other people's work that built up for example a full UML 2.5.x model on top of that **as well as** the KerML metamodel.

danyeaw:
Is there a .ecore file available for SysMLv2 or UML somewhere?

sz332:
For SysML V2 I think this is what you are looking for:

https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/tree/master/org.omg.sysml/model

For UML:

https://git.eclipse.org/c/uml2/org.eclipse.uml2.git/tree/plugins/org.eclipse.uml2.uml/model?id=95a47511090741046b3e492d2f185632c36134ff

amolenaar:
One way forward is to implement the _Root_ (and maybe _Core_) parts of KerML in our core model. That will allow for a unified basis for all languages. The _Root_ package doesn't look that much different from the UML root package.

pbrown12303:
[...]

PR #2639, opened 2023-08-12
Add Initial Data Models for SysML v2

danyeaw:
This is an initial implementation of what was discussed in #2584, to create our Python data models for SysML v2 by using the SysML Pilot Implementation EMF ecore meta models with pyecore.

I also filed https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/issues/496, since it appears that the SysML ecore doesn't actually depend on the KerML ecore, it redefines everything. To work around this, I did a lot of manual editing of the resulting pyecoregen models. We could work towards getting this fixed upstream and automating any other changes needed.

What does everyone think about this approach?

### PR Checklist

### PR Type

- [ ] Bug fix
- [X] Feature

Issue Number: #2584

- [ ] Yes
- [X] No

### DIFF
--- a/gaphor/KerML/__init__.py
+++ b/gaphor/KerML/__init__.py
@@ -1,0 +1,365 @@
+from gaphor.KerML.kerml import (
+    eClassifiers,
+    LiteralString,
+    EndFeatureMembership,
+    Structure,
+    Class,
+    DataType,
+    LibraryPackage,
+    LiteralExpression,
+    LiteralRational,
+    LiteralInteger,
+    SelectExpression,
+    CollectExpression,
+    LiteralBoolean,
+    NullExpression,
+    LiteralInfinity,
+    AssociationStructure,
+    Invariant,
+    ReturnParameterMembership,
+    SuccessionItemFlow,
+    BindingConnector,
+    Comment,
+)
+from gaphor.KerML.kerml import eClass
+from gaphor.KerML.kerml import (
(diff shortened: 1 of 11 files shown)


danyeaw:
I created https://github.com/pyecore/pyecore/pull/135 to fix the Dependency Review failure.

danyeaw:
> For this to really work, the code should interact with our event system and emit the right events (attribute changes, association updates, etc.).

The intention of this was an equivalent of our UML data model that we have today but for SysML. I think you are pointing out that since Gaphor assumes that the underlying model is UML, we may need to make significant updates to accomodate a new one. Is that what you mean?

> Does this imply we should also be able to load/save ecore files in Gaphor?

Potentially, but I didn't have that use case in mind.

amolenaar:
> > For this to really work, the code should interact with our event system and emit the right events (attribute changes, association updates, etc.).
> 
> The intention of this was an equivalent of our UML data model that we have today but for SysML. I think you are pointing out that since Gaphor assumes that the underlying model is UML, we may need to make significant updates to accomodate a new one. Is that what you mean?

I think it's fine that the model is not based on our UML model. It should be based on the Core model though, since Element contains the nitty gritty details on how changes are propagated from the model to the rest of the application.

amolenaar:
@danyeaw Do you still want to try to make Gaphor work with ECore? Or can we close thos PR?

Issue #2583, opened 2023-07-30
Using images in diagrams


sz332:
### Is your feature request related to a problem?

I would like to be able to see images (like png, jpeg) on a diagram, move it around, and resize. It would be nice to be able to see the same image on multiple diagrams as well.  

### Describe the solution you'd like

The way an image could be created are the following:

-  Drag and drop from file browser
-  Copy paste
-  Create a new image element in the model and selecting an image as a property 

All these operations would create an image element in the model that would contain the image, not just a reference to the image.

Then this image could be dragged to the model into the diagram, then the image will be displayed as a box. It can be moved around and/or resized as any box like model element.

danyeaw:
Closed by #2651.

PR #2651, opened 2023-08-17
Add picture as core element

sz332:
### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Pictures cannot be displayed on a diagram.

Issue Number: #2583 

### What is the new behavior?

Pictures can be displayed on a diagram.

- [ ] Yes
- [X] No

### Other information

This is a hopefully fixed version of the previous PR.


gaphor/core/modeling/coremodel.py
```diff
@@ -54,6 +54,11 @@ class RefChange(PendingChange):
     property_ref: _attribute[str] = _attribute("property_ref", str)
 
 
+class Picture(Element):
+    content: _attribute[str] = _attribute("content", str)
+    dimension: _attribute[str] = _attribute("dimension", str)
```
amolenaar, line 59: Shouldn't we just use a width and height property?

```suggestion
    width: _attribute[int] = _attribute("width", int)
    height: _attribute[int] = _attribute("height", int)
```

tests/test_architecture.py
```diff
@@ -43,6 +43,7 @@ def test_diagram_package():
         .may_import("gaphor.diagram*")
         .may_import("gaphor.UML.recipes")
         .may_import("gaphor.UML.uml")
+        .may_import("gaphor.ui.*")
```
amolenaar, line 46: Let's keep this for now.

I consider this a regression, though. But it will require a bit more work internally to restructure code ina way that we can use something like a file dialog from the diagram module too.

amolenaar (changes requested):
Looks good. I think we can save width and height separately, though, instead of constructing a string.

gaphor/core/modeling/coremodel.py
sz332: Let me try it out whether it works as expected.

sz332: Commited a full change.

sz332: @danyeaw can you please run the makefile for image generation, it still crashing for me :(

danyeaw: Done!

poetry.lock
```diff
@@ -1,9 +1,10 @@
-# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
```
danyeaw, line 1: If I run `poetry lock --no-update` with Poetry 1.5.1, I am seeing a lot of changes to the lock file. Are you using an older version of Poetry?

pyproject.toml
```diff
@@ -55,6 +55,7 @@ sphinx-copybutton = { version = "^0.5.0", optional = true }
 sphinx-intl = { version = "^2.1.0", optional = true }
 myst-nb = { version = "^0.17.1", optional = true }
 furo = { version = ">=2022,<2024", optional = true }
+pillow = "^10.0.0"
```
danyeaw, line 58: This is a nitpick, it would be better to move pillow up with the other non-optional dependencies.

gaphor/diagram/general/generalpropertypages.py
```diff
@@ -88,3 +97,74 @@ def construct(self):
...
+        return builder.get_object("picture-editor")
+
+    @transactional
+    def _on_select_picture_clicked(self, button):
+        open_file_dialog(
+            gettext("Select a picture..."), self.open_files, filters=FILTER_IMAGES
+        )
+
+    @transactional
+    def open_files(self, filenames):
```
danyeaw, line 137: I really like the resize that happens when we run the set Default Size button. Could we run this resize when we first load an image as well?

gaphor/diagram/general/picture.py
```diff
@@ -0,0 +1,111 @@
...
+        cr.restore()
+
+    def _from_pil(
+        self,
+        im: Image,
+        alpha: float = 1.0,
+        format: cairo.Format = cairo.FORMAT_ARGB32,
+    ) -> cairo.ImageSurface:
+        """
+        :param im: Pillow Image
```
danyeaw, line 100: We are using the Google Docstring format instead of the reST one.

danyeaw (commented):
Nice job @sz332, this is a cool feature.

Would it be possible to constrain the handles at the corners to the image border? It seems unnatural that it allows you to drag them larger than the image itself.

gaphor/diagram/general/picture.py
```diff
@@ -0,0 +1,111 @@
...
+            scale_xy, surface = self.create_default_surface()
+        else:
+            scale_xy, surface = self.create_content_surface()
+
+        cr = context.cairo
+        cr.save()
+        cr.scale(scale_xy, scale_xy)
+        cr.set_source_surface(surface, 0, 0)
+        cr.paint()
+        cr.restore()
```
danyeaw, line 91:
```suggestion
        cr.restore()
       
        context_width = surface.get_width() * scale_xy
        context_height = surface.get_height() * scale_xy
        self.width = min(self.width, context_width)
        self.height = min(self.height, context_height)
```

After our discussion, I messed with this and this constrains the handles to match the image size. :+1:

gaphor/diagram/general/generalpropertypages.py
```diff
@@ -88,3 +97,76 @@ def construct(self):
...
+                "select-picture": (self._on_select_picture_clicked,),
+                "set-default-size": (self._on_default_size_clicked),
+            },
+        )
+        return builder.get_object("picture-editor")
+
+    @transactional
+    def _on_select_picture_clicked(self, button):
+        open_file_dialog(
+            gettext("Select a picture..."), self.open_files, filters=FILTER_IMAGES
```
danyeaw, line 133: Since we aren't passing in the parent window, we are getting `Gtk-Message: 12:26:09.933: GtkDialog mapped without a transient parent. This is discouraged.`

I am not sure the best way to get the parent window of the propertypage to be honest, maybe @amolenaar has an idea.

danyeaw: The file filter also isn't working on Windows (it seems to work in Linux). I think we may need to modify it so that the filters are passed as a list instead of a comma separated string.

gaphor/diagram/general/generalpropertypages.py
sz332: Yeah, we are doing this as we discussed yesterday.

pyproject.toml
sz332: Moved upwards, thanks for the idea.

gaphor/diagram/general/picture.py
sz332: Replaced the docstring format.

gaphor/diagram/general/generalpropertypages.py
sz332: Tried the replace it with list but received a python error. Now trying it out with a simple space separated string, maybe it works.

poetry.lock
sz332: Thank you for noticing it, it seems that I really had an older version of poetry. Removed the old one, installed the new one, replaced they lock file from main, used poetry lock, now it shall work as expected.

gaphor/diagram/general/generalpropertypages.py
sz332: I am not a big fan of singletons, but it would be nice to be able to acquire a reference to the `Application` and then simply ask for the `main_window`.

danyeaw: @sz332 I pushed a fix for the filter not working on Windows by using the built-in [add_pixbuf_formats](https://docs.gtk.org/gtk4/method.FileFilter.add_pixbuf_formats.html) filter. Please let me know what you think.

@amolenaar, any ideas on how we can get the parent window here? I don't feel very comfortable with us merging this with the current GtkDialog warning.

sz332: I think it looks totally fine, especially that it is being supported "natively".

danyeaw (approved):

danyeaw:
Thanks @sz332!

Issue #2640, opened 2023-08-13
Show activities on bdd with block representation


marek-piirikivi:
### Is your feature request related to a problem?

SysML describes "Activities as blocks" which means that activities can have similar properties to the blocks and their structure can be shown.

"Activities as blocks can have associations between each other, including composition associations. Composition
means that destroying an instance at the whole end destroys instances at the part end. When composition is used
with activity blocks, the termination of execution of an activity on the whole end will terminate executions of
activities on the part end of the links." [SysML 1.6 spec]

![image](https://github.com/gaphor/gaphor/assets/45357878/846c6be2-1653-45ef-9f8b-22e140b57f8f)

### Describe the solution you'd like

I would like the activities to have block representation when they are placed on the block definition diagram.

### Additional context

This issue solves purely visible problem. Behind the scenes, forming composition associations is already possible between activities.

PR #2663, opened 2023-08-20
Represent activity, interaction and state machine as blocks on block definition diagram

marek-piirikivi:
#### Separate item only factory from item with element factory in toolbox and support specifying diagram specific items

In order to support creating diagram specific items for the element,
it is important that element representations are consistent between toolbox added items
and dragged element created items. For that, toolbox no longer is allowed to specify both
item and element - new_element_item() must be used instead.

represents() takes now additional parameter of the diagram class where the item can be represented on.

#### Add missing SysML diagrams in the sysml model

According to the SysML specification. Necessary to precisely select diagram types. Also, potentially, enables to support all kinds of diagram specific representations - such as adding frame with ports/parameters.

#### Use correct SysML diagram types

SysML toolbox diagram definitions should use correct SysML diagram types.

####  Add ActivityAsBlock, InteractionAsBlock, StateMachineAsBlock diagram items for SysML block definition diagram

SysML specification defines a construct called "<BehaviorElement> as block" which enables to represent behavioral element on block definition diagram to define behavioral compositions.

### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: #https://github.com/gaphor/gaphor/issues/2640

- [ ] Yes
- [x] No

It does not explicitly break anything, but the representations wont start working without re-creating the elements on the block definition diagram. Additionally, SysML diagrams will also be either Diagram or SysMLDiagram. It is rather easy to refactor these things, but I would prefer to make it explicit for the user.

### DIFF
--- a/gaphor/SysML/blocks/block.py
+++ b/gaphor/SysML/blocks/block.py
@@ -14,11 +14,12 @@ from gaphor.diagram.shapes import (
 )
 from gaphor.diagram.support import represents
 from gaphor.diagram.text import FontStyle, FontWeight
-from gaphor.SysML.sysml import Block, ValueType
+from gaphor.SysML.sysml import Block, ValueType, SysMLBlockDefinitionDiagram
 from gaphor.UML.classes.klass import attributes_compartment, operation_watches
 from gaphor.UML.classes.stereotype import stereotype_compartments, stereotype_watches
 from gaphor.UML.recipes import stereotypes_str
 from gaphor.UML.umlfmt import format_operation, format_property
+from gaphor.UML.uml import Activity, StateMachine, Interaction
 
 
 @represents(Block)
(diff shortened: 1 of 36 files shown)


marek-piirikivi:
closes https://github.com/gaphor/gaphor/issues/2640

gaphor/SysML/blocks/block.py
```diff
@@ -207,3 +208,58 @@ def lazy_format(operation):
             },
             draw=draw_top_separator,
         )
+
+
+class BehaviorAsBlockItem(Classified, ElementPresentation[Block]):
```
amolenaar, line 213: Should we introduce different elements for things represented as block items? We could also add an attribute to the default element and define a different shape.

gaphor/diagram/diagramtoolbox.py
```diff
@@ -91,33 +98,110 @@ def get_tool_def(modeling_language, tool_name):
...
+
+
+def new_deferred_element_item_factory(
+    subject_class: Type[Element], config_func: Optional[ConfigFuncType] = None
+):
+    """
+    Creates new item depending on the diagram, but does not create the element.
+    ``config_func`` may be a function accepting the newly created item.
+    """
+    return DeferredElementItemFactory(
```
amolenaar, line 194: I'm not sure if "deferred" is the right term. The factory function has no stake in creating the model element.

amolenaar (approved):
Nice addition. I like how the diagram item - model element mapping is much more centralized now.

I had a few remarks, but nothing that would require a code change per-se.

gaphor/SysML/blocks/block.py
marek-piirikivi: I suspect you do not mean "model element" here. Am I correct? Do you mean a base item which can be extended from to create items that are block shaped?

Please elaborate bit more :slightly_smiling_face:

gaphor/diagram/diagramtoolbox.py
marek-piirikivi: That's true :thinking: 

It is for connectable items which, upon connection create elements or connect existing ones. I'll try to come up with something slightly different.

gaphor/SysML/blocks/block.py
marek-piirikivi: I thought about this in the beginning actually.

My first concern was that ActivityItem is defined under UML for UML activities. And it makes sense that it is represented as it was and I would not want to couple UML with SysML in a sense that a part of UML has to be aware that "bdd" exists.

I really like how dispatch and registering is used to avoid throwing in if-else statements to customize the function behavior. I saw that we can definitely use similar approach to give activities their customizable representations without coupling them all together in a single place - especially since the mechanism is already set up.

Moreover, we would have to start thinking about "where do we draw a line?" because in that sense we may as well make a single Item() class and just add more and more if-else statements to represent things. This is bit over the top example, but I think it illustrates the point. 

Activity as a block is SysML construct. Actually, it is SysML diagram construct - diagram construct to be more specific. The items are kinda like what elements are to the model, but for the diagram. So, model consists of elements; diagram consists of items; model may contain diagrams, which are model elements. And depending on the diagram type, the set of items is technically defined. Registering them would make it more explicit and more easily definable on what diagrams can have what items. This information can later be used for validation and/or to guide proper SysML usage.

We want to be approachable from both UML modellers and SysML modellers. I fear that when UML modeller would see "Show as block" option and be confused because that's not UML specific thing. So your second suggestion would be more appropriate. As a SysML modeller, I would like to be able to represent my behavior architectures similarly how I represent other architectures on the bdd; and feel familiarity with other tools. I would expect the default to be "ActivityAsBlock" instead of having to select "Show as a block".

Putting it all together, I understand now your suggestion and I thought about it, but I had the concerns/reasoning that I explained above. In short, I feel as if your suggestion is more as a "hack" to increase productivity. It definitely would meet my needs and enable me to show activities as a block on bdd, so it is not bad suggestion. If we look more into the future; I think it wont hurt to be able to define diagram explicit items; and leave the door to query a list of elements and their possible representations on the different kinds of diagrams. :slightly_smiling_face:

amolenaar: You're right. I did not consider Activity(Item) to be a UML construct.

We need to have a conditional somewhere that determines the shape, and the dispatch functionality can do this job.

Thanks for clarifying. Let's do it as you propose.

amolenaar (approved):

marek-piirikivi:
I'll close this MR. I think item's option "Display as a block" would be better suited - in that way, modeler can decide on its own.

PR #2665, opened 2023-08-21
Support state entry, exit, do behavior selection amongst the existing behaviors via dropdown

marek-piirikivi:
### PR Checklist

### PR Type

Adds drop-down selection to select entry, exit, do behaviors amongst the existing behaviors.

- [ ] Bug fix
- [x] Feature

Currently, a textfield is used to define entry, exit, do behaviors which always create new activities. It is not possible to select existing activities.

- [ ] Yes
- [x] No

### Other information

Ideally, I would like to preserve the option to create the behavior when it does not exist on the property page; or defer its creation at the later point and just keep it as a text.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------
 - Fix missing derive relationship icon in the model browser
 - Fix block not showing parts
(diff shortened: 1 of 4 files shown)


amolenaar (approved):
This works really nice! Thanks @marek-piirikivi

Issue #2115, opened 2023-03-02
Ability to toggle dark mode off without fiddling with code


YosypC:
Disclaimer: total noob here.

Without the " background-color: rgb(255, 255, 255); " line in the Style Sheet, I see no way of reverting back to light mode. I think that some users would like light mode by default, so a convenient command to switch would be nice to have.

amolenaar:
Dark and light mode depend on your OS settings. On linux (GNOME) they can be toggled, on Windows and macOS the mode is set at application startup. This works fine with me if I clean the style sheet.

You want to switch to dark/light mode independent on the OS settings?

YosypC:
> You want to switch to dark/light mode independent on the OS settings?

Correct

amolenaar:
@YosypC Which OS are you on?

It's doable, there are some other apps allowing you to set the color scheme independent of the system scheme.

Would you expect the whole UI to change, or just the diagrams?

YosypC:
> @YosypC Which OS are you on?
> 
Windows 10
> It's doable, there are some other apps allowing you to set the color scheme independent of the system scheme.
> 
Thanks for the workaround, but an external app shouldn't be required to change a setting of a program, any software has both auto and selectable option.
> Would you expect the whole UI to change, or just the diagrams?
>
The diagram or both (more choices are better), but wasn't the UI already black before?

amolenaar:
> Thanks for the workaround, but an external app shouldn't be required to change a setting of a program, any software has both auto and selectable option.

No, I mean there are code examples available on how to make it work :wink:.

> The diagram or both (more choices are better), but wasn't the UI already black before?

The UI is only black if you use dark-mode on your OS.

YosypC:
> No, I mean there are code examples available on how to make it work 😉.
Aaaahh that's clearer, now. I wish I had the skills and time to implement them :(

> The UI is only black if you use dark-mode on your OS.
I understand. Thank you.

danyeaw:
I think apps should respect system-wide settings by default, which is what we are currently doing. If on Windows you go to Settings > Personalization > Colors and select Dark, you get Dark UI and diagrams. If you set it to Light, you get light UI and diagrams.

While adding an option to override the system-wide settings in possible, I think that it is a lower priority feature. If someone really wants it and wants to submit a PR, I'm all for it though. :smile: 

What @amolenaar is talking about is changing the CSS so that your diagrams are always black (if that's what you want):

```
diagram {
  background-color: #343131;
[...]

PR #2320, opened 2023-05-22
Replace deprecated Gtk.FileChooser by FileDialog

amolenaar:
### PR Type
What kind of change does this PR introduce?

- [ ] Bug fix
- [ ] Feature
- [x] Chore (refactoring, formatting, local variables, other cleanup)
- [ ] Documentation content changes

### What is the current behavior?

Issue Number: #2231 

### What is the new behavior?

Use the modern `FileDialog` API.

### Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

### Other information

Not sure if it's still using the native dialogs on Windows and macOS.

`Gtk.FileDialog` is introduced in GTK 4.10. ~We're still building on 4.6 (Ubuntu 22.04). 
I think we can wait until we have GTK 4.10+ available on our build machines.~

PR #2756, opened 2023-10-08
Add App Preferences for Dark Mode and Locale Override

### LINKED ISSUE #2586: Ability to set the UI language of Gaphor

Gaphor uses the operating system's default language however very often it cannot be changed to english due company policy however I would like to use it in english.

The current solution is to set some kind of environment variable and start Gaphor but I always forget what it is, so it would be nice to have it configurable on the user interface somewhere.

This might be configurable in the initial dialog.

danyeaw:
This PR creates a new Preferences Window to allow users to override the system dark mode settings and also override the language to English.

Fixes #2586 and https://github.com/gaphor/gaphor/issues/2115

- [X] Preferences window with dark mode override
- [x] Save and load Gio.Settings
- [x] Solution for overriding locale
- [X] Refactor settings in to its own module
- [x] Add tests

image
image

### PR Checklist

### PR Type

- [ ] Bug fix
- [X] Feature

Gaphor doesn't support any app wide preferences

Issue Number: N/A

### What is the new behavior?
Gaphor has a minimum amount of preferences

- [ ] Yes
- [X] No


danyeaw:
@amolenaar I could use some architecture advice, should Settings be a separate service? It started out as something that loaded with the help module, but now we need to check the setting while loading i18n to see if we have Use English set.

amolenaar:
> @amolenaar I could use some architecture advice, should Settings be a separate service? It started out as something that loaded with the help module, but now we need to check the setting while loading i18n to see if we have Use English set.

That is a very good question :smile:. Let's see:

* i18n current does not work as a service.
* Changing will require a restart, or at least a reload of the application, since the language in all UI elements needs to change.

Maybe start with a module that takes a global config from a `~/.config/gaphor/settings` file? The module provides a few functions that load the global settings. This should allow the i18n module to obtain the preferred language.

It would for sure be nice to have it as an (application) service, but then the i18n code needs to be reworked too.

UI-wise I would drop the sub-text ("Change the color scheme..."), and instead make the text "Override Dark Mode" more descriptive.

danyeaw:
@amolenaar Thanks!

I made the changes to UI to remove the subtitles and try to make the descriptions more clear:
image

I created a new settings module. Gio.Settings takes care of saving and loading settings based on the schema, so I think creating a separate config file may not be needed. Do you think sharing the Settings class instance with i18n and help is OK?

danyeaw:
Should I look at putting an exception in for our pytest-archon rules?

danyeaw:
Never mind, I added settings as a Gaphor core module which should make archon happy.

A couple of other updates needed: I think we need an Adw Banner to popup when the language settings are changed:
image

I had a run at this and couldn't get it to work with the Adw Preferences Page, but I'll have another go at it.

The nice Adw SwitchRow looks like it is only supported with libadwaita 1.4. We may need a fallback using a Gtk.Switch if isn't available or don't use it at all. It is really cool though :smile:

amolenaar:
> The nice Adw SwitchRow looks like it is only supported with libadwaita 1.4. We may need a fallback using a Gtk.Switch if isn't available or don't use it at all. It is really cool though 😄

The reason we are on an older GTK/libadwaita version is mainly due to our build server (Ubuntu), which is stuck on an older version. We have #2320 waiting to be merged for the same reason. I suggest we make the build work with the latest GTK/libadwaita version

* Maybe we can find a PPA with up-to-date versions of GTK?
* We can also use a container image (Fedora?) an build there. 
* We can exclude tests based on the GTK/libadwaita version used. This may be a good idea anyway. They're tested on Windows and macOS.
* We can even compile GTK/libadwaita from source if we have to. I think there's even more things we can do,

danyeaw:
Hi @amolenaar, ya, I like your idea to either find a PPA with updated versions of GTK or use a container image. Let me start to investigate this!

danyeaw:
Linux is working great now. Neither Windows or macOS have libadwaita 1.4 yet due to AppStream compatibility. I'll work on some yak shaving to see if I can't get that going.

danyeaw:
I noticed that we are getting a Cairo error in Windows from the last Gvsbuild update, let's hold off merging this until I investigate

.github/workflows/full-build.yml
```diff
@@ -143,7 +143,7 @@ jobs:
   macos:
     name: macOS
     needs: lint
-    runs-on: macos-11
+    runs-on: macos-latest
```
amolenaar, line 146: What's the reason we cannot stay on macos-11? I'd rather have a fixed version anyway, so we can communicate that on our website.

gaphor/settings.py
```diff
@@ -0,0 +1,34 @@
+"""Application settings support for Gaphor."""
+
+from gi.repository import Gio
+import logging
+
+APPLICATION_ID = "org.gaphor.Gaphor"
```
amolenaar, line 6: This variable is also defined in `gaphor/ui/__init__.py`. There we should import this value instead.

gaphor/settings.py
```diff
@@ -0,0 +1,34 @@
...
+        Gio.Settings.__init__(self)
+
+    @classmethod
+    def new(cls) -> Gio.Settings | None:
+        """Create a new Settings object."""
+        schema_source = Gio.SettingsSchemaSource.get_default()
+        if (
+            Gio.SettingsSchemaSource.lookup(schema_source, APPLICATION_ID, False)
+            is not None
+        ):
```
amolenaar, line 24:
```suggestion
        if (
            Gio.SettingsSchemaSource.lookup(schema_source, APPLICATION_ID, False)
        ):
```

gaphor/settings.py
```diff
@@ -0,0 +1,34 @@
...
+            gio_settings = Gio.Settings.new(APPLICATION_ID)
+            gio_settings.__class__ = Settings
+            return gio_settings
+        logger.info(
+            "Settings schema not found and settings won't be saved, run `poe install-schemas`"
```
amolenaar, line 29: This comment is only applicable for development environments. I suppose it works well for our packaged applications.

Should we have a self-test that verifies that gaphor-specific settings are properly loaded?

gaphor/settings.py
```diff
@@ -0,0 +1,34 @@
...
+APPLICATION_ID = "org.gaphor.Gaphor"
+
+logger = logging.getLogger(__name__)
+
+
+class Settings(Gio.Settings):
+    """Gaphor settings."""
+
+    def __init__(self):
+        Gio.Settings.__init__(self)
```
amolenaar, line 15: We do not need this constructor, right?

amolenaar (approved):
Looks good. I think there are a few minor improvements to be made.

It may be worth to see if the settings module code can be moved into the UI module - so that settings made in the (ui) application are not applied when using (for example) a jupyter notebook.

.github/workflows/full-build.yml
danyeaw: Hey @amolenaar, macos-12 has been available as a runner for over a year and homebrew is printing lots of warnings:

```
Warning: You are using macOS 11.
We (and Apple) do not provide support for this old version.
It is expected behaviour that some formulae will fail to build in this old version.
It is expected behaviour that Homebrew will be buggy and slow.
Do not create any issues about this on Homebrew's GitHub repositories.
Do not create any issues even if you think this message is unrelated.
Any opened issues will be immediately closed without response.
Do not ask for help from Homebrew or its maintainers on social media.
You may ask for help in Homebrew's discussions but are unlikely to receive a response.
Try to figure out the problem yourself and submit a fix as a pull request.
We will review it but may or may not accept it.
```

I was having issues with homebrew pulling in old versions. I think the brew update was what actually resolved this, but I think we should probably upgrade anyway. How about I pin macos-12?

gaphor/settings.py
danyeaw: Hi @amolenaar, this creates a circular import that's why I redefined it:

gaphor.ui -> gaphor.application -> gaphor.core -> gaphor.i18n -> gaphor.settings -> gaphor.ui

gaphor/settings.py
danyeaw: Thanks, simplified!

gaphor/settings.py
danyeaw: @amolenaar Thanks, good one.

gaphor/settings.py
danyeaw: Great idea, I added a self test!

danyeaw:
It looks like gsettings schema doesn't default searching in the users .local directory on Windows. We could override it there using the GSETTINGS_SCHEMA_DIR environmental variable. We could also copy it to the Gvsbuild share directory which it would search by default.

gaphor/settings.py
amolenaar: And if you define `APPLICATION_ID` here and import it in `gaphor.ui`?

amolenaar (approved):
LGTM

gaphor/settings.py
danyeaw: I swapped this, good idea!

PR #2775, opened 2023-10-14
Build on CI with a Fedora 39 image

danyeaw:
This PR changes from using the default Ubuntu runner on GitHub Actions to using Fedora 39. :star_struck:  This allows us to have more recently released versions of GTK and related dependencies to test against, and should unblock #2756 and #2320.

Additionally, I dropped the Python 3.10 build since Python 3.12 is now released.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Linux tests are run on Ubuntu 22.10

Issue Number: N/A

### What is the new behavior?
Linux tests are run on Fedora 39

- [ ] Yes
- [X] No

### Other information
I initially tried to use a Ubuntu 23.10 image, but I ran in to a lot of problems installing the different versions of Python using the deadsnakes PPA. Ubuntu disables ensurepip and pip wasn't working outside a virtualenv. Fedora was much smoother to get going.

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -50,44 +50,45 @@ jobs:
     name: Linux (Wheel) with Python ${{ matrix.python_version }}
     needs: lint
     runs-on: ubuntu-22.04
+    container: fedora:39
     timeout-minutes: 30
     if: "!contains(github.event.head_commit.message, 'skip ci')"
     strategy:
       matrix:
-        python_version: ['3.10', '3.11', '3.12']
+        python_version: ['3.11', '3.12']
     outputs:
       targz: gaphor-${{ steps.install.outputs.version }}.tar.gz
       wheel: gaphor-${{ steps.install.outputs.version }}-py3-none-any.whl
       version: ${{ steps.install.outputs.version }}
     steps:
-      - name: Harden Runner
-        uses: step-security/harden-runner@1b05615854632b887b69ae1be8cbefe72d3ae423 # v2.6.0
-        with:
-          egress-policy: audit
-
+      - name: Install Linux Dependencies
+        run: >
+          dnf install -y gcc git gobject-introspection-devel
+          cairo-gobject-devel graphviz pkg-config gtk4 gtksourceview5-devel
+          libadwaita-devel cairo-devel upx python${{ matrix.python_version }}
+          python${{ matrix.python_version }}-devel python-launcher
+          xorg-x11-server-Xvfb
+      - name: Install pip
+        run: python${{ matrix.python_version }} -m ensurepip
+      - name: Install pipx
+        run: |
+          py -${{ matrix.python_version }} -m pip install pipx
+          echo "/github/home/.local/bin" >> $GITHUB_PATH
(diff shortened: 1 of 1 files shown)


.github/workflows/full-build.yml
```diff
@@ -50,44 +50,45 @@ jobs:
...
       - name: Install Linux Dependencies
         run: >
-          sudo apt-get update -qq && sudo apt-get install -qq --no-install-recommends upx
-          gir1.2-gtk-4.0 libgirepository1.0-dev libgtksourceview-5-dev libadwaita-1-dev
-          graphviz
-      - name: Set up Python
-        uses: actions/setup-python@65d7f2d534ac1bc67fcd62888c5f4f3d2cb2b236 # v4.7.1
+          dnf install -y gcc git gobject-introspection-devel
+          cairo-gobject-devel graphviz pkg-config gtk4 gtksourceview5-devel
+          libadwaita-devel cairo-devel upx python${{ matrix.python_version }}
```
amolenaar, line 68: Is `python${{ matrix.python_version }}` required if `python${{ matrix.python_version }}-devel` is already installed?

.github/workflows/full-build.yml
```diff
@@ -50,44 +50,45 @@ jobs:
...
+          python${{ matrix.python_version }}-devel python-launcher
+          xorg-x11-server-Xvfb
+      - name: Install pip
+        run: python${{ matrix.python_version }} -m ensurepip
```
amolenaar, line 72: I would not make this a separate step, but include it with either the previous or next step. Having it as a separate doesn't add anything imho.

.github/workflows/full-build.yml
```diff
@@ -50,44 +50,45 @@ jobs:
...
+        run: |
+          py -${{ matrix.python_version }} -m pip install pipx
+          echo "/github/home/.local/bin" >> $GITHUB_PATH
+      - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
         with:
-          python-version: ${{ matrix.python_version }}
-          allow-prereleases: true
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Set ownership of Checkout directory
+        run: chown -R $(id -u):$(id -g) $PWD
```
amolenaar, line 81: Same here. I'd probably add it to the `Install Linux Dependencies` step.

amolenaar (approved):
Quite a straighforward change. Nicely done :).

I'd reduce the number of steps though, so we only see the relevant (higher level) ones, such as install and testing.

.github/workflows/full-build.yml
danyeaw: Yup, great feedback, thanks!

.github/workflows/full-build.yml
danyeaw: Thanks, yup, this simplifies the dependencies a bit.

.github/workflows/full-build.yml
danyeaw: In the Linux dependencies step we need to install Git for the checkout dependency to work properly, and then finally set the correct permissions. So we need these 3 steps separately to keep this order.

Issue #2612, opened 2023-08-08
Defined ports not adding to parts


mikekidner:
### Describe the bug
If a port has already been manually added to a block or part, when the block is on a diagram double-clicking the port in the model browser displays the port on the block. However, a predefined port cannot be added to the part by the same method. A new port has to be created.

### To Reproduce
1. Create a block
2. Use the corresponding part on an ibd
3. Add a port to the part
4.Go to the bdd, and double click the port in the model browser -> the port adds to the block.
5. Now try the above, but creating the port on the block first and then double clicking to automatically add to the part in the ibd.
6. It does not.

If possible, add the error report from `≡` (top-right menu) → `Tools` → `Error Reports`.

### Expected behavior
If a port has been defined on a block or part, double clicking the port property in the model browser should add the port to the block or part on the diagram not just the block.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [x] macOS
- [ ] Windows

### Version
Version of Gaphor:2.20

danyeaw:
Thanks @mikekidner. This sounds like a good idea, I would expect this is a new feature since probably no one tested this use case.

marek-piirikivi:
If possible, would @mikekidner, be so kind and rephrase the expected behavior? I have trouble understanding what should happen. :slightly_smiling_face: 

Currently, I can create the block. Add port to it. Then I can reuse the same block in another diagram. When I double-click on the port, it is correctly added to the block as I would expect.

mikekidner:
Yes, no worries.

Create a parent (B0) and child (B1) block, 
Add a port to B1. 
Now use the B1-Part on the ibd of B0, try to add a port to the B1-Part by double-clicking in the model browser.

marek-piirikivi:
![image](https://github.com/gaphor/gaphor/assets/45357878/35d2ed75-a0ad-4cf3-a327-e67ded173496)
![image](https://github.com/gaphor/gaphor/assets/45357878/f63992cf-dd80-46bf-9c7e-e19e08ad35b3)

@mikekidner I followed your instructions. See the screenshots whether it looks as if I did. It behaved as I expected. I have trouble understanding your expectations. Especially "double clicking the port property in the model browser should add the port to the block or **part on the diagram** not just the block." It did add it to the port. What is meant by the "part on the diagram"?

mikekidner:
Hi @marek-piirikivi, by "part" I mean the property B1 owned by B0.
Here is a bdd for block B0, 
[...]

PR #2842, opened 2023-11-15
Proxy Port Improvements

marek-piirikivi:
I added additional check to see if there exists an item on the diagram that may own the proxy port.

### PR Checklist

### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: #2612

- [ ] Yes
- [x] No

### Other information

Additionally fixed proxy port drag icon. Automatically generated icon name is "proxy-port".

### DIFF
--- a/gaphor/SysML/drop.py
+++ b/gaphor/SysML/drop.py
@@ -2,7 +2,8 @@ from gaphor.diagram.drop import drop
 from gaphor.diagram.presentation import connect
 from gaphor.diagram.support import get_diagram_item
 from gaphor.SysML.sysml import ProxyPort
-from gaphor.UML.drop import diagram_has_presentation
+from gaphor.SysML.sysml import Block
+from gaphor.UML.uml import Property
 
 
 @drop.register
@@ -11,7 +12,28 @@ def drop_proxy_port(element: ProxyPort, diagram, x, y):
     if not item_class:
         return None
 
-    head_item = diagram_has_presentation(diagram, element.encapsulatedClassifier)
+    def port_owning_item_block(item):
+        match item.subject:
+            case Property():
+                block = (
+                    item.subject.type if isinstance(item.subject.type, Block) else None
+                )
+            case Block():
+                block = item.subject
+            case _:
+                return False
+        return element in block.ownedPort if block else False
+
+    def drop_distance_to_item(item):
+        local_x, local_y = item.matrix_i2c.inverse().transform_point(x, y)
+        return item.point(local_x, local_y)
+
+    head_item = min(
+        diagram.select(lambda item: port_owning_item_block(item)),
+        key=drop_distance_to_item,
+        default=None,
+    )
+
     if not head_item:
         return None
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 2.21.1
 ------
-(Unreleased)
+- Add proxy port to the closest port owning item
+- Allow adding existing proxy port on the owner block property
+- Fix proxy port drag icon
 
 2.21.0
 ------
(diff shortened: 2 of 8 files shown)


gaphor/SysML/drop.py
```diff
@@ -11,7 +13,23 @@ def drop_proxy_port(element: ProxyPort, diagram, x, y):
...
+            isinstance(item, PropertyItem)
+            and isinstance(item.subject.type, Block)
+            and element in item.subject.type.ownedPort
+        )
+
+    head_item = next(
+        (
+            item
+            for item in diagram.select(
+                property_item_which_is_typed_by_block_owning_the_port
```
amolenaar, line 27: I would inline the code.

```suggestion
                lambda e: isinstance(e, PropertyItem)
                and isinstance(e.subject.type, Block)
                and element in e.subject.type.ownedPort
```

danyeaw:
Hi @marek-piirikivi, I tried to test this by doing the following:

1. Create a block called Car
2. Create a 2nd block named Wheel
3. Add a composition relationship from the Car to the Wheel
4. Add the Wheel Part Property to the diagram
5. Add a proxy port to the Wheel Part Property
6. Add the Wheel Part Property to the diagram again
7. Try to drag the proxy port on to the 2nd Wheel Part Property

However, this doesn't seem to work for me. Am I testing this correctly or was it fixing some other use case?

marek-piirikivi:
@danyeaw Could you tell me what symptoms there were? What happened? What you expected to happen?

I'll check this out myself, but the fix is about reusing already existing elements from the model browser. It does not fix the problem that we have before about the problem with placing ports when there are multiple items that can hold the port on the diagram. It is known older problem that when there are multiple items which the port can belong to, that only the first found item gets the port - is this what you experienced? I can solve this within this merge request as well, but I see this as a separate problem. This change merely "adds block's property into the list that even can own a port that's dragged from the model btowser". :slightly_smiling_face:

danyeaw:
@marek-piirikivi Sorry for the confusion, it sounds like I was testing out a different use case, definitely don't expand this PR to fix that other thing. I reread the issue you linked to. With this PR, if I try this:

1. Create a block
2. Use the corresponding part on an ibd
3. Add a port to the part
4. Go to the bdd, and double click the port in the model browser

The port is not adding to the part again instead of the block.

marek-piirikivi:
@danyeaw Are you sure you are looking at the correct changes? Here is a video - I tried to reproduce the issues, but maybe I am not understanding the use-case properly :thinking: 

https://github.com/gaphor/gaphor/assets/45357878/86d4414a-69c0-4d58-9b9f-28becfcd4978

marek-piirikivi:
> definitely don't expand this PR to fix that other thing

:smiling_imp: I did that. It was bugging me so much... even if I do not see the use-case for having the two items representing the same block on the same diagram.

Removing unused elements with proxy port is still broken. When I delete it from any of the items referencing it, it removes it from everywhere.

marek-piirikivi:
... I should add tests as well... :sweat_smile:

marek-piirikivi:
@danyeaw @amolenaar 

I made some updates:
1. You can now drag existing ports onto all the items that may own the ports - either the owning block or a property typed by owning block. The closest item will get the block.
2. Deleting a port item will not delete the port from the model if there are any port items presenting the port. I tested manually, seems to be working with both "Remove unused elements" enabled and disabled as expected. Now, making a mistake while dragging a duplicate port, you can remove it without having to worry about it being removed from the model.
3. I added tests for the first one. Second one had already test coverage and it did not brake.

From my point of view, I tried to brake the functionality, but couldn't - it behaves as I expect it to behave. Please review. :slightly_smiling_face: 

Here is a small demo as well (ignore jerky mouse movement - I am not very good with mouse, shaky hands)
https://github.com/gaphor/gaphor/assets/45357878/a77847ab-1f4a-4c91-acca-0763dcd9d6bf

danyeaw (approved):
Thanks @marek-piirikivi!

PR #2850, opened 2023-11-17
Arch rules & build update

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

### What is the new behavior?

- Architecture test rules have been fine tuned. `gaphor.diagram` is still a bit problematic since it has a (cyclic) dependency on `gaphor.ui` and `gaphor.UML`, but the dependencies have been explicitly specified now.
- Build artifacts also for PR's created by Dependabot. We should be able to assess updates on libraries like PyInstaller.

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -183,7 +183,6 @@ jobs:
         run: poetry run pytest --cov
       - name: Create macOS Application
         id: create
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
Looks good, thanks @amolenaar!

PR #2858, opened 2023-11-21
Reduce duplicate GitHub Action runs

danyeaw:
Currently, when pushing a branch and opening a PR, GitHub Actions runs two complete builds. This PR tries to reduce duplicate runs by only running the full build when:

1. A PR is opened against the main branch
2. Push to the main branch
3. On release
4. On manually trigger a build with workflow dispatch from the actions menu

The tradeoff of this will be that if you want to run GitHub Actions, you'll need to open a draft PR or manually start it from the Actions tab, since pushing to a branch will not start a run.

I also added logic to cancel running previous workflows using concurrency.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

2 full builds for each push of a PR

Issue Number: N/A

### What is the new behavior?
Avoid duplicate builds under most circumstances

- [ ] Yes
- [X] No

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -2,7 +2,13 @@ name: Full Build
 
 on:
   push:
+    branches:
+      - main
+    paths-ignore:
+      - '*.md'
   pull_request:
+    branches:
+      - main
   release:
     types: [published]
 
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -13,6 +13,10 @@ on:
     types: [published]
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   python_version: '3.11'
   mainline_build: ${{ github.ref == 'refs/heads/main' || github.event.label.name == 'translation' || github.event_name == 'release' }}
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -11,6 +11,7 @@ on:
       - main
   release:
     types: [published]
+  workflow_dispatch:
 
 env:
   python_version: '3.11'


amolenaar:
I like the ability to run a pipeline before I create a PR. But I agree that building for both pull and PR is overhead we can do without.

How does it work for forks? Should they open a PR in order to build something?

danyeaw:
> I like the ability to run a pipeline before I create a PR. But I agree that building for both pull and PR is overhead we can do without.

I think the workflow dispatch would help fill in this use case by [manually running a workflow](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow) by selecting the branch you push.

> How does it work for forks? Should they open a PR in order to build something?

Yup, create a PR.

.github/workflows/full-build.yml
```diff
@@ -2,9 +2,20 @@ name: Full Build
...
+    branches:
+      - main
+    paths-ignore:
+      - '*.md'
   pull_request:
+    branches:
+      - main
   release:
     types: [published]
+  workflow_dispatch:
```
marek-piirikivi, line 14:
```suggestion
  workflow_dispatch
```
It's my first time seeing github workflows, so I did some reading. I think `:` is not necessary since you are not providing any additional inputs.

amolenaar:
> > I like the ability to run a pipeline before I create a PR. But I agree that building for both pull and PR is overhead we can do without.
> 
> I think the workflow dispatch would help fill in this use case by [manually running a workflow](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow) by selecting the branch you push.

Fair enough. That's an option indeed.

> > How does it work for forks? Should they open a PR in order to build something?
> 
> Yup, create a PR.

This implies they cannot check if all builds, unless a build id triggered manually. Which is fine, I suppose.

amolenaar (approved):

.github/workflows/full-build.yml
danyeaw: Hi @marek-piirikivi, this didn't pass our yaml pre-commit check, so I had to undo it.

Issue #2213, opened 2023-04-07
Support Ports on Boundary of IBD


danyeaw:
### Is your feature request related to a problem?
This was originally reported by @marek-piirikivi on Matrix.

Diagrams that provide internal views of objects should have a border around them. While developing an IBD, I would like the ability to add ports on the outside border of the diagram to show interfaces with the Block that the IBD represents.

### Describe the solution you'd like
Render a border around the outside of a diagram and allow ports to be placed on it.

### Describe alternatives you've considered
@amolenaar mentioned on Matrix that we could remove the whole IBD and then have users draw internal to a block using a Block itself. 

### Additional context

![image](https://user-images.githubusercontent.com/10014976/230646774-a2881388-7acb-43e6-84c8-721c1eebea1e.png)

amolenaar:
Do you envision something similar to Parameters in an Activity? Or something like the ProxyPorts we already have?

Parameters are defined in the property editor and stick to the border of the Activity.

amolenaar:
Should we also change Block, so it looks like a diagram (type and name in top-left corner), when it contains nested items?

danyeaw:
> Do you envision something similar to Parameters in an Activity?
> Should we also change Block, so it looks like a diagram (type and name in top-left corner), when it contains nested items?

Something like this might be a stepping stone. However, the outside of an Activity diagram should already be the border of an activity, there shouldn't be two separate constructs. Similarly, we shouldn't have to add a block to an IBD as a separate object.

amolenaar:
> Something like this might be a stepping stone. However, the outside of an Activity diagram should already be the border of an activity, there shouldn't be two separate constructs. Similarly, we shouldn't have to add a block to an IBD as a separate object.

I think currently the concepts are a bit interwoven, since we have the different diagram types. Yet, a diagram is something different from a modeling element.

For block diagrams: if we allow a "diagram" layout for blocks, we can basically drop the _ibd_ from the diagram from the list of diagram types. A _ibd_ represents a block and shows its internals, right?

In case f the espresso machine, it could look like this:

![image](https://github.com/gaphor/gaphor/assets/96249/a646a0e0-6fc0-448e-932c-2620fcd5c32c)

mikekidner:
sysML does define a "structure" compartment of a block which is the IBD. Going down this route though would still require a diagram to show the internal structure and another to show the definition, as they'd likely not fit on one diagram. Users will probably expect IBD diagrams.
[...]

PR #2867, opened 2023-11-25
Add SysML diagram frame item

marek-piirikivi:
* Refactors activity item property page as just activity property page, because it did not have item-specific properties. Prerequisite for diagram frame property page.
* Ensures that all SysML diagrams have a frame presenting diagram's element.
* Diagram frame item's property page is the same as the underlying element's property page.
* Migrates older models to automatically attach the frame item on SysML diagrams; replaces a single diagram's element representation, if present, with diagram frame item.

It makes it more explicit what element the diagram represents; what are its interfaces, if any. Overall, the diagrams would look much clearer, conforms better with SysML specification, and allows to be future compatible and control better the way each diagram looks like and "behaves".

### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Currently, the diagram frame is attached to the diagram page. In order to get the ports/parameters of the diagram's element; you need to drag the element from the model browser onto the diagram and get the ports/parameters from there.

Issue Number: #2213 (partially)

### What is the new behavior?

See top description.

- [x] Yes
- [ ] No

Kinda. It modifies existing SysML diagrams to include diagram frame; and replaces diagram's element representation (only if one is found) with a frame item. So, existing diagrams are expected to change.

### Other information

This is prerequisite to complete parametric modelling. I did not want to do this without the proper frame.

It would support potential future features such as preventing adding activities onto activity diagram, instead, when dropping activity, we can automatically replace this with CallBehaviorAction; or when the activity was dropped on the Action, we can convert the Action into CallBehaviorAction → I see this as a great user experience improvement.

https://github.com/gaphor/gaphor/assets/45357878/7ecac5f8-9708-4a65-b460-5298e09c4f46

### DIFF
--- a/gaphor/SysML/blocks/connectors.py
+++ b/gaphor/SysML/blocks/connectors.py
@@ -9,31 +9,42 @@ from gaphor.SysML.blocks.block import BlockItem
 from gaphor.SysML.blocks.interfaceblock import InterfaceBlockItem
 from gaphor.SysML.blocks.property import PropertyItem
 from gaphor.SysML.blocks.proxyport import ProxyPortItem
+from gaphor.SysML.diagramtype import DiagramFrameItem
 from gaphor.UML.deployments import ConnectorItem
 
 
 @Connector.register(InterfaceBlockItem, ProxyPortItem)
 @Connector.register(BlockItem, ProxyPortItem)
 @Connector.register(PropertyItem, ProxyPortItem)
+@Connector.register(DiagramFrameItem, ProxyPortItem)
 class BlockProperyProxyPortConnector:
     def __init__(
         self,
-        block_or_property: Union[BlockItem, PropertyItem, InterfaceBlockItem],
+        element: Union[BlockItem, PropertyItem, DiagramFrameItem, InterfaceBlockItem],
         proxy_port: ProxyPortItem,
     ) -> None:
-        self.element = block_or_property
+        self.element = element
         self.proxy_port = proxy_port
 
     def allow(self, handle: Handle, port: Port) -> bool:
-        return (
-            bool(self.element.diagram)
-            and self.element.diagram is self.proxy_port.diagram
-            and (
+        # Early return when diagrams are not the same
+        if (
+            not bool(self.element.diagram)
+            or self.element.diagram is not self.proxy_port.diagram
+        ):
+            return False
+
+        if isinstance(self.element, DiagramFrameItem):
+            # Allow only on blocks
+            return isinstance(self.element.subject, sysml.Block) and isinstance(
+                self.element.subject, UML.EncapsulatedClassifier
+            )
+        else:
+            return (
                 isinstance(self.element.subject, UML.EncapsulatedClassifier)
                 or isinstance(self.element.subject, UML.Property)
                 and isinstance(self.element.subject.type, UML.EncapsulatedClassifier)
             )
-        )
 
     def connect(self, handle: Handle, port: Port) -> bool:
         """Connect and reconnect at model level.
--- a/gaphor/SysML/diagramframe.py
+++ b/gaphor/SysML/diagramframe.py
@@ -1,0 +1,155 @@
+from gaphas.geometry import Rectangle
+
+from gaphor.core.modeling import DrawContext, UpdateContext
+from gaphor.core.styling import (
+    FontWeight,
+    TextAlign,
+)
+from gaphor.diagram.diagramlabel import diagram_label
+from gaphor.diagram.presentation import (
+    Classified,
+    ElementPresentation,
+    connect,
+)
+from gaphor.diagram.shapes import (
+    cairo_state,
+)
+from gaphor.diagram.text import Layout
+from gaphor.UML.actions.activity import ActivityParameterNodeItem
+from gaphor.UML.uml import Activity, ActivityParameterNode
+
+
+class DiagramFrameItem(Classified, ElementPresentation):
+    def __init__(self, diagram, id=None):
+        super().__init__(diagram, id)
+
+        self.watch("diagram[Diagram].name", self.update_shapes).watch(
+            "diagram[Diagram].diagramType", self.update_shapes
+        ).watch("diagram[Diagram].element", self.update_element).watch(
+            "subject", self.update_element
+        ).watch("subject[Activity].node", self.update_frame_attachments).watch(
+            "diagram[Diagram].element[NamedElement].name", self.update_shapes
+        )
+
+    def update_element(self, event=None):
+        if not self.diagram or not self.diagram.element:
+            return
+
+        self.subject = self.diagram.element
+        self.update_shapes()
+        self.update_frame_attachments()
+
+    def postload(self):
+        super().postload()
+        self.update_frame_attachments()
+
+    def update_shapes(self, event=None):
+        if not self.diagram:
+            return
+
+        self.shape = DiagramFrameShape(
+            kind=self.diagram.diagramType,
+            label=diagram_label(self.diagram),
+        )
+
+    def update_frame_attachments(self, event=None):
+        if not self.subject:
+            return
+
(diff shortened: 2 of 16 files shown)


gaphor/diagram/tools/tests/test_shortcut.py
```diff
@@ -20,6 +20,7 @@ def test_delete_selected_items(create, diagram, event_manager):
     assert not diagram.ownedPresentation
 
 
+# TODO: Do we still want this feature?
```
marek-piirikivi, line 23: @amolenaar @danyeaw @sz332 
This test fails. I did not want to just remove it. I am not sure whether we want to have a feature where we delete diagram's element from the model when a diagram gets deleted.

If you have "Remove unused elements" feature enabled, it performs this cleanup anyways (I tested it after my change); but now we have an option to not delete it when the feature is disabled - which would be expected behavior in my opinion.

I would like to have a discussion on whether I should delete the test because I have removed the feature.

gaphor/storage/storage.py
```diff
@@ -153,8 +156,12 @@ def update_status_queue() -> Iterable[float]:
         if progress % 30 == 0:
             yield (progress * 100) / size
 
-    # First create elements and canvas items in the factory
+    # First ensure that all sysml diagrams have a frame
+    if version_lower_than(gaphor_version, (2, 22, 0)):
+        ensure_sysml_diagrams_have_a_frame(elements)
```
marek-piirikivi, line 161: I am not sure whether it is taking place at the correct step in model loading. The way I implemented it; it has to happen before we start loading elements and items.

Another option is to do it afterwards, but replacing an item is easier to be done by simply changing the type string rather than to rewire everything to a new item.

gaphor/storage/storage.py
```diff
@@ -462,3 +469,135 @@ def upgrade_note_on_model_element_only(
...
+                )
+                if item.references.get("subject", None) == el
+            ]
+
+            if len(element_items) == 1:
+                replace_diagram_element_item_with_frame(element_items[0])
+            else:
+                add_frame(diagram)
+        else:
+            add_frame(diagram)
```
marek-piirikivi, line 603: Are these acceptable defaults?

It ensures that there is always a frame which surrounds all the existing diagram items. If there is a single item which presents the same element as the diagram itself; we replace it with frame instead.

The latter may not be the safest thing to do, especially if we have a bdd representing a block; and we have the same block on the diagram which has the associations with other blocks; it may either: a) look like we form the associations with a diagram; b) the associations visualization may be broken. I suppose the option (a). But it is not desirable substitution in this case, so I am also considering just adding frame and surrounding all the items and call it a day and leave the cleanup for the modeler.

I tested this with my models; and on all of them; it was as I expected, but I can be considered rather strict in modeling and I do not use the block to define the block's internals as it would be kinda like defining systems engineering as "Systems Engineering is something that Systems Engineers do" where Systems Engineer is defined as "Someone who does Systems Engineering".

Again, I am interested in having a discussions.

sz332:
I actually have a question. Do we want to have some constraints for the diagram, so that elements can be added only inside the border? Or if they added outside, then the border auto-resizes itself in order to contain the elements?

Also, how should the border resize if it gets smaller than the area occupied by the existing elements? Shall we enable resize for those situations?

gaphor/SysML/blocks/connectors.py
```diff
@@ -9,31 +9,45 @@
...
-        )
+        # Early return when diagrams are not the same
+        if (
+            not bool(self.element.diagram)
+            or self.element.diagram is not self.proxy_port.diagram
+        ):
+            return False
+
+        match self.element:
+            case DiagramFrameItem():
```
amolenaar, line 38: I would use an `if isinstance(self.element, DiagramFrameItem)` instead.

gaphor/diagram/diagramlabel.py
```diff
@@ -4,3 +4,8 @@
 @singledispatch
 def diagram_label(diagram):
     return diagram.name
+
+
+@singledispatch
+def paint_diagram_type(diagram) -> bool:
+    return True
```
amolenaar, line 11:
```suggestion
    return not diagram.diagramType
```

gaphor/SysML/diagramframe.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+
+    def update_frame_attachments(self, event=None):
+        if not self.subject:
+            return
+
+        if isinstance(self.subject, Activity):
+            self.update_activity_parameters()
+
+    def update_activity_parameters(self):
```
amolenaar, line 63: Now it's only activity diagrams. I guess you have an idea of how to extend this to other diagram types?

gaphor/storage/storage.py
amolenaar: I would do it as early as possible.

What would happen if we leave out this upgrade? Is it required to do this upgrade or existing models just remain as is?

marek-piirikivi: Since it is not possible to add diagram frame item explicitly later, existing SysML diagrams need to get the item at minimal. I think I'll go with just adding frame item on SysML diagrams to surround all the elements there and leave the other refactoring for the users. Seems to be the safest option. Then you can choose when to start utilizing the frame for parameters/ports. I suppose we could leave a guide or add blog entry about this as well on how to utilize SysML frame.

Another alternative is to perform no upgrade, but in order to get diagram frame, you need to create brand new diagram.

gaphor/diagram/deletable.py
```diff
@@ -1,9 +1,14 @@
...
 
 @singledispatch
 def deletable(element: Element) -> bool:
     """Determine if a single element can safely be deleted."""
     return True
+
+
+@singledispatch
+def item_explicitly_deletable(item: Presentation):
+    return True
```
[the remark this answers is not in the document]
marek-piirikivi: I'll look whether I can use deletable(). Seem to be the same thing, somehow I did not consider this. :slightly_smiling_face:

marek-piirikivi:
> I actually have a question. Do we want to have some constraints for the diagram, so that elements can be added only inside the border? Or if they added outside, then the border auto-resizes itself in order to contain the elements?
> 
> Also, how should the border resize if it gets smaller than the area occupied by the existing elements? Shall we enable resize for those situations?

@sz332 Definitely would be a fine addition. I would prefer if the frame would surround its internal element's presentations then. It is fine to add metadata and comments and other drawings outside of the frame. :slightly_smiling_face: 

I also know how to do that, so the question is whether we want to do it or not. And whether we want to do it now; or take some time to think it through and add this later. Currently you have to drag the diagram smaller/larger. There is also an option to use a tool similar to alignment where you can explicitly select when you want to automatically fit the frame.

gaphor/diagram/deletable.py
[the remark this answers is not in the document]
marek-piirikivi: I substituted to use `deletable()`.

gaphor/diagram/diagramlabel.py
marek-piirikivi: discarded `not` because we are checking whether we should paint the diagram type.

gaphor/storage/storage.py
marek-piirikivi: I've decided to play it save an simply add a diagram frame item on all the sysml diagrams without any fancy magic. I leave it up to the modeler to decide whether they want to continue modeling the old ways; or they like to leverage the frame's ability to hold ports/parameters. I wont enforce this. :slightly_smiling_face:

danyeaw:
Hi @marek-piirikivi, are you still interested in finishing this up? It would be a really nice feature!

mikekidner:
@marek-piirikivi , this would be a nice feature and it could be extended to allow for alternate regions within sequence diagrams.
Screen Shot 2024-11-27 at 5 01 07 am

danyeaw:
Closed by #3689

PR #2879, opened 2023-11-29
Settings cleanup

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

If a preferred color scheme is selected, it's not set when you (re)start the application.

Issue Number: N/A

### What is the new behavior?

I made the Settings object a more rich object and moved many of the settings there. This makes the interface we have to use easier (no optional `None` object).

Changed the message that settings cannot be loaded to warning, so it's shown before the logging module is configured.

The application color scheme is updated according to the setting in a reactive fashion.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/i18n.py
+++ b/gaphor/i18n.py
@@ -57,7 +57,7 @@ def translation(lang) -> _gettext.GNUTranslations | _gettext.NullTranslations:
     return _gettext.NullTranslations()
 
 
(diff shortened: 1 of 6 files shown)


danyeaw (approved):
Hi @amolenaar, really nice refactoring! I'm always learning from you 🥇

amolenaar:
😊

PR #2918, opened 2023-12-09
Fix macOS build

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

The macOS build fails when installing dependencies with an error:

```
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.12
Target /usr/local/bin/2to3-3.12
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.12'
```

- [ ] Yes
- [x] No

### Other information

Pipx is said to be pre-installed on all runners. Although it doesn't seem to work on all.
It looks like the WIndows runner is dependent on python-launcher.

For macOS we do not need that per se. Hence I removed it and just changed the python version so it matches the old style (`python3.12`).

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -6,7 +6,7 @@ outputs:
     value: ${{ steps.meta.outputs.version }}
 inputs:
   python-version:
-    description: 'Python version number for Linux'
+    description: 'Python version. Can be a version python$V or $V depending on the platform.'
     required: false
 runs:
   using: composite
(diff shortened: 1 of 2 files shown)


danyeaw:
Here is the issue for this: https://github.com/actions/setup-python/issues/577

Should we do something hacky like:

```
brew update
brew install ... || true
brew link --overwrite python@3.12
```

danyeaw (approved):
Nice, that works!

amolenaar:
Yeah. It took quite some attempts, but I'm happy with the small change that's needed to make it work.

PR #2952, opened 2023-12-23
Less padding around windows on macOS

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

The Adwaita theme draws a wide shadow around the window. This shadow is part of the view on macOS.

You can see this easily when you open a file. The parent window is overlayed with a "sheet":

image

### What is the new behavior?

The shadow has been tweaked to look more like a macOS shadow, and the shadow size is smaller. This improves the usability: the big (invisible) border is gone, so if you click on an underlying window, it will actually work.

image

Also the shadow looks a bit more mac-like:

image

### Other information

GTK issue: https://gitlab.gnome.org/GNOME/gtk/-/issues/6255

### DIFF
--- a/gaphor/ui/styling-macos.css
+++ b/gaphor/ui/styling-macos.css
@@ -1,0 +1,21 @@
+/*
+ * Styling workarounds on macOS.
+ *
+ * The issue: shadows defined by libadwaita are very wide: up to 60px.
+ * The shadow is drawn inside the (clickable) window.
+ *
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
Thanks @amolenaar, nice improvement!

amolenaar:
I just did a check on Windows, and it suffers from the same problem. I think we can apply the same styling patch there as well.

PR #2955, opened 2023-12-24
Apply custom window shadow to macOS and Windows

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

As discussed in #2952, the shadow for macOS is very wide, and the shadow is drawn as part of the `NSView`, hence mouse clicks are consumed.

The same issue appears on Windows.

Issue Number: N/A

### What is the new behavior?

Update the window shadow style for both macOS and Windows.

The shadow style is now resembling the default shadow style in GTK.

I found that `CssProvider.load_from_path()` does not cascade, so only the shadow styling was loaded. This has been fixed.

### DIFF
--- a/gaphor/ui/styling-box-shadow.css
+++ b/gaphor/ui/styling-box-shadow.css
@@ -1,7 +1,7 @@
 /*
- * Styling workarounds on macOS.
+ * Styling workarounds on macOS and Windows.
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
Looks good!

PR #2957, opened 2023-12-24
Convert Parameter nodes view to use ColumnView

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Parameter Nodes view is still using ListView, while Attributes and Operations is using ColumnView.
This PR makes that consistent.

Issue Number: N/A

### What is the new behavior?

Same, but use ColumnView, and reuse the TextField column.


danyeaw (approved):
Thanks @amolenaar!

Issue #2482, opened 2023-07-15
Ability to set type and multiplicity of activity parameter


sz332:
### Is your feature request related to a problem?

When creating an activity it is important to be able to set the parameters. The parameter has a name, type, and direction (in,out,inout,return). While it is configurable in the activity from the usability point of view the best would be to be able
to configure those settings by clicking the parameter and setting the other settings as well.

### Describe the solution you'd like

- Add combobox with label Type (type selector functionality)
- Add combobox Direction (in, out, inout, return), based on the UML standard
- Add combobox Display (or similar label): Name, Type, Name and Type 
- When type and directions are set it should be visible on the diagram as well

Relevant UML specification parts:

![image](https://github.com/gaphor/gaphor/assets/8182138/4e2b48cc-3db9-414c-b6c6-22b0c208e23e)

![image](https://github.com/gaphor/gaphor/assets/8182138/9b085752-0008-4d39-9c44-feed33a9b5b5)

Mock screen: 

![image](https://github.com/gaphor/gaphor/assets/8182138/7d852628-1a4e-4109-9fb7-53a2e11ff34e)

## Additional hints

As more and more cases we might need the ability to set which attributes shall be visible, maybe this is a common feature and start thinking about that application wise (using DRY principle).

sz332:
We basically need the same thing that has already been implemented for pin. This might be a quick win.

danyeaw:
Completed by #2962. Thanks @sz332 for submitting the enhancement idea.

PR #2962, opened 2023-12-28
Add types for parameters

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Types cannot be defined on parameter nodes

Issue Number: #2482

### What is the new behavior?

Types are supported.

- [ ] Yes
- [x] No

### Other information

image

@sz332 Should the direction be visible in the diagram as well (similar to `Show Type`)?

Update: direction can be toggled.


sz332:
@amolenaar I think it would be nice to render a label (in, out, inout, return) before the name, but it could be made in a second PR as well.

amolenaar:
Added a "Show Direction" option. And fixed the type assignment: type should be assigned to the parameter.

danyeaw (approved):
Nice typed parameters FTW! :shipit:

sz332:
Awesome work!

amolenaar:
Thanks :blush:

Issue #2919, opened 2023-12-11
Add option to show abstract classifier by textual annotation {abstract}


nomisge:
### Describe the bug
UML 2.5.1 states in section 9.2.4.1
"The name of an abstract Classifier is shown in italics, where permitted by the font in use. Alternatively or in addition, an abstract Classifier may be shown using the textual annotation {abstract} after or below its name"

In Gaphor this is currently only manually possible for classes, and not at all for methods.

If fonts are used, that don't support italics or small print doesn't show nicely, than readers won't notice, that a class/method is supposed to be abstract.

### To Reproduce
1. Add a class with at least one method
2. Change the classes name and append {abstact} -> this works, but then the class name is changed.
3. Change the methods name and append {abstract} -> not working, since after confirming the change it is reverted.
4. Toggling the abstract options only has the 'italic effect'

### Expected / Suggested behavior
Add an option to settings, that when a class or method is declared abstract, the {abstract} annotation is automatically added according to the standard.
Maybe the option could offer three choices: only italic, only {abstract}, both

### Version
Any Version below 2.22

amolenaar:
Hi @nomisge, Thank you for your report.

I do not understand why you want this. Why is the italics notation not enough?

For classes it only works because you add it to the Class name. For methods and attributes, the text is parsed and the tags (`{abstract}`) is dismissed.

Isn't the solution to use a font that has a proper italics version available?

nomisge:
Thanks @amolenaar for looking at the report.

There are several reasons, the biggest beeing, that if the diagrams are printed on paper - no matter the font -  italics can always be overlooked. In my case this are exams I give to my students. And to add on that the state (in Germany) of the school I'm teaching at, does this with their final exams.
Another reason is, that other programs (used in education) already do this e.g. in JavaEditor https://javaeditor.org/.  This is for java only though.

Yes I know that it kind-a works by adding it to the name, but that is more like a bad workaround with caveats.

Since UML 2.5.1  describes an alternative to italics, I think this should be possible in gaphor. Thats a good reason too, isn't it?
There should be an option, that adds {abstract} to both class names and method names.

Should not be too hard to implement. If you give me a hint, which files need to be looked at, I would even try to have I go at it, although I am not familiar with gaphors code.

Thanks,

amolenaar:
@nomisge Thanks for the clarification.

I think the best way to add this is to make it a CSS feature. Something like:

```css
[...]

PR #2969, opened 2023-12-31
Fine grained CSS support

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

CSS can only be applied to elements (class, association). We can not write CSS that target specific parts in the element, e.g. attributes shown in a Class.

Issue Number: #2919 

### What is the new behavior?

* I added a new shape type: `CssNode`. With this shape we can tell Gaphor to update the style sheet for a specific element, such as an attribute owned by a class.
* Add support for the `::after` pseudo-element. This allows us to create custom tags (e.g. `{abstract}` to make it more clear a class/operation is abstract).
* Added more type annotations.
* Computed styles are cached
* Style names have been applied to names (at least the ones shown in bold) and compartments.

- [x] Yes
- [ ] No

### Other information

There's still some work ahead:

* Add `CssNode`s for all elements that makes sense.
* Move inline styling to a style sheet.
* Support for classes, so we can address styling for compartments in a Class item as `.compartment`. I think this makes more sense that referring to it as an element. 
* Currently the CSS model is rooted in the Element. E.i. an element that is contained by another element does not inherit any properties from its owner. Should this change too? (#2977)
* Add a way to figure out what CSS nodes are used in an item. (#2977)

!! NB. The caching is still a bit flaky. I disabled caching for now, but that may result in a lot more style computations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ------
 - Support types for parameters
 - Restore windows in maximized and full-screen state
+- Fine grained CSS: elements in a presentation item can be changed from CSS (experimental)
 - Very long element names are now wrapped
 - Format files accessed from Flatpak via portals
 - Update Property Editor to use Gtk.ColumnView and ListView
--- a/_packaging/gaphor.spec
+++ b/_packaging/gaphor.spec
@@ -42,6 +42,7 @@ a = Analysis(  # type: ignore
     pathex=["../"],
     binaries=[],
     datas=[
+        ("../gaphor/diagram.css", "gaphor"),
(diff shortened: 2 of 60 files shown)


danyeaw (approved):
This is a really nice feature!!

Would you mind please adding a changelog entry so that we keep the release ready?

amolenaar:
@danyeaw I added a change log entry.

BTW. My long term (ultimate) goal is to get rid of the inline styles we use a lot currently, because they cannot be overridden.

danyeaw:
Thanks @amolenaar!

amolenaar:
It needs a little work to fix the linting errors introduced by this PR. Linting was less strict before #2979 was merged.

PR #2977, opened 2024-01-03
Make style sheets cascade

amolenaar:
This is a follow-up PR on  #2969.

### PR Type

- [ ] Bug fix
- [x] Feature

Currently, every item on the diagram is a root item, even if it's contained in another item

Issue Number: N/A

### What is the new behavior?

* Cascade styles down. From diagram -> item -> nested item -> in-item css node.
* `:root` has been fixed. Refers to the top-level diagram. (Note that there's also a DiagramItem, which is referred to as diagram!)
* CSS nodes in a presentation item are also iterated (for `:has()` and `:empty`)
* Styles are cached per diagram. Cache is reset as soon as an item needs updating.
* Removed all inline styling for C4 model items.
* Fixed traversal of item to nested `CssNode`s.
* Simplified `compile_style_sheet()` output. The function now makes sure the rules are in ascending order.
* Added more type annotations.
* Text now uses `color` as color if `text-color` is not defined. `text-color` is deprecated. 

- [x] Yes
- [ ] No

* Minor change in CSS computation.

### Other information

The `gaphor.core.modeling.diagram.StyledItem` class now depends on the interface of `gaphor.diagram.shapes.CssNode`. This is not so clean, but can be fixed by implementing a protocol in the future.


danyeaw (approved):
This looks like a nice improvement, thanks @amolenaar!

I merged main to resolve a merge conflict. However, mypy isn't happy since the node argument to `compute_node` is getting a `PseudoStyleNode`, and importing it from the styling module creates a circular import.

amolenaar:
I'm not done yet :)

amolenaar:
@danyeaw I think it's ready for review

danyeaw (approved):
Thanks @amolenaar!

Issue #2988, opened 2024-01-07
Swimlanes aren't displayed


bjornstromberg:
### Describe the bug
the picture in the online docs example does not reflect the text (swimlanes missing) nor the actual view of the example model in the repo.

### To Reproduce
1. Go to 'https://docs.gaphor.org/en/latest/coffee_machine_logical.html#functional-boundary-behavior'
2. watch pictures below and read text, text says there is swimlanes, picture does not..
3. open model in example dir (coffee-machine.gaphor) and watch the swimlanes the text talk about..
4. doublechecked the svg that was uploaded and it does not contain any swimlanes.

### Expected behavior
picture in the tutorial actually correctly represented, cause the example otherwise is a clever/fun way to make an example worthy of actually reading... 

this is also the reason i took myself the time to atleast give a bug report that the picture does not match between the model and the generated example pictures, unless the pictures are not auto generated from the model it might be more that differ, but this one stuck out..

### Screenshots
1. from online docs
![picture in docs](https://docs.gaphor.org/en/latest/_images/9fdf0b3c-7170-11ec-999a-f47b099bf663.svg)
2. uploaded copy for history reasons when the docs change..
![uploaded copy so old state of picture is stored for history](https://github.com/gaphor/gaphor/assets/5454791/3164a0e9-1584-4e1c-8bad-8ec7afe4766b)
3. exported as svg from my system from the model
![Functional_Boundary_Behavior](https://github.com/gaphor/gaphor/assets/5454791/965bb62a-4d68-4a15-86df-5fd1554b06bc)
4. exported as png on my system from the model
![Functional_Boundary_Behavior](https://github.com/gaphor/gaphor/assets/5454791/f3094f5b-c943-4612-acf0-d354fd39d634)

### OS
- [X] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.22.1 (Flatpak)

### Additional information
Alpine Linux 3.19 with KDE and 2.22.1 (Flatpak)

amolenaar:
Hi @bjornstromberg,

This is a weird one, indeed. The swimlanes element is there, but it never rendered anything:

<img width="1952" alt="image" src="https://github.com/gaphor/gaphor/assets/96249/6908d56d-b9b5-4278-9770-82ade4f6b56f">

bjornstromberg:
@amolenaar yea it felt like one of those wierd ones when the model actually contained the swimlane, 
[...]

PR #2996, opened 2024-01-09
Fix swimlanes

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Any item with a Box shape and no children, does not draw.
This goes for Swimlanes (partitions), but also rectangles.

Issue Number: #2988 

### What is the new behavior?

Swimlanes are shown again.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/UML/actions/partition.py
+++ b/gaphor/UML/actions/partition.py
@@ -6,9 +6,8 @@ from gaphas.item import NW, SE
 from gaphor import UML
 from gaphor.core.modeling import DrawContext
 from gaphor.core.modeling.properties import association
-from gaphor.core.styling import JustifyContent
 from gaphor.diagram.presentation import ElementPresentation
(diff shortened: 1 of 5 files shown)


danyeaw (approved):
Great fix @amolenaar!

I'm not sure if this is related to this bug, but shouldn't be make the default size of Swimlanes a bit larger?
image

amolenaar:
That should not be too hard.

amolenaar:
This is also a regression. In 2.22.1 the lanes are wider by default.

amolenaar:
@danyeaw I added some dummy boxes to ensure a minimal width. This broke since empty boxes no longer yield false in if-stamenets, hence the update code of `ElementPresentation` was executed, updating the `min_width` property.

The proper solution is to break it up in shapes, for text and the actual swimlanes. I think we lack a few features there still.

Issue #2994, opened 2024-01-08
CSS "highlight disconnect associations" causes all items to be red


mikekidner:
### Describe the bug
In 2.23 the CSS example that highlights disconnected items in red causes all items to be red.
/* Highlight disconnected lines in red */

:not([subject], :is(line, box, ellipse, commentline)) {
  color: firebrick;
}

### To Reproduce
Apply the above example CSS lines to your CSS.
See error

### OS
- [ ] Linux (Please put in notes the specific distro)
- [ x] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.23

### Additional information
Not sure if this points to any deeper issues with recent changes to CSS, or just needs updating/removing as an example CSS snippet in the documentation.

amolenaar:
Thanks. I noticed it also with our [UML documentation](https://docs.gaphor.org/en/latest/models/uml.html).

In 2.23, I added the option for an element to have nested css nodes (the element name is contained in it's own `name` node). This finally allows us to style on a much more fine grained level.

The problem is (IMHO), that the attribute rule (`[subject]`) does not make a distinction between the attribute not being set and the attribute not being available at all.

When I read the [docs at MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors), `[subject]` should match any object that has a subject attribute.

So if we would change the rule to:

```css
:not(:is(line, box, ellipse, commentline))[subject=""] {
color: firebrick;
}
```

it should actually do what we expect (which it currently doesn't).

PR #2997, opened 2024-01-09
Fix attribute check

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

`[subject]` should check for the existance of the subject property, not its value.

Issue Number: #2994

### What is the new behavior?

The `[subject]` expression now checks for the presence of the subject *attribute*, whether it's set or not.

- [x] Yes
- [ ] No

Fix CSS attribute.


danyeaw (approved):
Nice fix, thanks @amolenaar!

PR #3013, opened 2024-01-12
Run Linux tests on Mutter, instead of Xvfb

amolenaar:
Alternative for #3014 

### PR Type

- [ ] Bug fix
- [ ] Feature

We use Xvfb for testing in CI

Issue Number: N/A

### What is the new behavior?

Use Mutter instead: Wayland has the future.

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -94,11 +94,10 @@ jobs:
         with:
           python-command: python${{ matrix.python_version }}
       - name: Run Gaphor Tests
+        env:
+          XDG_RUNTIME_DIR: /tmp
         run: |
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
Hi @amolenaar, this looks great to me! I noticed that Workbench also recently started using mutter, maybe there is a tip we can learn from their workflow? https://github.com/workbenchdev/Workbench/blob/main/.github/workflows/CI.yaml

amolenaar:
I think the approach is pretty similar. I prefer executing the tests as sub-process of mutter over starting it as a service. I do not see them launch a dbus session, or XDG_RUNTIME_DIR variable, but I guess that's because they use the default Ubuntu image.

danyeaw (approved):

PR #3025, opened 2024-01-14
Fix no matching distribution right after publish

danyeaw:
I noticed during the last two releases, that the dispatch action is failing to update the flatpak due to a no matching distribution error. When I restart the workflow it immediately passes, so I think PyPI needs a little time to sync the publish operation. This PR adds a wait time before running the workflow.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Dispatch workflow is failing, see https://github.com/flathub/org.gaphor.Gaphor/actions/runs/7521064978/attempts/1

Issue Number: N/A

### What is the new behavior?
Dispatch should pass

- [ ] Yes
- [X] No

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -436,6 +436,8 @@ jobs:
         with:
           egress-policy: audit
 
+      - name: Wait for PyPI to sync after version update
+        run: sleep 30s
       - name: Repository Dispatch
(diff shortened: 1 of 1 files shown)


amolenaar:
That seems reasonable. Alternatively, we can build in a retry mechanism.

amolenaar (approved):

PR #3028, opened 2024-01-16
Install Poetry as part of the RtD build

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

We're using [optional dependencies](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) to define the dependencies for our documentation generation on Read the Docs. This looks a bit out of order, since there's no use for those optional dependencies, except for the documentation generation.

### What is the new behavior?

Use Poetry. Now `docs` is just a Poetry group, like any other group.

I think this is more consistent. 

- [ ] Yes
- [x] No

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,7 +13,7 @@ build:
     - pip install --constraint=.github/constraints.txt poetry
     - poetry config virtualenvs.create false
     post_install:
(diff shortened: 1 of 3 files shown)


danyeaw (approved):
I love this! Thanks @amolenaar!

PR #3029, opened 2024-01-16
Make greeter window insensitive when a model is selected

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

If a model is big, a user does not get any feedback when it's selected in the greeter.

### What is the new behavior?

A small quality of life thing, that provides feedback when a user clicks on a recent model or template.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/ui/greeter.py
+++ b/gaphor/ui/greeter.py
@@ -161,16 +161,26 @@ class Greeter(Service, ActionProvider):
         self.close()
 
     def _on_recent_file_activated(self, row):
+        def load():
+            self.application.new_session(filename=filename)
+            self.close()
+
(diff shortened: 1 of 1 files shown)


danyeaw:
Hi @amolenaar, I think I am being a bit dense on this one, but I don't see any difference in behavior between this branch and main. Could you please explain more about what I am supposed to see?

amolenaar:
The change is minimal. With this PR, the greeter window will be made grayed out when you select a model. 

I change the priority f the load function. Now it should be more visible that the greeter window is made insensitive.

danyeaw (approved):
Got it, thanks @amolenaar!

PR #3041, opened 2024-01-22
Make Gaphor REUSE compliant

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

### What is the new behavior?

Add [REUSE](https://reuse.software/) compliance to this project.
This makes it explicit which licenses are used where.

To avoid editing every file, I used the [dep5](https://reuse.software/spec/#dep5) approach.

- [ ] Yes
- [x] No


danyeaw:
Hey @amolenaar, great idea! One thing I have been thinking about for our Copyrights as well is to remove the ending date, so Copyright © starting year, instead of range of years, since I don't think the range is really needed. Also it might be more inclusive of additional contributors if we changed it to The Gaphor Project instead of you and me. What do you think?

amolenaar:
> Hey @amolenaar, great idea! One thing I have been thinking about for our Copyrights as well is to remove the ending date, so Copyright © starting year, instead of range of years, since I don't think the range is really needed. Also it might be more inclusive of additional contributors if we changed it to The Gaphor Project instead of you and me. What do you think?

According to the REUSE docs, all versions (begin data, end data, begin-end data, no date) are all okay.

I think it's a good idea to contribute it to The Gaphor Project. All contributors are in our `.all-contributorsrc` file anyway.

danyeaw (approved):
Great! I added a few additional copyright updates for the Gaphor Development Team. I also added gaphorcss.lang to the dep5 file

pyproject.toml
```diff
@@ -187,7 +187,7 @@ docs-gettext-pot = { "cwd" = "docs", "shell" = "sphinx-build -b gettext . locale
 clean = { "shell" = "rm -rf dist build _packaging/dist _packaging/build _packaging/windows/file_version_info.txt docs/_build" }
 package = { "cwd" = "_packaging", shell = "pyinstaller -y gaphor.spec"}
 win-installer = { "script" = "_packaging.windows.build-win-installer:main" }
-gettext-pot = "pybabel extract -o po/gaphor.pot -F po/babel.ini -k i18nize gaphor"
+gettext-pot = "pybabel extract --omit-header --project Gaphor -o po/gaphor.pot -F po/babel.ini -k i18nize gaphor"
```
amolenaar, line 190: Why omit the header?

danyeaw: Hey @amolenaar, I thought the build was failing since there was license info in the .pot file. But looking again maybe not. If we are going to keep the header - we should update the project name and other info. Do you think the header is important?

danyeaw: I went ahead and dropped this commit, I think we can make this part of another PR if we try to remove or update the po headers.

danyeaw (approved):

pyproject.toml
amolenaar: Indeed. Let's change that in a future commit.

Issue #3085, opened 2024-02-06
[Flatpak] In some cases, files are stored as `.xdp-myfile.gaphor-<random string> `


ebousse:
### Describe the bug

When running Gaphor using Flatpak, creating a new file called `myfile.gaphor` with Gaphor may yield a file named `.xdp-myfile.gaphor-<random string> ` on the filesystem. 

Strangely, in my case, this only happens when I create a new project and save it, and this does not happen when I do _Save as_.

A workaround is to use a tool such as [Flatseal](https://flathub.org/apps/com.github.tchx84.Flatseal) and to give Gaphor new filesystem permissions, such as access to the home folder.

### To Reproduce

1. Start Gaphor with Flatpak.
2. Create a new empty project.
3. Save it with _Ctrl+S_ as `myfile.gaphor`

The file ends up saved as `.xdp-myfile.gaphor-<random string> ` on the filesystem.

### Expected behavior

The file should keep its original name.

### OS

Linux (Fedora 39 GNOME)

### Version

Version of Gaphor: 2.23.2
Version of Flatpak: 1.15.6

### Additional information

This seems to be a consequence of the [documents portal API of Flatpak](https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.Documents), which (I think?) makes it possible for Flatpak app to access the filesystem quite safely without any permissions.

But I do not understand why this is erratic (at least in my case), nor how the whole thing works.

Maybe Gaphor should consider asking for actual permission to use the filesystem, instead of using thie API? But this would reduce sandboxing, which would be a shame…

ebousse:
Good to know it's not just me :). 

Just to understand, is this a bug or the expected behavior of this portal?

PR #3089, opened 2024-02-07
Fix file saving for FreeDesktop Portals

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Files are not properly saved from Flatpaks if you do not define the `.gaphor` extension to the file.

Issue Number: #3085

### What is the new behavior?

Instead of adding an extension (which does not sync via XDP portals), provide a default name with extension.

This allows users to easily change the file name, with the extension already visible in the file entry field.

Do no longer "magically" add an extension.

- [ ] Yes
- [x] No


danyeaw (approved):
This looks great, thanks @amolenaar!

PR #3113, opened 2024-02-17
Add virus scan with VirusTotal to CI

danyeaw:
Since we have been having a lot of problems with false positives with virus scanners, this PR adds a scan with Virus Total each time we build the Windows installer. If it is a release, it adds links to the scan results. 

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: N/A

- [ ] Yes
- [X] No

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -390,6 +390,30 @@ jobs:
         run: type gaphor-self-test.txt
         shell: cmd
 
+  scan-for-viruses:
+    name: Virus Check
+    needs: check-windows-installer
+    runs-on: windows-latest
+    continue-on-error: true
+    timeout-minutes: 15
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@63c24ba6bd7ba022e95695ff85de572c04a18142 # v2.7.0
+        with:
+          egress-policy: audit
+      - uses: actions/download-artifact@eaceaf801fd36c7dee90939fad912460b18a1ffe # v4.1.2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/full-build.yml
```diff
@@ -390,6 +390,39 @@ jobs:
...
+      - name: VirusTotal Scan for Release
+        uses: crazy-max/ghaction-virustotal@810f296f41fda53a3e74461d76fc4ce1cab242d3 # v4.0.0
+        if: github.event_name == 'release'
+        with:
+          vt_api_key: ${{ secrets.VIRUSTOTAL_API_KEY }}
+          request_rate: 4
+          update_release_body: true
+          files: |
+            ${{ needs.windows.outputs.installer }}
+
```
amolenaar, line 425: Why not update the VirusTotal state on all builds?

amolenaar (approved):

.github/workflows/full-build.yml
danyeaw: I could try to simplify it so that update_release_body is set for all builds. Maybe it will ignore it if there isn't a release body. Is that what you meant by the state?

amolenaar: That may work. The docs say `update_release_body` is only applied for `release` events.

I did not realize it adds a comment to the release, instead of the pipeline results, though.

danyeaw: ya, it is a little weird because the build provides a link to the scanned results, but it doesn't wait until the scanning is done and provide results back to the build. I think this will at least be a step in the right direction

PR #3125, opened 2024-02-22
Do not show "block" stereotype if other stereotypes are applied

### LINKED ISSUE #2975: Removal or toggle of <<Block>> on stereotyped blocks
When using stereotyped blocks in SysML the stereotype label starts to get crowded due to appended stereotypes `<<block, stereo1>>`. If multiple stereotypes are explicitly specified, I'd expect them all to show up. However, `<<block>>` is defined as implicit for all unlabeled blocks:

The SysML 1.7 spec says:
image

So, there is an implicit association that everything's a block if unlabeled on a sysml bdd. 

When a block is stereotyped remove the text `<<block>>`

- add a toggle switch for displaying <<block>> stereotype
- Use classes instead of blocks, where possible, for a cleaner diagram (this is not an option if you need the block-specific attributes defined by SysML, but is okay for things like metamodels)

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Block stereotype is always shown

Issue Number: Fixes #2975

### What is the new behavior?

Show only "block" stereotype if no other stereotypes are applied.

- [ ] Yes
- [x] No

### Other information

I'm not sure if this should be applied to other similar elements.

### DIFF
--- a/gaphor/SysML/blocks/block.py
+++ b/gaphor/SysML/blocks/block.py
@@ -15,6 +15,7 @@ from gaphor.SysML.sysml import Block, ValueType
 from gaphor.UML.classes.klass import attributes_compartment, operation_watches
 from gaphor.UML.classes.stereotype import stereotype_compartments, stereotype_watches
 from gaphor.UML.compartments import name_compartment
+from gaphor.UML.recipes import get_applied_stereotypes
 from gaphor.UML.umlfmt import format_operation, format_property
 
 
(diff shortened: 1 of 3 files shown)


danyeaw:
@amolenaar When I tested this, I didn't see a stereotype when I create a block (with no other stereotypes applied).

amolenaar:
Thanks. I should have checked that properly. Now it's fixed.

danyeaw (approved):
This looks great! Nice idea to add some additional tests as well.

Issue #2563, opened 2023-07-27
Sequence diagram - drag and drop support


sz332:
### Is your feature request related to a problem?

I would like to be able to set things using drag and drop instead of manually selecting for sequence diagrams.

### Describe the solution you'd like

- Drag and drop a class, block, etc. to lifeline block shall set it's "type" (technically through a property, but as a modeler I don't really care)
- Drag and drop of an operation shall assign an operation to a message
- Drag and drop of a Signal should set the type of the message

amolenaar:
Fancy :wink:

sz332:
> Fancy 😉

Haha, and I just started thinking about how much we can help the work of a modeler by using drag and drop as many places as possible.

amolenaar:
Implemented in #3126.

PR #3126, opened 2024-02-22
Drop a ConnectableElement on a Lifeline

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: #2563

### What is the new behavior?

When you drop a Connectable Element (Interface, Property, Parameter) on a Lifeline, the type is set.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/UML/interactions/__init__.py
+++ b/gaphor/UML/interactions/__init__.py
@@ -1,3 +1,4 @@
+import gaphor.UML.interactions.drop  # noqa: F401
 from gaphor.UML.interactions import (  # noqa: F401
     copypaste,
     interactionsconnect,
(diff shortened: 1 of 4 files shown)


danyeaw:
Hi @amolenaar, I have tried to get this to work by dropping Blocks, Classes, and Properties on a Lifeline, but I haven't been successful at getting a type set. What am I doing wrong?

amolenaar:
This works for if you drag a Connectable Element (Property, Interface) from the tree model on a lifeline.

It doesn't work for Classes and Blocks, since those are not Connectable Elements.

Since the issue mentoined "drag and drop", I actually implemented it as a DnD actions.

Maybe this code is only complicating things, and we should just use the property editor?

danyeaw (approved):
Thanks for the explanation, sorry I was slow to understand the feature and slow to review it :smile:

Issue #2905, opened 2023-12-06
Start with empty model or example already filled


SebCanet:
### Is your feature request related to a problem?
When launching SysML profile, it's not empty so I have to modify everything: delete diagrams I don't need, erase blocks or change names, etc.

### Describe the solution you'd like
Start with an empty profile or an example (like it is actually).

### Describe alternatives you've considered
None possible.

### Additional context
Educational context.

SebCanet:
Thanks, I understand the way you think we should use. But for my stdent they will see in big the "SysML" and click on it,
I think you should seperate "new model" and "template", this panel seems confusing for newbies on your software.

marek-piirikivi:
"template" is actually in the description where you see SysML button.
![image](https://github.com/gaphor/gaphor/assets/45357878/fe5535ce-2594-4d14-a95b-dc70e6d03df8)

I partially agree with @SebCanet that people who want to create new SysML model will most likely click "SysML" option (me included).

I do not have a good proposal on the course of action. I admit that I am also guilty of making the mistake with this UI, but then again, the UI itself is explicit on what it promises and that's precisely what you get.

SebCanet:
As you choose a profile after generic choice, maybe we could have 2 columns:
UML Template ->                          UML empty model ->
SysML Template ->                        SysML empty model ->
RAAML Template ->                       RAAML empty model ->
CA Template ->                             C4 empty model ->
Empty model ->

Or start choose between Template or Empty model, then choose UML or SysML or ... or empty.
I'm trying to imagine how my students will react :-D

marek-piirikivi:
There have also been thoughts about not having the profile distinction from the drop-down at all and you can choose explicitly whether you want to create UML elements/diagrams or SysML diagrams. For example, right clicking on package > SysML > Create activity diagram; or right click > UML > Create activity diagram. First UI would kinda still remain the same, though whether you want to start with SysML template or UML template.

danyeaw:
My experience with production SysML models is that the starting templates are extremely complete. Users do exactly what you mentioned of renaming elements and modifying diagrams set up in a modeling method already. In the case of our starting template, the modeling method is a form of MagicGrid.

This has huge advantages, users don't have to start modeling from scratch which is a tremendous amount of work, it improves modeling consistency, and provides guidance to a modeling method.

We should continue to improve the starting model templates, not remove them.

[...]

PR #3134, opened 2024-02-24
Create separate section for examples and templates

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Users have to clean the UML, SysML, and RAAML templates if they want to use only the structure.

Issue Number: #2905

### What is the new behavior?

Templates should not contain additional elements.

The Examples for the same structure as the templates, but filled with example data.

I think examples do not need a separate title.

image

- [ ] Yes
- [x] No


danyeaw (approved):
Thanks @amolenaar, this is a nice solution!

PR #3136, opened 2024-02-24
Use TypedElement.type for Lifeline and other types

### LINKED ISSUE #2514: Lifeline and message traceability to Block/Actor/Property
Lifelines are currently untyped or referenced to other blocks which inhibits traceability between behaviors (operations) and components. I would like to be able to assign types to lifelines (from other preexisting blocks). Alternatively, (better) would be to display existing parts (blocks or Actors) on interaction diagrams as lifelines so they are inherently tied together with the associated messages/operations. 

See SysML Table 12.1. "The Sequence diagram describes the flow of control between actors and systems (blocks) or between parts of a system."

image

The individual lifelines shown in the spec are pre-existing components in the model. Either an unnamed block (just typed), a named block (e.g., a typed property from an internal block diagram), or an actor. Creating lifelines from blocks/actors/properties may be the more optimal route as opposed to adding a type selector for a lifeline.

Related: the SysML site specifically calls out avoiding using strings as the messages and instead using operations/signals. It would be useful if new messages introduced in the interaction diagrams could be tied back to their source block/actor as operations. Auto-assigning the operation to the source block might work initially. However, it may be desirable to assign operations to ports on the source block instead of the block itself. The corollary is that operations already defined on the originating block (or its ports) could be selected when inserting a message instead of generating a new operation on the source components.

image

#### Example
Shown below is the intended end goal with an interface behavior diagram and accompanying port definition for the control unit taken from Holt, "Systems engineering demystified", Pakt, 2023

image

image

Noted above

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Lifeline should show name of represented type (`ConnectableElement`)

Issue Number: Fixes #2514

### What is the new behavior?

Show represented name and type.

In addition I changed all cases where `TypedElement.typeValue` is used to also check for `TypedElement.type.name`. In all cases `type` takes precedence over `typeValue`, since type is actually linked to a model element and `typeValue` is something specifically added for Gaphor.

Also fixed messages: when attached to "life lines" without an actual lifeline, they act as a communication diagram.
In this case, lines so not need to be horizontal.

- [ ] Yes
- [x] No

### Other information

It would be nice if the Property Editor also shows settings for elements selected in the Model Browser. E.g. now it's not possible to change the attribute type easily.

### DIFF
--- a/gaphor/UML/umlfmt.py
+++ b/gaphor/UML/umlfmt.py
@@ -52,10 +52,10 @@ def format_property(
         s.append(name)
 
     if type:
-        if el.typeValue:
-            s.append(f": {el.typeValue}")
-        elif el.type and el.type.name:
+        if el.type and el.type.name:
             s.append(f": {el.type.name}")
+        elif el.typeValue:
+            s.append(f": {el.typeValue}")
 
     if multiplicity:
         s.append(format_multiplicity(el))
@@ -163,7 +163,9 @@ def format_parameter(
 
     s.append(name or "")
 
-    if type and el.typeValue:
+    if el.type and el.type.name:
+        s.append(f": {el.type.name}")
+    elif type and el.typeValue:
         s.append(f": {el.typeValue}")
 
     if multiplicity:
--- a/gaphor/UML/umlfmt.py
+++ b/gaphor/UML/umlfmt.py
@@ -163,10 +163,11 @@ def format_parameter(
 
     s.append(name or "")
 
-    if el.type and el.type.name:
-        s.append(f": {el.type.name}")
-    elif type and el.typeValue:
-        s.append(f": {el.typeValue}")
+    if type:
+        if el.type and el.type.name:
+            s.append(f": {el.type.name}")
+        elif el.typeValue:
+            s.append(f": {el.typeValue}")
 
     if multiplicity:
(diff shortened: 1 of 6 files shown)


gaphor/UML/umlfmt.py
```diff
@@ -163,7 +163,9 @@ def format_parameter(
 
     s.append(name or "")
 
-    if type and el.typeValue:
+    if el.type and el.type.name:
+        s.append(f": {el.type.name}")
+    elif type and el.typeValue:
```
amolenaar, line 168:
```suggestion
    if type and el.type and el.type.name:
        s.append(f": {el.type.name}")
    elif type and el.typeValue:
```

gaphor/UML/umlfmt.py
```diff
@@ -163,7 +163,9 @@ def format_parameter(
...
         s.append(f": {el.typeValue}")
```
amolenaar, line 169:
```suggestion
    if type:
        if el.type and el.type.name:
            s.append(f": {el.type.name}")
        elif el.typeValue:
            s.append(f": {el.typeValue}")
```

danyeaw (approved):
These updates look really nice, thanks @amolenaar!

PR #3144, opened 2024-02-25
Clean up property pages

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Property editors are often (but not always) encapsulated in a `GtkExpander`. There's little need for that, since the number of editors is not *that* long.

Issue Number: N/A

### What is the new behavior?

Simplify property editors. 

* Info/help boxes are now placed on a more logical place.
* Expanders have been removed.
* _Show Xxx_ switches are placed after data entry.

- [ ] Yes
- [x] No

### Other information

This will make it easier to split model and view based properties in the future. In the end we want to be able to use the property editor for elements selected in the Model Browser.

image

### DIFF
--- a/gaphor/SysML/propertypages.ui
+++ b/gaphor/SysML/propertypages.ui
@@ -2,11 +2,12 @@
 <interface>
   <requires lib="gtk" version="4.0" />
 
-  <object class="GtkExpander" id="item-flow-editor">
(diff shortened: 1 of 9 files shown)


danyeaw (approved):
Hi @amolenaar, this is a really nice simplification - this gives the Property Pages a much cleaner look!

PR #3145, opened 2024-02-25
Update dependencies and translations

danyeaw:
- Update poetry to version 1.8.1. This was causing the docs to fail to build, so I had to set the virtualenv environmental variables per the readthedocs instructions
- Bump our dependencies
- Update translation files


amolenaar:
Nice to be up to date again 👍

PR #3148, opened 2024-02-26
Element selected in Model Browser shows in Property Editor

### LINKED ISSUE #2545: Open element property page for an element selected from the model explorer
Sometimes I want to look/edit the details of an element without having to find and click on it from the diagram.

Clicking on an model element from the model explorer should open its property page.

An alternative is to click on the model element from the diagram.

### Additional context
This feature would be first steps towards supporting modeling directly in the context menu, which is known to be practiced by more seasonal modelers because it enables to do some things much faster than through the diagram. A follow up that would allow to create some modeling elements using a context menu should follow to support that.

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Only items selected in a diagram are shown in the Property Editor.

Issue Number: Fixes #2545 

### What is the new behavior?

When you select an element in the Model Browser, it will show in the Property Editor.

Property pages that contain both model and presentation settings (normally a _Show Something_ setting) have been split in two pages.

- [x] Yes
- [ ] No

### Other information

Still to do:

- [x] Attributes and Operations for Block, Requirement, etc.
- [x] Select Extension relation in model browser causes error
- [x] ActivityParameterNode also shows ObjectNode properties

### DIFF
--- a/gaphor/UML/actions/activitypropertypage.py
+++ b/gaphor/UML/actions/activitypropertypage.py
@@ -19,6 +19,7 @@ from gaphor.diagram.propertypages import (
 )
 from gaphor.UML.actions.activity import ActivityParameterNodeItem
 from gaphor.UML.propertypages import (
+    ShowTypedElementPropertyPage,
     TypedElementPropertyPage,
     create_list_store,
     list_item_factory,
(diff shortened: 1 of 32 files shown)


amolenaar:
@danyeaw By separating the model and presentation specific properties, I found out that some property editors are now applied to elements they were not applied to before.

E.g. `ActivityParameterNode` now also has _Upper Bound_ and _Ordering_ editors, since it inherits from `ObjectNode`.

image

Now we can explicitly blacklist those uses, but I'm not sure if we should. What do you think?

danyeaw (commented):
Hi @amolenaar, wow, this is a really nice feature to be able to show the property editor from the model browser!

One thing I noticed while testing this:
1. Create a Class and a Block
2. Select the Block in the Diagram -> Block Property Page is shown
3. Select the Class in the Model Browser -> Class Property Page is shown
4. Try to click on the Block in the Diagram again -> Class Property Page is still shown

amolenaar:
Thanks for testing this. It's a bit tricky to orchestrate the events coming both from the model browser and diagram. I'll have a look into this.

amolenaar:
@danyeaw I think I resolved the (re)selection issue. `DiagramSelectionChanged` events are now emited explicitly, and not based on changes in the diagram selection.

gaphor/UML/actions/activitypropertypage.py
```diff
@@ -206,47 +207,74 @@ def _on_name_changed(self, entry):
...
+            "show-parameter-direction-editor",
+            signals={
+                "show-direction-changed": (self._on_show_direction_changed,),
+            },
+        )
+
+        show_direction = builder.get_object("show-direction")
+        show_direction.set_active(self.item.show_direction)
+
+        print("(self.item.subject and self.item.subject.parameter)")
```
danyeaw, line 276: One more print statement nit

danyeaw (approved):
Nice, this is working great now! :star_struck:

gaphor/UML/actions/activitypropertypage.py
amolenaar: Ah no. I need a linting rule that errors on print statements!

PR #3204, opened 2024-03-21
Make self-test a command, instead of an option

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Self test is an option. This is from the time we did not have commands.

Issue Number: N/A

### What is the new behavior?

Now you can self-test by calling `gaphor self-test`.

- [ ] Yes
- [x] No


danyeaw (approved):
This looks great, thanks!

Issue #3154, opened 2024-02-27
Ends are not detached in a relationship when cancelling out the animation 


0lione:
### Describe the bug
When pressing ESC while detaching a relationship from one end that end will not be detached 

### To Reproduce
1. Create 2 Classes
2. Create a Relationship between them
3. Try to detach one end of the relationship and press ESC 
4. See that not only the end is still red it will also show that the end is still attached 

If possible, add the error report from `≡` (top-right menu) → `Tools` → `Error Reports`.

### Expected behavior
I think that when canceling out the animation in this circumstance the end should be detached

### Screenshots
![image](https://github.com/gaphor/gaphor/assets/101414931/839fd353-640e-473c-9e36-c317345c5769)

### OS
- [x ] Linux (Fedora 39)
- [ ] macOS
- [x ] Windows

### Version
Version of Gaphor: 2.24.0

### Additional information
If this is truly a bug can i work on it?

PR #3209, opened 2024-03-25
Fix #2736 & #3154: Cancelling the relationship connection

### LINKED ISSUE #2736: Cancelling out of drawing a relationship causes an exception
### Describe the bug
Pressing ESC to exit the creation of a relationship causes an exception.

### To Reproduce
1. Create a diagram
2. Add Class 1
3. Add Class 2
4. Select a relationship and start drawing from the Class 1.
5. Before reaching the Class 2, press Escape to cancel out the operation
6. Observe the error.

```
Gaphor version:         2.21.0
Operating System:       Windows (10)
Display:                GdkWin32Display
Python version:         3.11.5
GTK version:            4.12.0
Adwaita version:        1.3.1
GtkSourceView version:  5.8.0
Cairo version:          1.17.8
Pango version:          1.50.14
PyGObject version:      3.44.1
Pycairo version:        1.24.0
pygit2/libgit2 version: 1.12.2  / 1.6.4


Errors:

Time since application startup: 0:02:17
|Traceback (most recent call last):
|  File "C:\Program Files\Gaphor\gaphor\diagram\tools\placement.pyc", line 105, in on_drag_end
|  File "C:\Program Files\Gaphor\gaphor\diagram\tools\placement.pyc", line 89, in connect_opposite_handle
|  File "C:\Program Files\Gaphor\gaphor\diagram\tools\handlemove.pyc", line 48, in glue
|  File "C:\Program Files\Gaphor\gaphas\handlemove.pyc", line 71, in glue
|AttributeError: 'NoneType' object has no attribute 'model'
└─────────────────────────────╌┄┈
Time since application startup: 0:02:25
|Traceback (most recent call last):
|  File "C:\Program Files\Gaphor\gaphor\diagram\tools\placement.pyc", line 105, in on_drag_end
|  File "C:\Program Files\Gaphor\gaphor\diagram\tools\placement.pyc", line 89, in connect_opposite_handle
|  File "C:\Program Files\Gaphor\gaphor\diagram\tools\handlemove.pyc", line 48, in glue
|  File "C:\Program Files\Gaphor\gaphas\handlemove.pyc", line 71, in glue
|AttributeError: 'NoneType' object has no attribute 'model'
└─────────────────────────────╌┄┈

```
### Expected behavior
The operation should be canceled and the incomplete relationship removed from the model.

### OS
- [x] Windows

### Version
Version of Gaphor: 2.21.0

0lione:
When pressing Esc while drawing a relationship caused an exception instead of cancelling the animation.

### PR Type

- [x ] Bug fix
- [ ] Feature

When the user presses Esc while connecting any 2 elements with a relationship the Pointer shortcut is invoked and nothing is done to handle the key pressed, so when we already have a relationship established and press Esc while disconnecting one of the handles nothing happens to the relationship and the handle continues to be red. 

Issue Number: 2736 & 3154

### What is the new behavior?
Now when we press Esc we invoke a handler that will check if the handle we were moving is connected or not, if it is not then we disconnect that handle and we unselect the connection, if it is connected or the selected object is not a connection we just unselect it. The key aspect is that this handle is called before we invoke the Pointer shortcut. 

- [ ] Yes
- [ x] No

### DIFF
--- a/gaphor/diagram/presentation.py
+++ b/gaphor/diagram/presentation.py
@@ -75,6 +75,9 @@ class HandlePositionEvent(RevertibleEvent):
 
 
 class HandlePositionUpdate:
+
+    _lastHandleMoved = None
+
     def watch_handle(self, handle):
         handle.pos.add_handler(self._on_handle_position_update)
 
@@ -85,6 +88,7 @@ class HandlePositionUpdate:
         for handle in self.handles():  # type: ignore[attr-defined]
             if handle.pos is position:
                 self.handle(HandlePositionEvent(self, handle, old))  # type: ignore[attr-defined]
+                self._lastHandleMoved = handle
                 break
 
 
@@ -226,6 +230,8 @@ class LinePresentation(gaphas.Line, HandlePositionUpdate, Presentation[S]):
 
         self.watch_handle(self.head)
         self.watch_handle(self.tail)
+
+        self._has_been_dropped = False
 
     head = property(lambda self: self._handles[0])
     tail = property(lambda self: self._handles[-1])
--- a/gaphor/diagram/tools/shortcut.py
+++ b/gaphor/diagram/tools/shortcut.py
@@ -3,15 +3,16 @@ from gi.repository import Gdk, Gtk
 
 from gaphor.core import Transaction
 from gaphor.core.modeling import Presentation, self_and_owners
+from gaphor.diagram.presentation import LinePresentation
 
 
 def shortcut_tool(event_manager):
     ctrl = Gtk.EventControllerKey.new()
-    ctrl.connect("key-pressed", on_delete, event_manager)
+    ctrl.connect("key-pressed", key_pressed, event_manager)
     return ctrl
 
 
-def on_delete(ctrl, keyval, keycode, state, event_manager):
+def key_pressed(ctrl, keyval, keycode, state, event_manager):
     """Handle the 'Delete' key.
 
     This can not be handled directly (through GTK's accelerators),
@@ -21,6 +22,8 @@ def on_delete(ctrl, keyval, keycode, state, event_manager):
     if keyval in (Gdk.KEY_Delete, Gdk.KEY_BackSpace):
         delete_selected_items(view, event_manager)
         return True
+    if keyval == Gdk.KEY_Escape:
+        unselect(view, event_manager)
     return False
 
 
@@ -32,3 +35,14 @@ def delete_selected_items(view: GtkView, event_manager):
             if i.subject and i.subject in self_and_owners(i.diagram):
                 del i.diagram.element
             i.unlink()
+
+def unselect(view: GtkView, event_manager):
+    items = view.selection.selected_items
+    with Transaction(event_manager):
+        for i in list(items):
+            if isinstance(i, LinePresentation):
+                if i._has_been_dropped and i._lastHandleMoved and not i._lastHandleMoved.glued and i._connections.get_connection(i._lastHandleMoved):
+                    i._connections.disconnect_item(i, i._lastHandleMoved)
+                    i._lastHandleMoved = None
+        view.selection.unselect_all()
+        view.selection.grayed_out_items = set()
--- a/gaphor/diagram/tools/itemtool.py
+++ b/gaphor/diagram/tools/itemtool.py
@@ -2,6 +2,7 @@ from gaphas.tool import item_tool as _item_tool
 from gi.repository import Gtk
 
 from gaphor.diagram.event import DiagramSelectionChanged
+from gaphor.diagram.presentation import LinePresentation
 
 
 def item_tool(event_manager) -> Gtk.GestureDrag:
@@ -21,4 +22,8 @@ def on_drag_begin(gesture, _start_x, _start_y, event_manager):
 
 def on_drag_end(gesture, _offset_x, _offset_y):
     view = gesture.get_widget()
+    for i in view.selection.selected_items:
+        if isinstance(i, LinePresentation):
+            i._lastHandleMoved = None 
+            i._has_been_dropped = True
     view.selection.dropzone_item = None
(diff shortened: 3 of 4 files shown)


amolenaar:
Hi @0lione,

Thanks a lot for this PR. At first glance it looks good. I need a bit of time for a more in-depth review.

gaphor/diagram/presentation.py
```diff
@@ -75,6 +75,9 @@ def revert(self, target):
 
 
 class HandlePositionUpdate:
+
+    _lastHandleMoved = None
```
amolenaar, line 79: Can you change this to underscore notation (snake case):

```suggestion
    _last_handle_moved = None
```

gaphor/diagram/tools/shortcut.py
```diff
@@ -32,3 +35,20 @@ def delete_selected_items(view: GtkView, event_manager):
             if i.subject and i.subject in self_and_owners(i.diagram):
                 del i.diagram.element
             i.unlink()
+
+
+def unselect(view: GtkView, event_manager):
```
amolenaar, line 40: I would use a more descriptive name here:

```suggestion
def cancel_handle_drag(view: GtkView, event_manager):
```

amolenaar (changes requested):
This looks good.

Just a few comments regarding naming.

Maybe it's possible to create a separate `Gtk.EventControllerKey` in the `item_tool()` function (it should return a tuple of controllers then), so they can share a single state object (as is [done here](https://github.com/gaphor/gaphor/blob/main/gaphor/diagram/tools/placement.py#L32)). Then there's no need to maintain that state on the presentation objects themselves.

But if you feel that complicates things, that's also fine.

amolenaar:
@0lione Can you run pre-commit on your changes? That should help you resolve the linting errors.

0lione:
> This looks good.
> 
> Just a few comments regarding naming.
> 
> Maybe it's possible to create a separate `Gtk.EventControllerKey` in the `item_tool()` function (it should return a tuple of controllers then), so they can share a single state object (as is [done here](https://github.com/gaphor/gaphor/blob/main/gaphor/diagram/tools/placement.py#L32)). Then there's no need to maintain that state on the presentation objects themselves.
> 
> But if you feel that complicates things, that's also fine.

Could you elaborate on that?  I'm not understanding what you want me to do.

amolenaar:
Sorry for the late reply.

What I meant was that it may be possible to keep the whole cancel logic in `itemtool.py`.

```python
# gaphor/diagram/tools/itemtool.py

class DragState:
    def __init__(self):
        self.dragging = False


def item_tool(event_manager) -> tuple[Gtk.GestureDrag, Gtk.EventControllerKey]:
    state = DragState()
    gesture = _item_tool()
    gesture.connect_after("drag-begin", on_drag_begin, event_manager, state)
    gesture.connect_after("drag-update", on_drag_update)
    gesture.connect_after("drag-end", on_drag_end, state)

    key_ctrl = Gtk.EventControllerKey.new()
    key_ctrl.connect("key-pressed", key_pressed, event_manager, state)
    return gesture, key_ctrl

...

def key_pressed(ctrl, keyval, keycode, _state, event_manager, state):
    if state.dragging and keyval == Gdk.KEY_Escape:
        Transaction.mark_rollback()
```
(may need to make `gaphor.transaction.Transaction.mark_rollback()` a class method)

I think what it should do, is roll back the transaction. Then the handle should be nicely be reverted to its old position and everything should be as before the drag operation started.

0lione:
> Sorry for the late reply.
> 
> What I meant was that it may be possible to keep the whole cancel logic in `itemtool.py`.
> 
> ```python
> # gaphor/diagram/tools/itemtool.py
> 
> class DragState:
>     def __init__(self):
>         self.dragging = False
> 
> 
> def item_tool(event_manager) -> tuple[Gtk.GestureDrag, Gtk.EventControllerKey]:
>     state = DragState()
>     gesture = _item_tool()
>     gesture.connect_after("drag-begin", on_drag_begin, event_manager, state)
>     gesture.connect_after("drag-update", on_drag_update)
>     gesture.connect_after("drag-end", on_drag_end, state)
> 
>     key_ctrl = Gtk.EventControllerKey.new()
>     key_ctrl.connect("key-pressed", key_pressed, event_manager, state)
>     return gesture, key_ctrl
> 
> ...
> 
> def key_pressed(ctrl, keyval, keycode, _state, event_manager, state):
>     if state.dragging and keyval == Gdk.KEY_Escape:
>         Transaction.mark_rollback()
> ```
> 
> (may need to make `gaphor.transaction.Transaction.mark_rollback()` a class method)
> 
> I think what it should do, is roll back the transaction. Then the handle should be nicely be reverted to its old position and everything should be as before the drag operation started.

But when does the transaction get committed?  Because from what I could understand it only gets committed after drag-end, and the problem is if we press Escape the key-pressed function is going to be called and right as that function ends the shortcut Escape is going to be called meaning the drag-end function is never called and so the Transaction is never committed.

0lione:
Okay I got it working using the transaction rollback but I had to do some changes to the transactional_tool function so the rollback could work. I also had to change the PlacementState because whenever we pressed any key while placing a connection the "drag-end" event was never called. Hope the modifications are satisfactory, if not feel free to provide feedback on what changes you'd like to see

amolenaar:
I've been playing with your PR a bit. This problem is a lot harder than I expected.
The problem is indeed, that when you press _any_ key during a drag operation, the drag is cancelled and we end up with an inconsistent state, because our transactions are not properly closed. This causes errors during undo, for example.

I think you're on the right track, by adding a key-pressed handler to `transactional_tool`. 

I've been fiddling a bit around with that. I found that if we cancel (reset) all transactions as soon as a key press is recorded, we can roll back the transaction and let the tools do their cleanup.

Due to limitations in the GitHub interface (I was not able to create a PR from this repo to yours), I created a PR here. I did not need the key-pressed handlers for item and placement tools anymore. I think we should give this some thorough testing, though.

@0lione can you review the changes I did in #3254, please?

PR #3221, opened 2024-03-30
Replace Pygit2 by Dulwich

### LINKED ISSUE #2864: Investigate Dulwich as an alternative to pygit2

We currently depend on pygit2, which is a C library. This means we have to take extra care that the (latest) pygit2 version contains wheels for all platforms we support.

[Dulwich](https://github.com/jelmer/dulwich) may be a good pure-python alternative.

### Additional context

Dulwich should support conflict resolution and 3-way merge, since that's what we use under the hood.

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

PyGit2 relies on libgit, which is a C library. Dulwich is (almost) all Python.

Issue Number: fixes #2864 

### What is the new behavior?

Dulwich is used to handle merge conflicts.

This simplifies everything, since we can now always assume git support is available.

### Other information

Dulwich can't create merge conflicts by itself, so we use the git cli for that bit.

NB. Does merge conflict resolution work from Flatpak? We only have access to the actual file, and not to the .git folder. Maybe we should change the permissions back to `--filesystem=home`?


danyeaw (approved):
This is a nice improvement, thanks @amolenaar!

PR #3224, opened 2024-03-30
Associate Action with right Partition

### LINKED ISSUE #3050: First activity partition gets all the nodes
### Describe the bug
When querying activity partitions and seeing what nodes under which partition is, it was discovered that the first partition in the set gets all the nodes and all the rest are empty.

### To Reproduce
1. Create the following activity diagram
image
2. Open console and write
```
from gaphor.UML import uml
[f'{p.name}: {[n.name for n in p.node]}' for p in select(uml.ActivityPartition)]
```
3. See that the output is: `["Swimlane One: ['A1', 'A2']", 'New Swimlane: []']`

### Expected behavior

Expected output of the steps should be `["Swimlane One: ['A1']", "New Swimlane: ['A2']"]`

### OS
- [x] Linux (Mint)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.23.2

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Actions are always assigned to the first partition when dragging.

Issue Number: fixes #3050

### What is the new behavior?

The partition below the mouse cursor is now found.

### Other information

* Add a bit of extra typing to the core collection class.
* Added tests for State machine dropzone behavior.


danyeaw (commented):
Hi @amolenaar, these changes look good. I gave it a spin and for me the first drop is still being associated with the first swimlane even if I drop it on the 2nd. When I select the action and release it, then it is associated with the 2nd one.

amolenaar:
Ah, I only fixed it for moving actions from one partition to the other. I'll have a look at the `drop()` logic.

danyeaw:
Hi @amolenaar, I was dropping from the toolbox since I thought that would be the primary use case.

amolenaar:
I extended the use of `drop()` to also support dropping items on other items in a diagram. This simplifies the code, I think, since we only need one concept (drop) to handle items being placed on top of one another.

danyeaw:
@amolenaar Oh nice one! I haven't had a chance to look at the updates yet, but I should be able to make some time tomorrow to do one last review.

danyeaw (approved):
Really nice update to the drop zones @amolenaar!

PR #3241, opened 2024-04-07
Updates to getting-started documentation and coffee-machine tutorial

teunhoevenaars:
Small updates and suggestions to clarify issues encountered during the getting-started and coffee-machine tutorial.

### PR Checklist

NB the link [PR Review Checklist](https://gist.github.com/audreyr/4feef90445b9680475f2) on [Contributor guide](https://github.com/gaphor/gaphor/blob/main/CONTRIBUTING.md) is broken. Something for a later fix as I'm not sure where it is supposed to reference to.

### PR Type

- [ ] Bug fix
- [ ] Feature

- failed build of app on MacOS when following instructions
- manual configuration of poetry in-project environment
- some minor unclarities / confusion in the documentation
- documentation talks of black but since v2.22.0 this has been replaced with ruff
- coffee-machine.gaphor model has typo and unused requirement IDs
- internal links in Contribute page broken
- link to GNOME code of conduct is out-dated

Issue Number: N/A

### What is the new behavior?
- added build-babel step to MacOS build intructions and made use of git hooks explicit
- default configuration of poetry in-project environment for gaphor
- minor improvements to the documentation
- added ruff to dev dependencies
- requirements in coffee-machine.gaphor model have no IDs
- internal links in Contribute page work
- link to GNOME code of conduct up-to-date

- [ ] Yes
- [x] No

### Other information
This is my first PR to this project, and have done my best to follow all guidelines. In case I've missed some, then this is absolutely not intentional. Please feel free to revert any changes that you deem inappropriate. Many thanks for all your efforts!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ _packaging/windows/file_version_info.txt
 _packaging/windows/output
 
 # MacOS
+.DS_Store
 package/gaphor.iconset/
 
 # Vagrant
--- a/.gitignore
+++ b/.gitignore
@@ -51,7 +51,6 @@ _packaging/windows/file_version_info.txt
 _packaging/windows/output
 
 # MacOS
-.DS_Store
 package/gaphor.iconset/
 
 # Vagrant
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -63,8 +63,8 @@ Install [Poetry](https://python-poetry.org) using [pipx](https://pypa.github.io/
 pipx install poetry
 ```
 
-[Clone the
-repository](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/cloning-a-repository).
+Next, [clone the
+repository](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/cloning-a-repository), after which you need to execute the following consecutive commands:
 
 ```bash
 cd gaphor
@@ -72,6 +72,10 @@ cd gaphor
 pyenv local 3.x.x # 3.x.x is the version you installed earlier
 poetry env use 3.x # ensures poetry /consistently/ uses latest major release
 poetry install
+poetry run pre-commit install
+```
+Now, you can run gaphor as
+```bash
 poetry run gaphor
 ```
 
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -95,7 +95,7 @@ Create a file `.vscode/launch.json` with the following content:
             "module": "gaphor",
             "justMyCode": false,
             "env": {
-                "GDK_BACKEND": "wayland",
+                "GDK_BACKEND": "wayland"
             }
         }
     ]
@@ -108,7 +108,7 @@ Create a file `.vscode/launch.json` with the following content:
 
 The main method that Gaphor is packaged for Linux is with a Flatpak package.
 [Flatpak](https://flatpak.org) is a software utility for software deployment
-and package management for Linux. It offer a sandbox environment in which
+and package management for Linux. It offers a sandbox environment in which
 users can run application software in isolation from the rest of the system.
 
 We distribute the official Flatpak using [Flathub](https://flathub.org), and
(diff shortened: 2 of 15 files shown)


pyproject.toml
```diff
@@ -63,6 +63,7 @@ hypothesis = "^6.54.5"
 ipython = "^8.5.0"
 sphinx = ">=6,<8"
 babelgladeextractor = "^0.7"
+ruff = "^0.3.5"
```
danyeaw, line 66: Normally we run ruff through `pre-commit run --all-files`. Is there an advantage to adding it as a dependency?

poetry.toml
```diff
@@ -0,0 +1,2 @@
+[virtualenvs]
+in-project = true
```
danyeaw, line 2: Although we could specify this as a configuration, it isn't really required to run Gaphor. If someone prefers to have all of their virtualenvs managed in the cache-dir for Poetry, that is a fine preference, and I'm not sure if we should override their preference.

docs/linux.md
```diff
@@ -118,25 +121,25 @@ repository](https://github.com/flathub/org.gaphor.Gaphor).
 
 1. [Install Flatpak](https://flatpak.org/setup)
 
-1. Install flatpak-builder
+2. Install flatpak-builder
```
danyeaw, line 124: It is fine that you renumbered these, but Markdown does automatic numbering. This allows you to insert another number without manually renumbering everything below it.

.gitignore
```diff
@@ -51,6 +51,7 @@ _packaging/windows/file_version_info.txt
 _packaging/windows/output
 
 # MacOS
+.DS_Store
```
danyeaw, line 54: This would probably be better in your own global ignore file, since it isn't specific to Gaphor.

danyeaw (commented):
Hi @teunhoevenaars, thank you so much for this contribution, these are some really nice documentation updates :star_struck:.

I have a few comments about some of the dependencies changed and config files.

pyproject.toml
teunhoevenaars: hi @danyeaw, thanks for the appreciation! Indeed, there is no need to add this; it was some time after the commit in which I added ruff to the toml that I realized how to use pre-commit. I'll remove it.

poetry.toml
teunhoevenaars: I'm indeed also a person who generally keeps all virtualenvs in a single location :), but since gaphor does not build without the virtualenv being locally available, and configuring the virtualenv as local is therefore a necessary step to configure the dev environment (e.g. https://docs.gaphor.org/en/latest/macos.html#development-environment). As such, I thought this approach makes life easier. 

Nonetheless, I understand there are use cases in which this is not necessary so I'm happy to revert this. In that case, I would like to add a brief explanation on the use cases for which `poetry config virtualenvs.in-project true`is needed and add the `--local` option to cater to people like me.

docs/linux.md
teunhoevenaars: Yes, I realized later but it was more effort changing it back ;)

.gitignore
teunhoevenaars: Agreed, corrected

poetry.toml
danyeaw: Hi @teunhoevenaars, I just ran:

```
rm -r .venv
poetry config virtualenvs.in-project false
poetry install
Creating virtualenv gaphor-bY5AB7R2-py3.12 in /home/dan/.cache/pypoetry/virtualenvs
Installing dependencies from lock file
...
poetry run gaphor
```

Gaphor ran fine for me, which is what I was expecting. Are you having issues with running Gaphor like that?

teunhoevenaars: Thanks for checking @danyeaw! And yes, that indeed also works fine for me. The use case where it doesn't work is if I want to run
```
poetry install --with packaging
poetry run poe package
```
That's the use case that I referred to when I wrote 'use case for which [...] is needed'. Sorry for the confusion. I'll revert this change back tomorrow and make an alternative suggestion.

danyeaw: Interesting, thanks for pointing that out, I didn't realize poe required a local virtual environment. We should report that as a bug to poe library.

teunhoevenaars: After reading into the poethepoet documentation (https://poethepoet.natn.io/global_options.html#change-the-executor-type) I believe this was not a bug of the poe library, but an issue with the gaphor pyproject.toml where the poe executor was changed from the default (auto, where it uses the poetry managed environment) to a hard-coded, but default (local), `virtualenv` location. Removing that resolved the issue for me, after which I see no longer any need to specify the location of the poetry virtualenv. 

However, the fact that this config existed means that someone put some thought into this so I'm slightly worried that I've broken something somewhere. If so, I'm guessing that would come up in one of the CI tasks..

pyproject.toml
```diff
@@ -197,9 +197,6 @@ gettext-mo-release = { "script" = "po.build-babel:compile_mo_release" }
 translations = ["gettext-pot", "gettext-po", "gettext-mo-all"]
 icons = { "shell" = "make -C gaphor/ui/icons" }
 
-[tool.poe.executor]
-type = "virtualenv"
-
```
danyeaw, line 202: Nice one, it looks like this puts the configuration back to auto which should allow for centralized virtual environments :+1:

danyeaw (approved):
Thanks again @teunhoevenaars! This is a really great contribution!

Issue #2967, opened 2023-12-31
Support controlling visibility of requirement's text


marek-piirikivi:
### Is your feature request related to a problem?
Currently text is always visible. This takes up too much space on diagrams where the focus is just on the requirement relationship.

### Describe the solution you'd like
I would like to have a switch "Show text" on requirement property page.

### Describe alternatives you've considered
I couldn't come up with any other alternative other than to just ignore the problem.

amolenaar:
Fixed by #3249.

PR #3249, opened 2024-04-13
Control visibility of requirements

teunhoevenaars:
PR that addresses feature request to support controlling visibility of requirement's text

### PR Checklist

Additionally:
- [x] I have added unit tests around new feature

### PR Type

- [ ] Bug fix
- [x] Feature

- requirement text is always visible on diagram
- requirement box (on diagram) automatically expands in height when more text is added, but does not reduce in height when text is removed

Issue Number: #2967

### What is the new behavior?
- requirement is added to diagram as item (similar to blocks)
- Whether requirement text is shown on diagram can be toggled in property window of requirements
- Upon a change in text the height is automatically cropped to the minimum required.

hopefully not, but would love to hear your verdict :)
- [ ] Yes
- [ ] No

### DIFF
--- a/gaphor/SysML/propertypages.py
+++ b/gaphor/SysML/propertypages.py
@@ -29,31 +29,28 @@ from gaphor.UML.propertypages import (
 new_builder = new_resource_builder("gaphor.SysML")
 
 
-@PropertyPages.register(RequirementItem)
+@PropertyPages.register(sysml.Requirement)
 class RequirementPropertyPage(PropertyPageBase):
     order = 15
 
-    def __init__(self, item):
+    def __init__(self, subject: sysml.Requirement):
         super().__init__()
-        self.item = item
+        assert subject
+        self.subject = subject
+        self.watcher = subject.watcher()
 
     def construct(self):
-        if not self.item.subject:
-            return
-
-        subject = self.item.subject
-        watcher = subject.watcher()
-
         builder = new_builder(
             "requirement-editor",
             "requirement-text-buffer",
             signals={
-                "show-requirement-text-changed": (self._on_show_text_change,),
                 "requirement-id-changed": (self._on_id_changed,),
             },
         )
+        subject = self.subject
+
         entry = builder.get_object("requirement-id")
-        entry.set_text(self.item.subject.externalId or "")
+        entry.set_text(subject.externalId or "")
 
         text_view = builder.get_object("requirement-text")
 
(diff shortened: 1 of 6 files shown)


teunhoevenaars:
hi @marek-piirikivi, I recently discovered gaphor and as an exercise to get to know this library a little better I picked up your feature request (given the first-timers-only flag)! I would love to here if this PR addresses your request adequately :). Many thanks in advance for your feedback.

gaphor/SysML/propertypages.py
```diff
@@ -28,53 +29,63 @@
...
             signals={
+                "show-requirement-text-changed": (self._on_show_text_change,),
                 "requirement-id-changed": (self._on_id_changed,),
             },
         )
-        subject = self.subject
-
         entry = builder.get_object("requirement-id")
-        entry.set_text(subject.externalId or "")
+        entry.set_text(self.item.subject.externalId or "")
```
teunhoevenaars, line 56: this can be shortened to subject.externalId again

gaphor/SysML/requirements/requirement.py
```diff
@@ -27,26 +28,30 @@ class RequirementItem(Classified, ElementPresentation[Requirement]):
...
+
     show_stereotypes: attribute[int] = attribute("show_stereotypes", int)
 
     show_attributes: attribute[int] = attribute("show_attributes", int, default=False)
 
     show_operations: attribute[int] = attribute("show_operations", int, default=False)
 
     def update_shapes(self, event=None):
+        # reset height so requirement block is reduced to minimum required height
+        self.height = RESET_HEIGHT
```
teunhoevenaars, line 54: I added this on a hunch, but I don't fully understand why this implementation works to always crop the box height to fit the text, would love to hear a brief explanation!

gaphor/SysML/propertypages.py
```diff
@@ -28,53 +29,63 @@
...
+            builder.get_object("requirement-editor"), watcher
         )
 
     @transactional
     def _on_id_changed(self, entry):
-        self.subject.externalId = entry.get_text()
+        self.item.subject.externalId = entry.get_text()
+
+    @transactional
+    def _on_show_text_change(self, button, gparam):
```
teunhoevenaars, line 83: removing qparam from the input variables caused an unhandled exception in the app. This was however not caught by any unit test. I'll think about a way to capture the scenario that let the exception.

teunhoevenaars (commented):
I had some questions while working on this PR, would be fantastic if someone could point me towards the answer

gaphor/SysML/propertypages.py
```diff
@@ -28,53 +29,63 @@
 new_builder = new_resource_builder("gaphor.SysML")
 
 
-@PropertyPages.register(sysml.Requirement)
+@PropertyPages.register(RequirementItem)
```
amolenaar, line 32: It's better to have 2 separate property pages: 1 for the `sysml.Requirement` and one for `RequirementItem`. When you select an item in the model browser, it will only show the property pages related to the model element. When you select the diagram item, it will show the "presentation" property page too.

If you set the order to a bit bigger value than the model element property page, it will show right below.

This is one of the (big) changes we did for the next release.

gaphor/diagram/presentation.py
```diff
@@ -18,6 +18,9 @@
 from gaphor.diagram.shapes import CssNode, Shape, Text, stroke, traverse_css_nodes
 from gaphor.diagram.text import TextAlign, middle_segment, text_point_at_line
 
+RESET_HEIGHT = 50
+RESET_WIDTH = 100
```
amolenaar, line 22:
```suggestion
DEFAULT_HEIGHT = 50
DEFAULT_WIDTH = 100
```

amolenaar (changes requested):
This looks pretty good already. Really nice to see.

gaphor/SysML/requirements/requirement.py
teunhoevenaars: Thanks for the clarification! Indeed, I thought about the intentions of users as well, but since resizing only happens when a user changes the text (or toggles the new button) I think this is acceptable for those users.

gaphor/SysML/propertypages.py
teunhoevenaars: I see, I'll read more into this later

gaphor/SysML/propertypages.py
teunhoevenaars: Great, thanks so much for this suggestion; that makes perfect sense! I've made the requested change

teunhoevenaars:
Thanks a lot for the review @amolenaar! I believe I've processed all your comments :)

amolenaar:
Thanks @teunhoevenaars. The code looks really good and it works perfectly!

@all-contributors: Add @teunhoevenaars for code.

amolenaar (approved):

PR #3272, opened 2024-04-24
Workaround: Install GTK 4.14.0 on macOS untill 4.14.4 is available

amolenaar:
4.14.0 does not have the delay in screen updates.
This workaround should be reverted as soon as 4.14.4 is available.

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -200,6 +200,8 @@ jobs:
         uses: ./.github/actions/install
         with:
           python-command: python${{ env.python_version }}
+      - name: Run Gaphor Self-Test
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
Hi @amolenaar, these updates look good. Does this help users who are using macOS with Gaphor installed by homebrew? Would they have to clone the repository and manually use the homebrew install file as well?

amolenaar:
Hi,

Yes, if you install GTK via homebrew you can do the same trick. However, this should be removed as soon as 4.14.4 is released, so it's only a temporary fix that re-introduces issues with IME in popovers.

NB. you can also install the Gaphor DMG via homebrew, but I guess that's not what you mean 😉

Issue #3275, opened 2024-04-25
Gaphor shows diagram changed bar after saving


amolenaar:
### Describe the bug

When I save a project on macOS, it shows the _diagram has changed on disk_ banner

### To Reproduce
1. Open a model
2. Save the model
3. See the banner appear.

### Expected behavior

The banner should only show if the diagram is changed outside of Gaphor.

It looks like Linux is okay. Probably because its using inotify, whereas macOS uses polling.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [x] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.25+

### Additional information

Is this issue also on Windows?

teunhoevenaars:
hi @amolenaar I encountered the same on my Mac. I did a quick experiment (see https://github.com/gaphor/gaphor/pull/3278) where I added a sleep timer between saving and restarting the monitor; this prevents the pop-up on my machine if the timer is set to 2 sec (1 sec apparently is not enough). This doesn't feel like a good solution but perhaps it gives someone an idea how to resolve this more robustly.

PS I tried to install Gaphor on my Windows machine to test this behavior but did not manage; Gaphor wouldn't start at all. I'll try again later and if indeed consistent will make a separate bug.

amolenaar:
I think the best fix is to fix it in glib directly. I haven't checked how the file monitor is implemented for macOS at the moment. Until then, I suppose adding a timeout is completely fine. This is kind of a convenience feature anyway.

danyeaw:
Closed by #3278.

matgaj:
The same thing still happens in Windows on 2.25.1

danyeaw:
Hi @matgaj, thanks for letting us know about the issue you are having, but can you please open a new issue instead of posting on this one which was specific to macOS? Thanks!

PR #3278, opened 2024-04-28
Ignore event_type to prevent unexpected model changed popup

teunhoevenaars:
### PR Checklist

### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: #3275

### What is the new behavior?
time-out after save action, before file monitor is restarted

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/ui/filemanager.py
+++ b/gaphor/ui/filemanager.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import logging
 import tempfile
+import time
 from functools import partial
 from pathlib import Path
 from typing import Callable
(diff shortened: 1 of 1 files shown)


gaphor/ui/filemanager.py
```diff
@@ -334,9 +333,12 @@ def async_saver():
                     window=self.parent_window,
                 )
                 raise
+            else:
+                time.sleep(2)
+                self.filename = filename
+                self._update_monitor()
```
amolenaar, line 339: This should be handled via `GLib.timeout_add`. This way the whole UI freezes.

teunhoevenaars: in the end, the timeout really did not fully resolve this... I still saw the popups. Therefore, I just ran Gaphor with debugger and realized that the popups on Mac are triggered specifically by ATTRIBUTE_CHANGED events, which I don't think are events we're interested in. As such, I've now added a check against this event-type, which resolves this issue. Happy to hear your thoughts on this!

danyeaw (approved):
Hi @teunhoevenaars, thanks! These changes look great!

Issue #2104, opened 2023-02-24
Auto-layouting property based test takes a long time to run


amolenaar:
### To Reproduce
1. run `pytest`
2. See how it's waiting on `tests/test_auto_layouting.py`

### Expected behavior

The test should take a couple of seconds to run instead of several minutes.

### Version
Version of Gaphor: main

### Additional information

It's related to some changes made for Hypothesis 6.68.1.

amolenaar:
For the time being, I disabled the test.

PR #3282, opened 2024-04-28
Re-enable auto-layout tests

teunhoevenaars:
### PR Checklist

### PR Type

- [x] Bug fix
- [ ] Feature

test is disabled due to large amount of time required

Issue Number: #2104 

### What is the new behavior?
test completes in few seconds. Realized by setting the hypothesis testing to limit itself to generating examples (and not compare, reduce etc), see https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.Phase. Also limit the number and extend of tests.   

- [ ] Yes
- [x] No

### Other information
Feel free to adjust the setting parameters; I just picked something that achieved completion in few seconds and which I believed made sense.

### DIFF
--- a/tests/test_auto_layouting.py
+++ b/tests/test_auto_layouting.py
@@ -9,15 +9,14 @@ import contextlib
 import itertools
 from typing import Iterable
 
-import pytest
+import hypothesis
+from hypothesis import settings
 from hypothesis.control import assume, cleanup
 from hypothesis.errors import UnsatisfiedAssumption
 from hypothesis.stateful import (
     RuleBasedStateMachine,
     initialize,
-    invariant,
     rule,
-    run_state_machine_as_test,
 )
 from hypothesis.strategies import data, integers, sampled_from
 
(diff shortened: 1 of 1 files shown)


tests/test_auto_layouting.py
```diff
@@ -97,7 +92,8 @@ def select(self, expression=None):
         assume(elements)
         return sampled_from(elements)
 
-    def relations(self, diagram):
+    @staticmethod
```
danyeaw, line 95: I know IDEs often provide guidance that a method could be static. I would avoid this unless you want to directly call AutoLayouting.relations(). You could also move it from a method to a function outside of this class.

danyeaw (commented):
Hi @teunhoevenaars, thanks so much for the contribution to help us get our test turned back on. I just had one more comment.

@amolenaar, you are much more the hypothesis expert, any other suggestions?

tests/test_auto_layouting.py
teunhoevenaars: thanks @danyeaw, I indeed followed my IDE's suggestion on this without too much thought. I agree with your suggestion to move it outside the class, which is now implemented.

danyeaw (approved):
Thanks @teunhoevenaars!!

Issue #2831, opened 2023-11-12
Add Automatic Save to a Temporary File


alkis05:
### Is your feature request related to a problem?
App crashed and lost a lot of work

### Describe the solution you'd like
Of course I could save often, but it wouldn't heart to have a temporary file saved, say, after changing a diagram or something. This could be loaded in case the program closes for some reason before saving. At least offer the option for that. Like there is on text editors and many other tools.

### Describe alternatives you've considered
Just remember to save often.

amolenaar:
I've been thinking about this before.

It's a bit hard to do a full save operation: that will take a while for big projects.

Maybe it's an idea to build an event log and register all events that modify the model? We already do that for the undo functionality. If Gaphor crashes, we can replay the event log. 

The only uncertainty is what happens if the event log cannot be fully replayed? Should it replay up to the point where it errors?

amolenaar:
I've been implementing a little service in Gaphor that records all (model) changes.

Doing this adds hardly any overhead and all changes (in case of a crash or abrupt termination of Gaphor) can be replayed (it's re-using undo functionality to some extents). 

The idea is that a user doesn't loose any work in case of a failure.

I see three approaches at this point:

1. On application startup, show to-be recovered sessions in the Greeter window and allow users to recover them, or delete them 
2. When a model is loaded, any unsaved changes are automatically applied.
3. The application instantly opens with all user sessions, as if nothing happened.

For case 1 I created some screenshots.

<img width="912" alt="image" src="https://github.com/gaphor/gaphor/assets/96249/a27b97f3-0c92-4452-9cf7-ba5a2c6d34d7">

NB. A delete button is missing on the "recover" rows.


I'm not sure if the difference between _Recently Opened_ and _Recover Models_ is clear. Maybe add them all to _Recently Opened_ and add a note `(with unsaved changes)` ?

<img width="912" alt="image" src="https://github.com/gaphor/gaphor/assets/96249/dae3df78-a8d5-4d57-8f9f-b0d5d888825e">

@danyeaw What do you think?

danyeaw:
Hey @amolenaar, I really like your idea to integrate this with the current greeter! You also told me that you thought this might be getting too complex.

Since recovering from a crash should hopefully not be a frequent event, and because users should almost always want what they were working on back, what if we directly open the restored window?

danyeaw:
@all-contributors please add @alkis05 for idea

danyeaw:
[...]

PR #3284, opened 2024-04-28
Model recovery from an event log (recovery 3/3)

amolenaar:
> This PR builds on #3319 and #3322

### PR Type

- [ ] Bug fix
- [x] Feature

Models are not automatically saved.

Issue Number: #2831

### What is the new behavior?

In this PR, I want to pivot the idea of an event log. The event log contains all (model related) events. This log can be stored on disk and replayed. The goal is to provide a feature in Gaphor that prevents users from losing their work in case Gaphor crashes and the user has not saved his work in time.

The idea is that we log all transactions to a file. This file is based on the hash of the absolute filename of the model , similar to the per model user properties.
If such a log exists after a model is loaded, the log is applied and a notification is shown to the user.

The log will replay all user transactions, so the user should be able to continue his work and undo any step he's done previously. The model will show as modified.

When a model is saved, the event log (on disk) is truncated/reset.
If we have an event log for an unsaved model, we display that as first entry in the "Recent Models" list on the greeter window.

New models should get a unique event log file, so that you can have multiple new models open at any time.
The Greeter window should show any new, unsaved model (with a date?) so a user can re-open it and continue working.

#### Extra

* Move diagram update to sanitizer service, so they're also performed if no diagrams UI is opened yet
* move `get_config_dir()` and `get_cache_dir()` to `gaphor.settings`
* better cache events in main window, so notifications are displayed as soon as the window shows.
* Clean undo/redo stacks on model flush, instead of model-ready, so it does not interfere with recovery updates.

#### To do

- [x] Load saved model if recovery fails -> undo changes and remove event log?
- [x] What to do if the model changed in the meantime (e.g. due to `git pull`)? -> Add a checksum/hash at the beginning of the event log file.
- [x] Clear event log when a model is reloaded (model changed banner) -> checksum/hash in event log preamble does not match.
- [x] Do not replay events when the same model is opened twice. -> Second time, open model is "forced". In that case, we do not store the model file name in the Recovery service and no event log can be created.
- [x] How to deal with recording if the same model is opened twice? -> Same as above.
- [ ] How does a user know if a recovery log is created?

#### Future ideas

* Compact recovery records: reduce matrix updates and handle position changes.
* If the user exits the application, we can just quit, any changes not saved can be re-applied when Gaphor is started again. Gaphor can then always start with the models that were open the last time Gaphor was quit.
* Instead of basing the event log file on the model loaded (filename), it could be a unique identifier (a session id if you like). In the Greeter we can show all progress sessions with unsaved data (at the top, before the recent files). This gives users the option to either continue where they left off, or just open a model. This will also work for for new models that have not been saved to file yet.

- [ ] Yes
- [x] No

### Other information

An alternative is to actually save the model every so many minutes. However, saving can take a bit of time, especially for big models. With the event log approach we can restore the session to the point where Gaphor was shut down (or crashed).


gaphor/storage/recovery.py
```diff
@@ -0,0 +1,441 @@
...
+        self.events.append(
+            (
+                "d",
+                event.element.id,
+                event.property.name,
+                event.old_value and event.old_value.id,
+            )
+        )
+
+    # We should handle all RevertibleEvent subtypes separately:
```
danyeaw, line 298: Hi @amolenaar, do we still need these comments here?

gaphor/storage/recovery.py
```diff
@@ -0,0 +1,441 @@
...
+                    element.handles()[handle_index],
+                    connected,
+                    connected.ports()[port_index],
+                ).revert(element)
+            case ("ls", element_id, segment, count):
+                element = element_factory.lookup(element_id)
+                LineMergeSegmentEvent(element, segment, count).revert(element)
+            case ("lm", element_id, segment, count):
+                element = element_factory.lookup(element_id)
+                LineSplitSegmentEvent(element, segment, count).revert(element)
```
danyeaw, line 439: Are line split and line merge reversed here?

gaphor/storage/recovery.py
```diff
@@ -0,0 +1,441 @@
...
+                event.element.id,
+                event.segment,
+                event.count,
+            )
+        )
+
+
+def replay_events(events, element_factory, modeling_language):
+    for event in events:
+        match event:
```
danyeaw, line 380: Nice usage of the match statement :star_struck:

amolenaar: This is actually the first time I use this construct 😄

danyeaw (approved):
@amolenaar These changes look really great!

One thing I noticed is if I create a model, save it, edit it, and then force close Gaphor, when it comes back up it recovers the changes when I open the model which is great. However, I am not then able to save the model during that session from the menu or by hitting Ctrl+s. For example, I can hit Ctrl+s, the unsaved changes shows in the titlebar, and it will ask if I want to save when I go to close the model even though I just tried to save them.

amolenaar:
I see that the "edited" dot remains after saving. Saving itself works with me.

amolenaar:
@danyeaw The modified indicator works again.

I also changed how the modified indicator is set when changes are done, so it will always show modified after modification. Now, if you do _edit, save, undo_, it will show as modified.

danyeaw:
Nice one @amolenaar!

PR #3296, opened 2024-05-06
Ensure the diagram always has default focus

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Diagrams do not have focus by default. This is not handy, since it causes users to click on a diagram once before they can actually use keyboard shortcuts.

Issue Number: N/A

### What is the new behavior?

The diagram always gets focus:

* When opened
* When another diagram is selected

This way shortcuts work instantly.

Additionally, there was an issue that not all items are updated after loading (due to the new diagram update strategy). Now diagrams are explicitly updated after loading.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/storage/storage.py
+++ b/gaphor/storage/storage.py
@@ -171,6 +171,9 @@ def load_elements_generator(
         assert elem.element
         elem.element.postload()
 
+    for diagram in element_factory.select(Diagram):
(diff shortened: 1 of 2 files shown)


danyeaw (approved):
Thanks @amolenaar!

Issue #2821, opened 2023-11-08
Provide ability to call gaphor console externally


hulsed:
### Is your feature request related to a problem?

I am a researcher who would like to build some tools simulation tools connected to gaphor. One thing I would like to be able to do is call Gaphor's console from an external console (python instance, powershell, jupyter).

Currently, this is impossible without a development installation, which may be difficult or impossible to setup without administrative privileges. Preferably, it gaphor's console should be possible to call from a fully packaged version so you don't need to build gtk to interact with it. This would make it a lot easier to develop scripts which interact with gaphor's data model.

### Describe the solution you'd like

Something along the lines of:
- Provide a .exe for the gaphor python console
- Provide some documentation of working with it (e.g., using it as a kernel in jupyter, setting it as a sphinx plugin)

### Describe alternatives you've considered

- Package a standalone version of command-line gaphor packages for interacting with/rendering the data model without having to install the gtk/UI pieces (if possible).

danyeaw:
Hi @hulsed, to summarize this feature request, you are looking for these types of scripting capabilities like described here: https://docs.gaphor.org/en/latest/scripting.html, except you want to use it from a fully packaged version of Gaphor, is that right?

hulsed:
I am referencing the scripting capabilities, yes, but I'm not necessarily tied to "calling fully packaged version" as a solution.

I think the key functionality would be being able to interact with gaphor command-line services without needing to build gtk. 

There are different ways of accomplishing that:

1. If I could call a packaged version of gaphor (and call its console), ostensibly there would be a decent glide path to interacting with the underlying data model from an external shell (python, etc). I put this up as there but I'm not sure it's really the most elegant approach and could mean anyone who is scripting has to write needless glue code. But it may (?) be easier to implement in terms of not changing much in the back-end.


2. Right now, if I clone gaphor and run `from gaphor.core.modeling import ElementFactory` from a script in the repository, in loading all the modules it will say I need `gaphas` (because it is being imported in `gaphor.core.modeling.diagram`). This puts me in a pickle--if I run a `pip install gaphas`, I then need to build gtk, even though it's not really necessary for the functionality I want to interact with. So my problem would be solved if elements of the library would be refactored so gui libraries are only loaded when needed. Potentially, this could be a more elegant solution, because then the user would be working in a single consistent python environment, but I'm not sure what would need to actually be done--it could be as little as changing some imports or as much as creating a separate repo/package for the data model.

amolenaar:
[...]

PR #3310, opened 2024-05-17
Adding instructions for anaconda install

hulsed:
Adding docs page for installing anaconda within anaconda per our discussion here: https://github.com/gaphor/gaphor/issues/2821. It may require some nudging to fit within the rest of the documentation, but I've verified the process on an otherwise blank windows machine (perhaps this should be titled as windows-only? I am not sure how ).

- Docs page added for anaconda installation
- See: #2821

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Currently, it's difficult to work with gaphor from an external python console without a full development install which may require root permissions. This provides instructions for installing gaphor in an anaconda environment so can install and work with gaphor modules without root permissions.

Issue Number: #2821

### What is the new behavior?
This commit adds instructions for working in an anaconda environment.

- [ ] Yes
- [x] No


danyeaw:
Hi @hulsed, thanks so much for the contribution! If libadwaita and gtksourceview5 aren't available with anaconda what is the result of this environment? Are able to successfully install Gaphor and run scripts, but not able to launch Gaphor itself?

hulsed:
@danyeaw I added some language here that I hope explains that better. Unfortunately, I haven't really been able to demo this extensively to see what I can/can't do, but I can say:
- it lets you interact with some gaphor classes like SysML, ElementFactory, and ModelingLanguageService
- interacting with Session does not work, ostensibly because it wants gtk inspection. The error it gives is:
image

I'd like to characterize what you can and can't do here a bit better, but unfortunately--if I get a bit further working with this, perhaps I could perform some tests/searches and make a list or diagram?

danyeaw (approved):
Thanks so much for the contribution @hulsed! I think over time we can figure out if using Anaconda or some other solution is best for your use case, but I think adding some initial docs for the Anaconda crowd is great :+1:. Maybe we need to work to get libadwaita and gtksourceview packaged.

danyeaw:
@all-contributors please add @hulsed for docs

PR #3319, opened 2024-05-22
Get rid of ModelLoaded event (recovery 1/3)

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Issue:  #2831

We had three events: SessionCreated, ModelLoaded, and ModelReady.

SessionCreated is emitted when a new session is launched. A session may be based on an existing model (`filename` is set).

ModelLoaded is emitted when a model is loaded, and it also contains the file name

ModelReady is emitted when a model is loaded.

### What is the new behavior?

Cleaned up event structure.

There's some overlap in those events. Back in the day we had the option to load a model in an existing session. We do not allow that anymore. Therefore the ModelLoaded event is a redundant and has been removed.

- [ ] Yes
- [x] No

### Other information

I ran into this while implementing #3284.

### DIFF
--- a/gaphor/application.py
+++ b/gaphor/application.py
@@ -20,7 +20,6 @@ from gaphor.entrypoint import initialize
 from gaphor.event import (
     ActiveSessionChanged,
     ApplicationShutdown,
-    ModelLoaded,
     ModelSaved,
     ServiceInitializedEvent,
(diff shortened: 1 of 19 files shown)


danyeaw (approved):
Thanks @amolenaar, this changes look great!

PR #3323, opened 2024-05-26
Session recovery (recovery 4/3)

amolenaar:
This is an attempt to perfect a way to recover work after Gaphor has crashed, or terminated. The idea is that a user doesn't loose his work in case of a failure.

I see three approaches at this point:

1. On application startup, show to-be recovered sessions in the Greeter window and allow users to recover them, or delete them 
2. When a model is loaded, any unsaved changes are automatically applied.
3. The application instantly opens with all user sessions, as if nothing happened.

### PR Type

- [ ] Bug fix
- [x] Feature

In #3284, models are recovered by file name. This does not work if:

1. A model is opened multiple times: now, changes are just not tracked in the subsequently opened models
5. A model is opened by multiple processes: Only one event log will be saved.
6. No event log is generated for newly created model: a model needs a file name before Gaphor can create an event log

### What is the new behavior?

This PR is an attempt to improve session recording and recovery.

Sessions in the "sessions" folder are automatically restored on startup, also newly created (unsaved) models.

Each session will get a unique id. We can continue a session by providing it the id of a previously interrupted session.

This ensures that event logs are unique, since they're based of a session, instead of a model name. Also, we can record changes made to new models.

- [x] Yes
- [ ] No

### Other information

* TODO: test if recovery and sanitizer service interfere
* TODO: test a lot with undo/redo
* TODO: deal with faulty headers (e.g. not a dict)
* TODO: Is a ModelSaved event emitted and handled when a user saves the model on quit?

### DIFF
--- a/gaphor/UML/sanitizerservice.py
+++ b/gaphor/UML/sanitizerservice.py
@@ -67,7 +67,7 @@ class SanitizerService(Service):
 
     @event_handler(TransactionBegin)
     def _begin_transaction(self, event):
-        self.in_undo_transaction = event.context in ("undo", "redo", "rollback")
+        self.in_undo_transaction = bool(event.context)
 
     @event_handler(TransactionCommit, TransactionRollback)
     def _end_transaction(self, _event):
--- a/gaphor/application.py
+++ b/gaphor/application.py
@@ -110,10 +110,12 @@ class Application(Service, ActionProvider):
             force=force,
         )
 
-    def recover_session(self, *, session_id):
+    def recover_session(self, *, session_id, filename=None, template=None):
         """Recover a (crashed) session."""
 
-        return self._spawn_session(session=Session(session_id=session_id))
+        return self._spawn_session(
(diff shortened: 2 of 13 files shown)


danyeaw:
One interesting thing about this is that it also paves the way (if we want) to have the behavior like Elementary OS wants for closing an app here: https://docs.elementary.io/hig/user-workflow/closing. If you close and then reopen the app, you go right back to where you left off.

amolenaar:
> One interesting thing about this is that it also paves the way (if we want) to have the behavior like Elementary OS wants for closing an app here: https://docs.elementary.io/hig/user-workflow/closing. If you close and then reopen the app, you go right back to where you left off.

In that case we should always save the model when it's closed. No more dialog 😄 .

danyeaw (approved):
Amazing @amolenaar, really nice job. This brings so much better reliability to Gaphor with users not having to worry about losing work.

Issue #2136, opened 2023-03-08
Color Picker Widget


vanillajonathan:
### Is your feature request related to a problem?
Not everyone knows CSS and it can be tricky for some people to know how to style using CSS even if they're familiar with CSS.

### Describe the solution you'd like
I would like a color picker in the property editor to easily be able to assign a background color to a node such as a class or a component.

danyeaw:
Thanks @vanillajonathan, this is a duplicate of #2115

vanillajonathan:
This issue is not about toggling dark mode, I want a color picker widget to change the color of a node such as a class, component or interface in WYSIWYG manner.

FranciscoTGouveia:
Hi! Me and my colleague @0lione have realized that not many people are familiar with *CSS*, and that can be an obstacle for some Gaphor users while using the style sheet. With that being said, we would like to take on this issue, creating an interface to abstract the *CSS* and offer a more *user-friendly* way of changing colors. Would it be possible for us to be assigned to this issue? Thanks in advance.

amolenaar:
@FranciscoTGouveia @0lione Do you have an idea on how to make CSS more user friendly?

FranciscoTGouveia:
We were thinking of starting with a simple color picker widget that would sit on the page property of the Item we wanted to change color. Initially we would be able to select the color in a *CSS-agnostic* way ([a visual color picker](https://docs.gtk.org/gtk3/iface.ColorChooser.html) maybe) but this interface could later support more style aspects than just color, like `line-width` for example.

amolenaar:
Maybe this also ties in with #2137 (style sheet presets)?

Also, we have not used CSS id's and classes yet, so you can also use those to your advantage :smiley:. 

I do have a slight preference that we do not add styling info on the presentation elements directly. I'm aware though that it's a lot to ask from a user to learn CSS in order to change a color.

0lione:
> Maybe this also ties in with #2137 (style sheet presets)?
> 
> Also, we have not used CSS id's and classes yet, so you can also use those to your advantage 😃.
> 
> I do have a slight preference that we do not add styling info on the presentation elements directly. I'm aware though that it's a lot to ask from a user to learn CSS in order to change a color.

But I thought that referencing with id's was not supported in gaphor

PR #3329, opened 2024-05-28
Implement StyleEditor (encompasses a Color Picker)

FranciscoTGouveia:
Instead of using the *StyleSheet*, one can now use the *StyleEditor*. The *StyleEditor* is a GUI that allows to style elements by their name and is coherent with the *StyleSheet*.

### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Right now, when we want to change the style of an Element we would need to go on the *StyleSheet* and reference it by it's name.
Many people aren't familiar with CSS and this can stand in their way while developing their diagrams.

Issue Number: #2136 

### What is the new behavior?
Instead of using the *StyleSheet*, one can now use the *StyleEditor*, a GUI that opens up from a button in the property pages. This allows a user to style their elements with visual cues and abstract themselves from the details of CSS and how it works under the hood.

- [ ] Yes
- [x] No

### Working example
[Use Case of StyleEditor](https://github.com/gaphor/gaphor/assets/93057203/d0af87da-4fd5-4c1b-812d-1ed05584f7b2)

### Important notes
+ When you style an element, you are styling that specific element (by their name) and not all elements of that type. This is true for all named elements except *relationships*.
+ The changes made in the GUI (*StyleEditor*) are coherent with the *StyleSheet*.
+ The changes made in the *StyleEditor* are only exported to the *StyleSheet* when asked, or when the file is saved.
+ The properties that the *StyleEditor* offers are related to "colors" (in order to fix the issue), and "border-radius", but more properties can be added (hopefully in a near future).

### Implementation details to discuss
+ We thought of a whole interface to abstract the CSS. Is this a good idea (that can be expanded), or shall we stay with just a simple Color Picker ?
+ The changes that we made with the *StyleEditor* are only being written to the *StyleSheet* upon request or when saving the file. Is this a good implementation or should we update the *StyleSheet* everytime we make a change on the *StyleEditor* ?
+ Since we didn't find a class that would cover all of the *Presentation* representations with names (i.g:. some inherit from *Named*, others from *ElementPresentation*, etc) we ended up using the most general one, *Presentation*. However, this means that all the *relationships* wouldn't be able to be referenced by name. Moreover, we had to add a *try-except* clause on the `load` method of *Presentation* since some *Items* don't have names.

Authors: 
[Francisco Gouveia](https://github.com/franciscotgouveia):  <francisco.t.gouveia@tecnico.ulisboa.pt>
[João Correia](https://github.com/0lione):  <joao.marques.correia@tecnico.ulisboa.pt>


amolenaar:
This looks pretty good.


>  * We thought of a whole interface to abstract the CSS. Is this a good idea (that can be expanded), or shall we stay with just a simple Color Picker ?

I think it's a good idea to keep the option for expansion open at least.

>  * The changes that we made with the StyleEditor are only being written to the StyleSheet upon request or when saving the file. Is this a good implementation or should we update the StyleSheet everytime we make a change on the StyleEditor ?

Is it easy to update elements in the. style sheet directly? It would be my preference since all styling is then in one place, but I can imagine it's requires a lot of processing to update.

>  * Since we didn't find a class that would cover all of the Presentation representations with names (i.g:. some inherit from Named, others from ElementPresentation, etc) we ended up using the most general one, Presentation. However, this means that all the relationships wouldn't be able to be referenced by name. Moreover, we had to add a try-except clause on the load method of Presentation since some Items don't have names.

I think you can just add one to `Presentation` and check if the subject exists and has a `name` attribute.
IIRC it becomes simpler once we start using KerML as our core model, since `Element` will have a name.

A few suggestions: I would move the StyleEditor to `gaphor.ui`, since it's a (g)ui related feature. I would also avoid abbreviated terms, especially in public methods (`style_elements` rather than `style_elems`).

0lione:
> This looks pretty good.
> 
> > * We thought of a whole interface to abstract the CSS. Is this a good idea (that can be expanded), or shall we stay with just a simple Color Picker ?
> 
> I think it's a good idea to keep the option for expansion open at least.
> 
> > * The changes that we made with the StyleEditor are only being written to the StyleSheet upon request or when saving the file. Is this a good implementation or should we update the StyleSheet everytime we make a change on the StyleEditor ?
> 
> Is it easy to update elements in the. style sheet directly? It would be my preference since all styling is then in one place, but I can imagine it's requires a lot of processing to update.
> 
> > * Since we didn't find a class that would cover all of the Presentation representations with names (i.g:. some inherit from Named, others from ElementPresentation, etc) we ended up using the most general one, Presentation. However, this means that all the relationships wouldn't be able to be referenced by name. Moreover, we had to add a try-except clause on the load method of Presentation since some Items don't have names.
> 
> I think you can just add one to `Presentation` and check if the subject exists and has a `name` attribute. IIRC it becomes simpler once we start using KerML as our core model, since `Element` will have a name.
> 
> A few suggestions: I would move the StyleEditor to `gaphor.ui`, since it's a (g)ui related feature. I would also avoid abbreviated terms, especially in public methods (`style_elements` rather than `style_elems`).

I feel the same way, however with our approach when an element changes it's name or when it's removed or even when it's loaded from a file, our "backend" takes care of this and updates the element accordingly. If we were to simply write the changes into the StyleSheet then, whenever the user deleted an element or change it's name there would be useless lines in the StyleSheet. For the backend to track these changes in the StyleSheet would be computationally heavy. That's why we've created the export button, to allow user to look under the hood if he wants to.

0lione:
Hi @amolenaar , I think we've already addressed the requested changes. Could we request another review?

tests/test_architecture.py
```diff
@@ -100,6 +101,7 @@ def test_modeling_languages_should_not_depend_on_ui_package():
         archrule("Modeling languages should not depend on the UI package")
         .match("gaphor.C4Model*", "gaphor.RAAML*", "gaphor.SysML*", "gaphor.UML*")
         .should_not_import("gaphor.ui*")
+        .may_import("gaphor.ui.styleeditor")
```
amolenaar, line 104: This rule should not be needed anymore.

If we allow (indirect) import of UI components that means that importing the UML/SysML packages does require GTK and libadwaita. This is something we do not want, since it makes it harder to use Gaphor as a library (without UI). By looking at the code it looks like it's not imported directly, only in the property pages, which are loaded as `gaphor.modules`.

amolenaar (commented):
Thanks @0lione, sorry for the late feedback.

Looks good. I think for now it's good to be specific on which elements have a custom styling.

I proposed some (mainly stylistic) changes. I have not been able to run the code yet.

tests/test_architecture.py
FranciscoTGouveia: Hi @amolenaar, I've deleted the rule, but since `gaphor.ui.styleeditor` will be (non directly) imported by property pages, I've added a flag to only check direct imports (`only_direct_imports=True`).

amolenaar: Ah, that's because I suggested to move the editor to the UI package. My bad, sorry 🙈 .

I pushed a commit that consolidates the style editor code (both property page and editor window) in one file, imported as module (you may need to `poetry install` for the module loaded).

I noticed that the style editor does not work when you open an existing model: it's linking `self.diagram.styleSheet` in the constructor, but during load time, no relations have been created yet when the presentation is created.

Given the number of fields we currently have in the editor, it may look nicer to use a `Gtk.Popover` instead of a window. Then the popover window can be linked to the editor button (with a dropdown icon). Also the text needs to be nicely alighed.

I also noticed the items do not update if a color is changed. I think we should trigger an update for the items on the current diagram, so they redraw with the new style.

0lione: > I noticed that the style editor does not work when you open an existing model: it's linking `self.diagram.styleSheet` in the constructor, but during load time, no relations have been created yet when the presentation is created.
> 
Yes it is linking in the constructor and then in the load method of Presentation we perform a "name_change" so it restores the name.

danyeaw:
Hi @FranciscoTGouveia, what is remaining to get this merged?

FranciscoTGouveia:
Hello @danyeaw! I think it is ready to be merged. In my understanding this could be a good substitute for the _Style Sheet_ as it is, offering a graphical and user-friendly way of styling elements. However, this still has room for improvement (e.g. not all style options are available). With that said, I think that feedback from the community would be beneficial to know what path to follow regarding the _Style Editor_.

danyeaw:
Thanks @FranciscoTGouveia, would you mind please resolving the merge conflicts?

FranciscoTGouveia:
Hi @danyeaw ! While resolving the conflicts I found out that the button for opening the _Style Editor_ is not showing. As far as I can tell, this is related with the fact that the `StylePropertyPage` is now defined in a separate file from the other property pages.

To fix this, I could just import the `diagram/styleeditor` inside of `diagram/propertypages` but this would cause a circular import. While doing the import in the end of the file works, it doesn't seem like a good practice. With that said, should I undo the last commit and leave the `StylePropertyPage` in the `diagram/propertypages` or am I missing something?

CC @amolenaar

amolenaar:
I was wondering if the `Open Style Editor` button should be placed below or above the CSS editor area. Since styles are applied via the style sheet, it would make it clear to users where the styles go. The editor can be used as a tool to simply define the style.

I think we can move some logic up to the `Presentation` class, so that it does not have to be defined for every model element.

amolenaar:
I simplified the code a bit. Instead of having styling on every element, the style editor adds its (temporary) styling to `StyleSheet.instant_style_declarations`. To ensure diagrams are updated, I added a new event type: `StyleSheetUpdated`.

The colors are updated based on the current style of the selected item.
The styling is only persisted if you add it to the style sheet.

I updated the layout of the style editor:

image

danyeaw (approved):
This looks great, thanks so much @FranciscoTGouveia and @0lione! Thanks @amolenaar for the final update - integrating it with the style editor like this is really good idea as well.

danyeaw:
@all-contributors please add @0lione for code

PR #3335, opened 2024-06-01
Fix issue when recovering from split and merged segments

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Fix an issue I ran into when recovering an auto-layout model.

* Create a new model
* Add a line
* Split the line in a few segments
* move both ends
* Merge the line back to 1 segment
* Interrupt Gaphor (ctrl-C)
* Start gaphor again, see it crash

Issue Number: #2831

### What is the new behavior?

The Handle position event contains the new location, so we can record it, even if the handle has been removed in the mean time.

- [ ] Yes
- [x] No

### Other information

Now all recovery events are written with info from the event only. No lookups (except for id) need to be done on the element.

### DIFF
--- a/gaphor/core/modeling/presentation.py
+++ b/gaphor/core/modeling/presentation.py
@@ -151,6 +151,7 @@ class MatrixUpdated(RevertibleEvent):
     def __init__(self, element, old_value):
         super().__init__(element)
         self.old_value = old_value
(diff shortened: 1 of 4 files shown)


danyeaw (approved):
Thanks @amolenaar!

Issue #3348, opened 2024-06-12
undo manager AssertionError


Sycophantic-Witty:
### Describe the bug
Selecting a relationship in the Model Browser or right-clicking a relationship in a diagram and selecting "Show in Model Browser" triggers an AssertionError.

### To Reproduce
1. Click on a relationship in the Model Browser

Time since application startup: 0:00:17
|Traceback (most recent call last):
|  File "C:\opt\dev\gaphor-2.25.1-portable\data\_internal\gaphor\transaction.pyc", line 166, in _transactional
|  File "C:\opt\dev\gaphor-2.25.1-portable\data\_internal\gaphor\transaction.pyc", line 59, in __init__
|  File "C:\opt\dev\gaphor-2.25.1-portable\data\_internal\gaphor\transaction.pyc", line 114, in _handle
|  File "C:\opt\dev\gaphor-2.25.1-portable\data\_internal\gaphor\transaction.pyc", line 190, in handle
|  File "C:\opt\dev\gaphor-2.25.1-portable\data\_internal\gaphor\application.pyc", line 191, in _transaction_proxy
|  File "C:\opt\dev\gaphor-2.25.1-portable\data\_internal\gaphor\core\eventmanager.pyc", line 77, in handle
|  File "C:\opt\dev\gaphor-2.25.1-portable\data\_internal\generic\event.pyc", line 61, in handle
|ExceptionGroup: Error while handling events (1 sub-exception)
└─┬──────────────────────────────╌┄┈
  |Traceback (most recent call last):
  |  File "C:\opt\dev\gaphor-2.25.1-portable\data\_internal\generic\event.pyc", line 57, in handle
  |  File "C:\opt\dev\gaphor-2.25.1-portable\data\_internal\gaphor\services\undomanager.pyc", line 171, in begin_transaction
  |AssertionError
  └─────────────────────────────╌┄┈

### Expected behavior
Select a relationship in either the Model Browser or a diagram and show it in the other view.

### Screenshots
N/A

### OS
- [ ] Linux (Please put in notes the specific distro)
- [ ] macOS
- [x] Windows

### Version
Version of Gaphor:

Gaphor version:         2.25.1
Operating System:       Windows (10)
Display:                GdkWin32Display
Python version:         3.12.3
GTK version:            4.14.2
Adwaita version:        1.5.0
GtkSourceView version:  5.12.0
Cairo version:          1.18.0
Pango version:          1.52.2
PyGObject version:      3.48.1
[...]

PR #3357, opened 2024-06-19
Property Pages can now wire up services

### LINKED ISSUE #3351: Dependency injection for property pages

The property editor (right pane) contains input fields for editing data in the selected model element. The property editor contents is constructed dynamically (based on the element type) from several smaller components called Property Pages.

Currently, for property pages there is no way to reach info from other places in the system. 
This is most evident in the way transactions need to be handled with the `@transactional` decorator.

When property pages are constructed, it should be possible to inject services. The injection mechanism can be similar to what we use for services: based on parameter name. 

Leaving as is requires the property page to find the active session from a global context, which is more error prone.

### Additional context

After this, we can drop the `@transactional` decorator.

The `component_registry` service may be able to provide the functionality. We should also be able to provide positional arguments.

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

We have a global `@transactional` decorator. This decorator works as long as the system is fully aware of the active session. This of course causes trouble if a non-active session starts to do something in the background.

Issue Number: Fixes #3351.

### What is the new behavior?

Instead, all places where transactions are started should have access to the services, mainly `event_manager`. This ensures that the transaction is run in the right context.

- [x] Yes
- [ ] No

The `@transactional` decorator has been removed.

### Other information

I wanted to fix this for a long time. #3348 was the trigger to rethink the problem again. In the end it seemed pretty simple: the `component_registry` should have a method to (partially) apply services to any function or class.

### DIFF
--- a/docs/locale/cs/LC_MESSAGES/docs.po
+++ b/docs/locale/cs/LC_MESSAGES/docs.po
@@ -3622,7 +3622,7 @@ msgid "Overview of Transactions"
 msgstr ""
 
 #: ../undo.md:9
-msgid "The recording and playback of changes in Gaphor is handled by the the Undo Manager. The Undo Manager works transactionally. A transaction must succeed or fail as a complete unit. If the transaction fails in the middle, it is rolled back. In Gaphor this is achieved by the `transaction` module, which provides a context manager `Transaction` and a decorator called `@transactional`."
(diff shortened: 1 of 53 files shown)


danyeaw (approved):
Hi @amolenaar, these changes look great, nice improvement!

PR #3363, opened 2024-06-24
Fix deletion of attribute/operation in element editor

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

* (regression) Deletion of attribute/operation/enumeration/parameter in property editor does not work. 

Issue Number: N/A

### What is the new behavior?

Deletion works: it's wrapped in a transaction.

### Other information

Other things I like to fix here:

* Double click to edit a stereotype value
* Attribute/operation/enumeration reorder does not properly redraw the diagram item.


danyeaw (approved):
Thanks @amolenaar!

PR #3365, opened 2024-06-25
KerML Root model

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

### What is new?

Gaphor has a minimal core model.

This PR implements (part of) the KerML spec for the core model. I think this is a nice stepping stone to implementing, as KerML is the basis got SysML v2.

The idea is to slowly migrate to the KerML metamodel as a basis for Gaphor.

The new model can be viewed in the readthedocs build.

Issue Number: #2584

### Other information

- [x] Move `Relationship` to core model
- [x] Move `Dependency` to core model
- [x] Move `name` to `Element` (as is, not as a derived property)
- [ ] ~Remove use of `NamedElement`, once `name` is moved to `Element`~ `NamedElement` is used all over the place
- [x] Move `Namespace` to core model
- [ ] ~Move `DependencyItem` to diagram package?~ Hard, since it still depends on UML elements too.
- [ ] ~Deal with `Relationship.abstraction`. It's not part of the core model.~ Leave as exception for now. We also have some properties for `Namespace`, but it looks like we do not use those (MyPy is not complaining at least).
- [x] Restore `Named` interface again on `DirectedRelationshipPropertyPathItem`, `DependencyItem`, and `InterfaceRealizationItem`.

Review notes:

This PR has become pretty big.

A lot of files changed due to `name` and `namespace` now being part of the `Element` class: no casting is required in `watch()` operations

I tried to remove NamedElement. So far I got it out of the Model Browser, but there's some other places that depend on it still. Although NamedElement only contains a visibility attribute (not sure if we even use that).

I think we need a base type for Element at some point, so Element can be used for "normal" model elements and we can derive "gaphor-internal" types like Presentation and StyleSheet from the base type. This way Element can more easily replace NamedElement.

I tried to change the semantics of the model as little as possible. E.g. I did not implement the `Membership` and `OwnedMembership` relations.

### DIFF
--- a/gaphor/codegen/tests/test_coder.py
+++ b/gaphor/codegen/tests/test_coder.py
@@ -266,11 +266,11 @@ def test_attribute_from_super_model(uml_metamodel: ElementFactory):
     class_.name = "Package"
 
     element_type, base = attribute(
-        class_, "relationship", [(UMLModelingLanguage(), uml_metamodel)]
+        class_, "member", [(UMLModelingLanguage(), uml_metamodel)]
     )
 
     assert element_type is UML.Package
-    assert base.owner.name == "Element"
+    assert base.owner.name == "Namespace"
 
 
 def test_replace_simple_attribute(uml_metamodel: ElementFactory):
--- a/gaphor/codegen/tests/test_coder.py
+++ b/gaphor/codegen/tests/test_coder.py
@@ -210,8 +210,8 @@ def test_simple_attribute(uml_metamodel: ElementFactory):
 def test_order_classes(uml_metamodel):
     classes = list(order_classes(uml_metamodel.select(UML.Class)))
 
-    assert classes[0].name == "Element"
-    assert classes[1].name == "NamedElement"
+    assert classes[0].name == "NamedElement"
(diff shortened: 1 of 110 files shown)


gaphor/codegen/tests/test_coder.py
```diff
@@ -261,16 +269,24 @@ def test_coder_write_association_opposite_not_navigable(
     assert a == ['A.b = association("b", B)']
 
 
-def test_attribute_from_super_model(uml_metamodel: ElementFactory):
+@pytest.mark.xfail(reason="Needs further investigation")
+def test_attribute_from_super_model(
+    uml_metamodel: ElementFactory, core_metamodel: ElementFactory
+):
```
amolenaar, line 275: TODO: find out why this test fails now

amolenaar: order matters: core model should be declared first, since the code generator is looking for the _first_ match (by name) in a super-model).

danyeaw (approved):
Hi @amolenaar, these changes look great, nice start on our way to KerML :+1:

PR #3373, opened 2024-07-07
Be more lenient with association end multiplicity

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

You cannot define a combination of visibilitity (and derived) and multiplicity, without providing an association end name.

### What is the new behavior?

Be more lenient with what to accept. Parse visibility and derived properties too.

- [ ] Yes
- [x] No


danyeaw (approved):
Another nice update, thanks!

PR #3384, opened 2024-07-12
Fix undo and recovery of Dependency type

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

* undo does not work when changing dependency type
* dependency type changes are not recorded

Issue Number: N/A

### What is the new behavior?

Changes of element type are recorded for undo and recovery.

- [ ] Yes
- [x] No


danyeaw (approved):
This fix looks great, thanks! Should we do a new minor release?

Issue #2845, opened 2023-11-16
State Machine Triggers / Actions missing


AdamZsofi:
### Is your feature request related to a problem?
The state machine diagram feels really nice to use with composite states, guards, etc., but the lack of actions and triggers on transitions makes Gaphor for state machine modeling inadequate (at least for me).

I would like to use Gaphor in education: teaching UML state machines (among other diagrams). Correct syntax and understanding of basic semantics (e.g. in which order events are executed) is crucial and these elements play an important role in that.

### Describe the solution you'd like
I would like to have two simple labels on state machine transitions, where I can add actions and triggers (simply strings, just as it is for guards).
I would like the value of these to show around the middle of the transition in the following manner: trigger [guard] / action.

### Describe alternatives you've considered
I considered using the "name" label, but that shows up at the end of the transition and is often hard to read. I also considered adding all these to the guard field, but then the action and the trigger are showing in brackets (syntactically incorrect).

If it would be possible to show the value of the name field around the middle of the transition or if it would be movable, that would be adequate for me as well. I do not care about code generation and similar features, I just need a label around the middle of the transition, which I can write to.

I am not sure why this has not come up yet, I tried to search in older issue, but could not find anything. If the reason for this missing is just lack of developer time, I would also be willing to implement it (it seems fairly simple).

Thanks in advance!

marek-piirikivi:
Thank you for drawing attention to that. This would definitely be a fine addition.

If you want to get started with this on your own, few pointers to get you started quicker:

1. from gaphor/UML/uml.py you can see that class Transition already has guard and effect
2. To allow editing effect, you have to work with gaphor/UML/states/propertypages.py
3. For trigger, UML model has to be changed. Gaphor uses auto-generated uml.py, so after model is changed, uml.py needs to be regenerated which would allow you to enable editing trigger on the property page as well. Open UML.gaphor with the gaphor itself, define necessary relationships in the UML model; save it; and run `poetry run poe uml` to generate code.
4. For getting the transition information correctly drawn on the diagram, see gaphor/UML/states/transition.py

mikekidner:
Hi @marek-piirikivi,  
Thanks for the tips re remaking the model from the gaphor file. I was wondering how to do that.
I had a go at running poetry to recreate raaml.py from the RAAML model and it complained:
```
poetry run poe raaml
ModuleNotFoundError: No module named 'defusedxml'

```
But 
[...]

PR #3385, opened 2024-07-12
Stm trigger action

ghillebrand:
Adds `trigger` and `action` to a State Transition Models, in addition to `guard`, as per Issue [#2845](https://github.com/gaphor/gaphor/issues/2845)
trigger and action associations have been added to the UML model Transition Class Diagram , and the code updated to edit and display these attributes as `trigger [guard] /action`

### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Currently, only `guard` text can be entered, which is displayed below the transition as `[guard]`. 
Issue Number: [#2845](https://github.com/gaphor/gaphor/issues/2845)

### What is the new behavior?
Now `trigger` and `action` can independently be entered in the properties dialog, and are displayed according to the UML spec `trigger[guard]/action` on the transition. The text is _below_ the transition, rather than above, which seems to be the UML preference for this.

- [x] Yes
- [ ] No

This does not seem to be backwards compatible - models created with these changes do not open in older versions of gaphor. This seems to be related to the file loading code. On trying to open a model created with these changes, gaphor (2.25.1 under windows) does nothing - no error, no crash, no model or diagram displayed. 

### Other information
My first PR/ change, so if I have omitted/ mangled something, my apologies!

### DIFF
--- a/gaphor/UML/states/propertypages.py
+++ b/gaphor/UML/states/propertypages.py
@@ -26,7 +26,7 @@ new_builder = new_resource_builder("gaphor.UML.states")
 
 @PropertyPages.register(UML.Transition)
 class TransitionPropertyPage(PropertyPageBase):
-    """Transition property page allows to edit guard specification."""
+    """Transition property page allows to edit trigger, guard, and action specification."""
 
     order = 15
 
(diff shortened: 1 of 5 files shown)


amolenaar:
Hi @ghillebrand,

You're definitely on the right track. The code looks okay. To fix the mypy issue, I moved some code around.

Some other changes:

* Merged the code with the main branch. The `@transactional` has been deleted and the UML model had been updated.
* Fixed a small issue in the model, where 2 relations referred to the same association end on behavior.
* I moved the generation of the trigger/guard/action name to a separate method. I think that makes it somewhat more readable.

I was wondering if trigger and action should actually refer to exiting behaviors, such as a activity, state machine, and Interaction. Currently it's creating a new "Behavior" instance, with no relation to the rest of the model (Behavior should have been an abstract class in the model).

ghillebrand:
Thank you :) That all makes sense. 

ITO the behaviour issue, the OP was happy if the trigger-guard-action could be displayed "properly", without too much methodology behind it. I certainly don't know/ use UML enough to begin to make a call on that! 
If you can point me at how to make the text appear _above_ the line, I would happily add that. I would like to understand the drawing side of gaphor better, and that would then meet all the of OP's requirements as I understand them.

My goal is to create a new model type, higraph, which would heavily re-use the STM code, and this was an excellent project to get my head around how the bits of gaphor interact. I should open a feature request, yes? I will outline the steps as I see them, and be guided by you guys on how/ where to implement the various parts.

amolenaar:
@ghillebrand 

You refer to this?

image

The text positioning is mostly done automatic. The `LinePresentation` class initiates text placement and calls [`text_point_at_line`](https://github.com/gaphor/gaphor/blob/main/gaphor/diagram/text.py#L120), which will place the text close to the begin/center/end of the line. The position is dependent on the angle of the respected line segment.

Text placement is the same for all lines/vertices.

amolenaar:
This PR looks good. So I'll merge it.

ghillebrand:
> @ghillebrand
> 
> You refer to this?
> image
> 
> The text positioning is mostly done automatic. The `LinePresentation` class initiates text placement and calls [`text_point_at_line`](https://github.com/gaphor/gaphor/blob/main/gaphor/diagram/text.py#L120), which will place the text close to the begin/center/end of the line. The position is dependent on the angle of the respected line segment.
> 
> Text placement is the same for all lines/vertices.

Yes, this is what I meant, and this looks like the code I was looking for. The OP, and other tools, write the text above the line. When I get a chance, I will play with the hinting values, and see how they behave/ influence things. Controlling text position I would like to be able to do for higraphs.

PR #3406, opened 2024-07-25
Prepare for RTD migration to Addons

danyeaw:
On 2024-10-07, RTD will enable RTD Addons by default on all projects. This PR ensures that the html baseurl is injected in order to use our custom domain.

Everyone can read more about this here:
https://about.readthedocs.com/blog/2024/07/addons-by-default/

Once this is merged, I should be able to enable addons ahead of time by following these steps: https://docs.readthedocs.io/en/stable/addons.html#enabling-read-the-docs-addons


amolenaar (approved):
Nice one!

PR #3425, opened 2024-08-07
Add documentation for developing and debugging plugins on Windows

pbrown12303:
Current documentation lacks guidance on setting up a development environment for plugins. The enhanced documentation provides an overview of setting up a Windows development environment for a plugin and debugging it with VSCode

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: N/A

- [ ] Yes
- [x] No


docs/windows.md
```diff
@@ -128,6 +128,33 @@ Install Gaphor's dependencies
...
+When setting up a plugin workspace you need to perform the following steps:
+
+cd (your project's workspace)
+
+If your project does not already have a pyproject.toml file, create one. For details see the [Poetry documentation](https://python-poetry.org/docs/basic-usage/)
+```PowerShell
+poetry init
+```
+
+Install your project's dependencies. If you have made your project dependent upon gaphor, this will pull in gaphor
```
amolenaar, line 153: Note that Gaphor should be installed as _development_ dependency, like in the [hello-world plugin](https://github.com/gaphor/gaphor_plugin_helloworld/blob/main/pyproject.toml#L16-L18).

docs/windows.md
```diff
@@ -139,11 +166,19 @@ poetry run pip install --force-reinstall (Resolve-Path C:\gtk\wheels\PyGObject*.
 poetry run pip install --force-reinstall (Resolve-Path C:\gtk\wheels\pycairo*.whl)
 ```
 
+Install your plugin
```
amolenaar, line 169: Hmm.. This is what `poetry install` takes care of.

docs/windows.md
pbrown12303: Added a note to the line 148 instructions indicating that gaphor should be set up as a development dependency in the .toml file and referencing the hello world plugin. Won't show up until I check in the change.

docs/windows.md
pbrown12303: Bear in mind here that I am a python/poetry novice, but by my observation the "poetry install" does not create the .venv/Lib/site-packages/<plugin>.dist-info directory needed to start the plugin. This command creates that directory (and its contents), so when you start gaphor the plugin is present (you don't even need the "GAPHOR_PLUGIN_PATH") environment variable.

danyeaw: If we have Gaphor as a development dependency for the plugin and the plugin is using poetry, why is there a need to pip install it? This doesn't sound right.

pbrown12303: I can verify by experiment that if you do not do this, gaphor will crash the first time you put a class in a diagram. I think the reason is that the two pip install lines place the results in the local repository. Making gaphor a design dependency does not, I think, bring the gaphor directory's .venv into play (which is where the results go when you install gaphor as a development environment). 

However, the ```poetry run pre-commit install``` line needs to be removed.

danyeaw: `poetry run` should use the poetry virtualenv. Do you have your plugin code somewhere? I could try to reproduce this to see if I can figure out what is going on.

pbrown12303: The experiment I ran was with the gaphor hello world plugin. I created a folder, did git clone to get the plugin, then followed the steps I documented. Try it without the two wheels step and see what you get.

danyeaw: Hi @pbrown12303, I just reran your steps with updated gaphor hello plugin dependencies that I submitted here: https://github.com/gaphor/gaphor_plugin_helloworld/pull/138

I was successfully able to launch Gaphor with the plugin without running `pip install -e .` manually. Do you want to see if you can reproduce this as well?

pbrown12303: Hi @danyeaw. Are you sure you don't have the hello plugin installed in the default location? Try putting a breakpoint in your development copy of the hello plugin and see if it actually stops there.

danyeaw: Hi @pbrown12303, I definitely didn't have it installed. I had to clone the repo because it wasn't on this computer.

PR #3440, opened 2024-08-12
Async support

amolenaar:
This PR explores the possibilities for async/await support that will be available in PyGObject 3.50.

asyncio is a common trait among Python developers. By moving from GLib idle handlers and callbacks to async/await we can lower the barrier for python devs to join.

### PR Type

- [ ] Bug fix
- [x] Feature

We're using idle functions to schedule work in the background. How cool would it be if we can just submit async/await jobs?

Issue Number: https://gitlab.gnome.org/GNOME/pygobject/-/issues/636

### What is the new behavior?

Explore how async/await can make our code more readable and possibly better testable.

Not sure if this will make it in main, but it's worth testing this new (experimental) functionality. We may come up with a few patterns that are useful for other PyGObject users.

- [x] Yes
- [ ] No

Async functions are introduced.

### Other information

* I tried to make everything async: open the file with `aiofiles.open()`, and await everything down to the `XMLWriter`. It works, but makes saving really slow. I'd rather have a fast save than a slow save with a progress bar.


gaphor/asyncio.py
```diff
@@ -0,0 +1,68 @@
...
+
+@contextlib.contextmanager
+def glib_event_loop_policy():
+    original = asyncio.get_event_loop_policy()
+    policy = GLibEventLoopPolicy()
+    asyncio.set_event_loop_policy(policy)
+    try:
+        yield policy
+    finally:
+        asyncio.set_event_loop_policy(original)
```
amolenaar, line 22: Here we should wait for all background tasks to be completed.

danyeaw:
Hi @amolenaar, this seems to work really well from my testing. Do you think we should try to merge this to main or make it configurable to turn on?

amolenaar:
@danyeaw I think we can merge this. There are some improvements we can implement once the next version of PyGObject is released.

danyeaw (approved):
Thanks @amolenaar!

amolenaar:
@danyeaw thanks for fixing the last bits.

PR #3461, opened 2024-08-24
File name width

### LINKED ISSUE #3460: Long filenames in header force width of main window preventing access to right hand toolbar
### Describe the bug
When opening a file with a very long path, the displayed filename above the main window forces the window width, which makes the RHS menu inaccessible if you have a small screen (like a laptop)
### To Reproduce
Save a file with a really long filename see how right menu is forced out of the application view. 

### Expected behavior
Maybe truncate the filename or allow the right hand main window edge to be moved, forcing the filename to be truncated. 
### Screenshots
Screen Shot 2024-08-22 at 8 22 44 am

Screen Shot 2024-08-22 at 8 21 18 am

### OS
- [x ] macOS
- [ ] Windows

### Version
Version of Gaphor: 2.26

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: Fixes #3460.

### What is the new behavior?

* file name in title bar is elipsized, so it does not push away the property editor.
* Fix model show up as modified on load.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/services/undomanager.py
+++ b/gaphor/services/undomanager.py
@@ -142,7 +142,7 @@ class UndoManager(Service, ActionProvider):
         event_manager.priority_subscribe(self.undo_association_add_event)
         event_manager.priority_subscribe(self.undo_association_delete_event)
         event_manager.priority_subscribe(self.undo_element_type_updated_event)
-        self._action_executed()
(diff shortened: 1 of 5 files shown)


danyeaw (approved):
Great fix!

Issue #3477, opened 2024-09-03
Drop of ControlFlow into an Activity diagram does not automatically connect the arrow to the source and target objects in the diagram


pbrown12303:
### Describe the bug
When programmatically creating an activity diagram, if you drop a ControlFlow that has its source and target set into the diagram and the diagram already contains images of the source and target elements, the control flow image is not automatically connected to  the source and target images.

### To Reproduce
1. Create an Activity and an Activity Diagram programmatically
2. Create two actions, A and B.
3. Drop actions A and B onto the diagram
4. Create a ControlFlow. Set its source and target to actions A and B.
5. Drop the ControlFlow onto the diagram.

### Expected behavior
The ControlFlow should connect automatically to the end-point images.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [ ] macOS
- [x] Windows

### Version
Version of Gaphor: 2.26.0

pbrown12303:
I think I found the problem. The code that handles connecting relationships is in drop.py. The specific function is:
```
@drop.register(Relationship, Diagram)
def drop_relationship_on_diagram(element: Relationship, diagram: Diagram, x, y):
    item_class = get_diagram_item(type(element))
    if not item_class:
        return None

    metadata = get_diagram_item_metadata(item_class)
    return (
        drop_relationship(
            element,
            metadata["head"].get(element),
            metadata["tail"].get(element),
            diagram,
            x,
            y,
        )
        if metadata
        else None
    )

```
The problem is that ControlFlow and ObjectFlow are not sub-classes of Relationship - they are sub-classes of ActivityEdge. I think the fix is to make a copy of this function and change the Relationship type to ActivityEdge in the copy.

danyeaw:
Closed by #3479.

Issue #2126, opened 2023-03-04
Autolayout is Inverting BDDs


danyeaw:
### Describe the bug
Autolayout is inverting BDDs:
![image](https://user-images.githubusercontent.com/10014976/222927765-84aa3c94-131b-4010-a0e1-868010bbe3e8.png)

### To Reproduce
1. Create a BDD
2. Tools -> Autolayout

### Expected behavior
I would expect that when doing autolayout on a BDD that is goes from top to bottom with the main part on the top and the parts contained within to be below. Something like this:
![image](https://user-images.githubusercontent.com/10014976/222927739-d176f83b-91a3-4785-89fa-329556fa79fe.png)

### OS
- [X] Linux (Please put in notes the specific distro)
- [ ] macOS
- [ ] Windows

### Version
Version of Gaphor: main

amolenaar:
This may be a tricky one. The direction in which the diagram is rendered depends on the direction (head, tail) of the association. If you would've added the association the other way around, you would probably end up with the desired result.

What we can do is check for aggregation type on associations.

https://github.com/gaphor/gaphor/blob/main/gaphor/plugins/autolayout/pydot.py#L272

danyeaw:
Closed by #3479.

PR #3479, opened 2024-09-06
Fixed drop of ObjectFlow, ControlFlow, Pin; Fixed autolayout

pbrown12303:
As reported in Issue #3477 dropping an existing ObjectFlow or ControlFlow onto an activity diagram does not automatically connect the flows to the end-point elements in the diagram. This change fixes that problem.

As reported in issue #3481, dropping a Pin does not automatically connect it with its owning action.

As reported in issue #2126, autolayout inverts associations, control flows, and object flows. 

### PR Checklist

### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: #3477 

### What is the new behavior?

When a ControlFlow or ObjectFlow is dropped on the diagram the diagram item automatically connects to the head and tail items already in the diagram.

When a Pin is dropped on the diagram it is automatically connected to its owning action

- [ ] Yes
- [x] No


gaphor/diagram/drop.py
```diff
@@ -119,6 +120,27 @@ def drop_relationship_on_diagram(element: Relationship, diagram: Diagram, x, y):
     )
 
 
+@drop.register(ActivityEdge, Diagram)
```
danyeaw, line 123: Hi @pbrown12303, thanks so much for the contribution and I really like that you included tests to go along with it!

Since this new function almost exactly repeats the one above it, did you consider stacking the `@drop.register` decorators? You should be able to put one on top of the other. If you then update the type hint for the `element` function argument to be `Relationship | ActivityEdge` it should be able to take either type of element as an input to the function and simplify the code to use just a single function.

danyeaw (changes requested):

gaphor/diagram/drop.py
pbrown12303: Dan,

I thought of that, but when the Dependency is selected in the tree (as
opposed to the diagram) there is no DependencyItem, which is diagram
specific. There are DiagramItem properties (i.e. the Automatic setting)
displayed in the DiagramItem properties ui. For this reason I had to
separate them.

danyeaw:
Hi @pbrown12303, could you please help me understand that more, aren't you creating a new drop function and registering to be called when an item is dropped? What is the relationship between that and the DiagramItem properties?

pbrown12303:
I tried combining the two types in the type hints for the register - it doesn't work.

pbrown12303:
Figured out how to register the same function for two different types - you need to use two ```@drop.register``` calls.

danyeaw (approved):
Hi @pbrown12303, these changes look great, and thanks for the contribution! In the future, it would be better to create a branch off of main for each issue you are fixing and then open separate pull requests for each topic. It makes it easier to review if unrelated changes are separated.

danyeaw:
@all-contributors please add @pbrown12303 for bug, code

PR #3509, opened 2024-09-27
Remove workaround for opening links on macOS

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

About dialog has a workaround for opening links on Darwin. It did not work before.

Issue Number: N/A

### What is the new behavior?

Links work in a (to be released) GLib version, not to far in the future, I hope.

See https://gitlab.gnome.org/GNOME/glib/-/issues/3484.

### DIFF
--- a/gaphor/ui/help/__init__.py
+++ b/gaphor/ui/help/__init__.py
@@ -4,9 +4,8 @@
 """
 import logging
 import sys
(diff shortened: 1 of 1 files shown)


amolenaar:
Fixed in GLib 2.82.2.

PR #3511, opened 2024-09-27
Allow to import Gtk from gaphor.ui

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

GTK version warning when starting Gaphor.

Issue Number: N/A

### What is the new behavior?

By moving the service to a module, instead of the `__init__.py` file, the GTK dependency is only needed when the (GUI) service is loaded.

- [ ] Yes
- [ ] No


danyeaw (approved):
Thanks @amolenaar!

PR #3524, opened 2024-10-08
Fix macOS release upload

danyeaw:
When trying to release version 2.27.0, we are getting an error when trying to upload the macOS release assets to the release. This is because we need to pass the release version to the `gh upload` command, but the version is currently mixed with the architecture suffix. This PR splits the version and the arch to two separate inputs so that we can keep the version to the actual version number and use the arch as a separate suffix where needed.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: N/A

- [ ] Yes
- [X] No

### DIFF
--- a/.github/actions/macos_dmg/action.yml
+++ b/.github/actions/macos_dmg/action.yml
@@ -4,6 +4,9 @@ inputs:
   version:
     description: 'Gaphor version number'
     required: true
+  arch:
+    description: 'macOS architecture'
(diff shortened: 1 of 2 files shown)


amolenaar (approved):
This looks a lot nicer indeed.

amolenaar:
The RtD build fails because GNOME sites are temporarily offline.

PR #3525, opened 2024-10-08
Add Association Subset Capability

pbrown12303:
Factored subset logic into new ```subsettable_umlproperty``` class; Updated ```association``` class to handle subsets.

This pull request enables subsetting of associations. I factored the subset logic into a new subclass of ```umlproperty``` (```subsettable_umlproperty```) and made it the parent class of all subsettable properties, including ```association```. I updated the ```association``` class to coordinate the subset contents. The documentation of association details the logic governing the superset/subset contents under various editing scenarios. Some of the logic is not determined solely by standard set theory, but is rather based on policy. Please examine these policies to see if they make sense. 

The test cases cover all of the cases described and verify that a subsetted association still works correctly as a subset of a derived union.

One of the checks I added in ```subsettable_umlproperty.add()``` is to ensure that a subclass with multiplicity of many is not allowed as a subclass of an association with a multiplicity of 1. I originally raised an error here, but that created its own problem.  The ```add()``` method is called during initialization when the ```uml.py``` file is read in, which would be appropriate except for the fact that the execution of ```poetry run poe uml``` also loads uml.py in its startup. If there is a subsetting error in the model (which there was - see below), then poe cannot generate a new uml.py file. In other words, there is no mechanism for fixing the error. Consequently, instead of raising an error, I log a warning that will appear when gaphor runs.

As mentioned, there was a modeling error in the existing UML model. The ```BehavioralFeature.ownedParameterSet``` was modeled with an upper multiplicity of many, but is declared to be a subset of ```namespace``` which has a multiplicity of 1. This was caught in the updated add() function. I changed the multiplicity to 1. That is the only change to the UML model. 

### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: N/A

- [ ] Yes
- [x] No


amolenaar:
This looks good. I'll have to give it a spin 😃

danyeaw:
Hi @pbrown12303 it looks like we are going to need to resolve the UML.gaphor conflict before we merge this one.

amolenaar:
I merged the changes from this branch with the main branch.

Since `Dependency` has moved from the core model to UML, I re-applied the changes there.

amolenaar (approved):

PR #3529, opened 2024-10-12
Remove custom macOS keybindings

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: N/A

### What is the new behavior?

Before GTK 4.16, there we're no native bindings for macOS.

With GTK 4.16 released a while back, GTK has proper keybindings on macOS, so we do no longer have to maintain our own.

- [ ] Yes
- [x] No


danyeaw (approved):
Nice, great that this moved upstream!

PR #3530, opened 2024-10-12
Separate New Base Type for KerML

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Currently our base type is Element. This class contains basic functionality _as well as_ relationships.

Issue Number: N/A

### What is the new behavior?

By separating the mechanism (`Base`) from the base class we use in our models (`Element`), we open the door for other element base types. This should help is implement models such as KerML/SysMLv2 from scratch, following the semantics of the model.

To proof this works, I changed the base type of `StyleSheet` to `Base`. Now the style sheet is not having any of the unused attributes and relationships anymore.

- [ ] Yes
- [x] No

### Other information

At some point we may also want to make `Presentation` inherit from Base directly.


danyeaw (approved):
Hi @amolenaar, I really like this approach to separate the base type like this!

Issue #3468, opened 2024-08-28
For a Dependency, display the client and supplier in the Property Editor


pbrown12303:
### Is your feature request related to a problem?
After creating a Dependency using the API there is no way to determine the supplier of the dependency

### Describe the solution you'd like
When a Dependency is selected, display the client and supplier values in the Property Editor

amolenaar:
Hi @pbrown12303,

This sounds like a good feature. For associations we have this to, since you can edit the association ends.

amolenaar:
Fixed by #3538.

Issue #3458, opened 2024-08-21
Classifier.isFinalSpecialization missing from gaphor model


pbrown12303:
### Describe the bug
The UML property Classifier.isFinalSpecification is missing from the gaphor model.

### To Reproduce

If possible, add the error report from `≡` (top-right menu) → `Tools` → `Error Reports`.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [ ] macOS
- [x] Windows

### Version
Version of Gaphor:

amolenaar:
Hi @pbrown12303,

I cannot find a reference to `isFinalSpecification` in the [UML specs](https://www.omg.org/spec/UML/2.5.1/PDF).

pbrown12303:
Section 9.2.2, Figure 9.1.

pbrown12303:
And sorry, I made a typo. Its isFinalSpecialization, not Specification.

PR #3538, opened 2024-10-16
Fixed property pages for Dependency; added isFinalSpecialization

pbrown12303:
Added property pages for Dependency that work when the dependency is selected from the tree. This required factoring the property pages to separate DependencyItem-specific properties from the properties of the model Dependency. Issue #3468

Added Classifier.isFinalSpecialization property as per UML specification. Issue #3458 

### PR Checklist

- [ ] I have read, and I understand the GNOME [Code of Conduct](https://conduct.gnome.org/)

### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: #3458, #3468

- [ ] Yes
- [x] No


gaphor/UML/classes/dependencypropertypages.py
```diff
@@ -10,8 +10,50 @@
 from gaphor.UML.classes.dependency import DependencyItem
 
 
-@PropertyPages.register(DependencyItem)
+@PropertyPages.register(Dependency)
```
danyeaw, line 13: Hi @pbrown12303, could you explain more why having a Property Page unique for the Dependency is needed? I think for the other relationships and elements we only have a Property Page for the item.

danyeaw (commented):
The isFinalSpecialization looks great, thanks for the contribution!

I am still trying to understand the property pages for the Dependency. For Association, when you expand the relationship in the tree you can see the Head and Tail elements. Is that what you are looking for, or do you think each element in the tree needs its own property page as well?

gaphor/UML/classes/dependencypropertypages.py
pbrown12303: The split is only necessary when the you  want to display the properties that are owned by the diagram element (the item) in addition to those of he subject model element. The need for the split into two property pages becomes apparent when a subject model element is selected in the tree. If you try and use a property page that shows the diagram element property, you'll get an error because the subject model element does not have such a property. 

My recommendation would be to think of the intent of the current property page to be a display of just the subject model element's properties and name it accordingly (which I think is by and large the case right now). Only when you want to additionally display properties belonging specifically to the selected diagram element item would you need a second property page specific to the diagram element's item properties. 

In this particular case, the property page (before my change) was showing only diagram element item properties. No properties of the subject diagram element were shown. In keeping with the thinking I just described, I re-named the existing property page to indicate that it is item specific and created a new property page to display the subject element properties.

danyeaw: @amolenaar, since you are much more the expert on how the PropertyPages should work, what do you think, do you we need a separate PropertyPage to be able to show the Source and Target?

pbrown12303: Done

amolenaar:
Looks really nice. Thanks @pbrown12303.

PR #3546, opened 2024-10-23
macos: Fix Gaphor model file conformance

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

### What is the new behavior?

According to the apple docs, the conformance type should be some sort of super-type for our Gaphor model.

Since our models are saved as XML, I suppose `public.xml` is the most approprate.

https://developer.apple.com/documentation/uniformtypeidentifiers/defining-file-and-data-types-for-your-app?language=objc

- [ ] Yes
- [x] No


danyeaw (approved):
Thanks @amolenaar!

Issue #3540, opened 2024-10-17
Dispense with Core in its present form: Its use prohibits compliance with the UML specification.


pbrown12303:
### Is your feature request related to a problem?
Splitting the UML model concepts between Core.gaphor and UML.gaphor makes it impossible to comply with the UML language specification. 

A specific case in point is the modeling of Dependency (issue #3528, pull request #3525). Shown below is the UML specification for dependency. In Gaphor, Dependency is defined in Core.gaphor, while PackageableElement is defined in UML.gaphor. This makes it impossible to make Dependency a subclass of PackageableElement. This, in turn, makes it impossible to set the owningPackage property of a Dependency (it doesn't have one because it is not a subclass of PackageableElement). This is but one consequence of trying to split the model.

![image](https://github.com/user-attachments/assets/bb6697e7-3c12-4c7e-a578-586c7a6cc6d5)

### Describe the solution you'd like
Migrate all language-specific modeling concepts from the Core model into language-specific models. If common cross-language concepts are desired for tool purposes, make them base classes (e.g. BaseElement) that can be made sub-classes of the language-specific modeling concepts.  With this approach the language models are not constrained in any way and the base classes can support tool-specific behavior.

### Describe alternatives you've considered
There are no alternatives is compliance with language specifications is sought.

danyeaw:
With #3541 merged, are we ready to close this one?

pbrown12303:
Not quite. The UML classes in Core.gaphor need to be moved to UML.gaphor first. I believe @amolenaar is in the process of doing that as part of the preparation for multiple language support.

amolenaar:
Those PR's have been merged now.

All we can do is move  the `Relationship` class back to the UML model (and maybe `Comment`).

We need `Element` in the core model, since `Diagram` also depends on it.

Not sure what to do with `Comment`: it's been moved to the core model and is displayed as part of the "general" tools, alongside lines and boxes.

The next big improvement will be integrating #3525.

pbrown12303:
If you put Element in Core the target language will not be able to create any new associations with Element. This, I think, would be a fatal flaw.

[...]

PR #3336, opened 2024-06-01
Diagram stereotypes

amolenaar:
### PR Type
What kind of change does this PR introduce?

- [ ] Bug fix
- [x] Feature
- [ ] Chore (refactoring, formatting, local variables, other cleanup)
- [ ] Documentation content changes

### What is the current behavior?

You can set stereotypes on diagrams.

Issue Number: N/A

### What is the new behavior?

Show stereotypes on diagrams.

image

### Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #3555, opened 2024-11-01
Move Relationship and Comment back to the UML model

amolenaar:
Move `Relationship` and `Comment` back to the UML model

Saved models have been updated by means of search/replace, no migration has been made.

Comment items have been moved to the UML model, as well as the "general tools" section.

### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: #3540

### Other information

This PR puts the model in a bit of a skew: `Element` is still in core, but [some its properties](https://github.com/gaphor/gaphor/blob/empty-core-model/gaphor/core/modeling/element.py#L43-L53) are placeholders for relations defined in the UML model.

This used to be (partly) solved by having `Relationship` and `Comment` as part of the core model. The only odd one out has always been `appliedStereotype`.

One other thing that's been bothering me is that `Element` is a model type that can be fully generated. It's still a hand written type, since `Diagram` depends on it.

If we proceed, it will not be possible to link comments to diagrams. Not sure if that's a problem, though. We also loose the ability to apply stereotypes to diagrams (#3336), which would be a pity.


danyeaw (approved):
Hi @amolenaar, these changes look great, thanks!

amolenaar:
Now we have to figure out how to deal with the Element class.

I have an idea: create generic functions (singledispatch) for the features we need for the model browser. Then we do no longer need Element in the core model.

PR #3567, opened 2024-11-14
Make model browsers ownership concept extendible

### LINKED ISSUE #3557: Properties remain in model browser after deletion
### Describe the bug

When an association is removed, properties owned (or at least a member) of the class they refer to are still present in the model browser.

It looks like removal from the model works fine: when the model is reloaded, the properties no longer show

### To Reproduce
1. create 2 classes
2. create an association and make both end navigable
3. remove the association
4. Note that the properties are still visible in the model browser. Interacting with those elements raises exceptions.

Selecting the "ghost" element raises an exception:

```
gaphor.ui.elementeditor ERROR Could not construct property page for TypedElementPropertyPage
  File "/Users/arjan/Development/gaphor/gaphor/ui/elementeditor.py", line 212, in create_pages
    page = adapter.construct()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/arjan/Development/gaphor/gaphor/UML/propertypages.py", line 40, in construct
    model = list_of_classifiers(typed_element.model, UML.Classifier)
                                ^^^^^^^^^^^^^^^^^^^
  File "/Users/arjan/Development/gaphor/gaphor/core/modeling/base.py", line 89, in model
    raise TypeError(
TypeError: Can't retrieve the model since it's not set on construction
gaphor.ui.elementeditor ERROR Could not construct property page for StereotypePage
  File "/Users/arjan/Development/gaphor/gaphor/ui/elementeditor.py", line 212, in create_pages
    page = adapter.construct()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/arjan/Development/gaphor/gaphor/UML/profiles/stereotypepropertypages.py", line 33, in construct
    stereotypes = UML.recipes.get_stereotypes(subject)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/arjan/Development/gaphor/gaphor/UML/recipes.py", line 125, in get_stereotypes
    model = element.model
            ^^^^^^^^^^^^^
  File "/Users/arjan/Development/gaphor/gaphor/core/modeling/base.py", line 89, in model
    raise TypeError(
TypeError: Can't retrieve the model since it's not set on construction
```

### Expected behavior

The Model Browser should represent the actual state of the model.

### Version
Version of Gaphor: main

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

The model browser mainly depends on the `Element.owner/ownedElement` relationship for ownership.
 It also supports `memberNamespace`, mainly for association ends without owner.

Builds on #3568.

Issue Number: #3540

Fixes #3557

### What is the new behavior?

Separate the concept of ownership out of the model browser, so it can be extended. 

The `owner()` dispatch method tells who's the owner of an element: another element, `Root`, or `None`. In case of `None`, the element is not supposed to be shown in the model browser.

- [x] Yes
- [ ] No

A separate element ownership function.

### Other information

The only issue is that event handling (in `ModelBrowser`) still checks for the properties used. This should change at some point, but I think this is already a nice simplification.

I do love to get rid of the `owns()` dispatch function -- basically the counterpart of `owner()` -- but I have no idea how, yet.

### DIFF
--- a/gaphor/core/format.py
+++ b/gaphor/core/format.py
@@ -6,17 +6,17 @@
 
 from functools import singledispatch
 
-from gaphor.core.modeling import Diagram, Element
+from gaphor.core.modeling import Base, Diagram, Element
 
 
 @singledispatch
-def format(el: Element, **kwargs) -> str:
+def format(el: Base, **kwargs) -> str:
     """Format an element."""
     raise TypeError(f"Format routine for type {type(el)} not implemented yet")
 
 
 @singledispatch
-def parse(el: Element, text: str) -> None:
+def parse(el: Base, text: str) -> None:
     """Parse text and update `el` accordingly."""
     raise TypeError(f"Parsing routine for type {type(el)} not implemented yet")
 
--- a/gaphor/diagram/deletable.py
+++ b/gaphor/diagram/deletable.py
@@ -1,9 +1,9 @@
 from functools import singledispatch
 
-from gaphor.core.modeling import Element
+from gaphor.core.modeling import Base
 
 
 @singledispatch
-def deletable(element: Element) -> bool:
+def deletable(_element: Base) -> bool:
     """Determine if a single element can safely be deleted."""
     return True
--- a/gaphor/ui/modelbrowser.py
+++ b/gaphor/ui/modelbrowser.py
@@ -7,6 +7,7 @@ from gaphor.abc import ActionProvider
 from gaphor.action import action
 from gaphor.core import event_handler
 from gaphor.core.modeling import (
+    Base,
     DerivedAdded,
     DerivedDeleted,
     DerivedSet,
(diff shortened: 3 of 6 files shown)


amolenaar:
Changes to the model come later. Once the concept of ownership is separated from the model browser, I can start implementing different variations: two for UML, one for diagrams.

In the end I plan to move the Element class to the UML model. `Diagram` will probably have a uni-directional relation to a parent. `Picture` can go away, since we have `Image` in the UML model already.

> How do you deal with a tree drag-and-drop that moves an element to another location in the tree?

I have to figure out parts of that still. I think I need something like a service that notifies the model browser when an element's owner has changed.

The ultimate goal would be to make the model browser modeling language agnostic. But that'll take several PR's to achieve.

danyeaw (approved):
Thanks @amolenaar, these changes look great!

PR #3568, opened 2024-11-16
Replace Core Picture by UML Image

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

In the Core we have a picture object. Instead of using that, we can use `UML.Image` and make the core package a bit smaller.

It makes sense, IMHO, since Comment has also moved back to the UML model.

Issue Number: #3540

### What is the new behavior?

Use `UML.Image` instead of `Core.Picture` for pictures.


danyeaw (approved):
Thanks @amolenaar, this is a nice update!

When I merged the other PR, it pushed this merge commit automatically which I was surprised by. Did we configure that on, or is that being done by one of our workflows?

amolenaar:
That's correct. I built the other MR on top of this. It was the only way to break the updates into smaller pieces.

PR #3571, opened 2024-11-17
Move Element class to UML model

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

`Element` was part of the Core model. This made it sort of a prerequisite for any other modeling language, including SysML2.

Issue Number: #3540 

### What is the new behavior?

The Element class and its relations have been moved to the UML model. Now the core model contains only some core concepts: Diagrams, change sets.

To make it work, Diagram is inherited in the UML model and enriched with Element as a base class. That way it can be placed nicely in the model browser and keeps the ability to be stereotyped.

- [x] Yes
- [ ] No

### Other information

* Many of the changes have to do with that `Presentation.subject` points to `Base`, instead of `Element`.
* In a lot of places we have to use `UML.Diagram`, instead of `Core.Diagram`.
* name and qualified name are still part of Element (not NamedElement). This is something I want to solve in a future PR. 
* A few architecture rules are xfailed, this is also future work.


danyeaw (approved):
These changes look great, thanks @amolenaar!

PR #3582, opened 2024-11-23
Reduce dependency on UML model in the code base

amolenaar:
### PR Type
What kind of change does this PR introduce?

- [ ] Bug fix
- [ ] Feature
- [x] Chore (refactoring, formatting, local variables, other cleanup)
- [ ] Documentation content changes

### What is the current behavior?

Follow up of #3571.

Reduce the blast radius of the UML model in the code base.

This is important so, we can support other modeling languages alongside UML.

Issue Number: #3540, #3577 

### Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

Issue #3577, opened 2024-11-20
Break import cycles


amolenaar:
### Describe the bug

In some places, imports can cause cycles. E.g.

```
ImportError: cannot import name 'drop' from partially initialized module 'gaphor.diagram.drop' (most likely due to a circular import)
```

Those are basically violations of the architecture, and should be fixed at some point.

(this is partly due to historical reasons, when we only had one UML model)

### To Reproduce
1. `python -c "import gaphor.diagram.drop"`

### Expected behavior

The architecture should be more strongly enforced for avoid those issues.

### Version
Version of Gaphor: >=2.27

### Additional information

I found the following files are in violation:

- [x] gaphor/diagram/drop.py
- [x] gaphor/diagram/general/diagramitem.py
- [x] gaphor/services/undomanager.py (#3577)

Those have to do with diagrams, now that `gaphor.UML.Diagram` is the one used in UML models.

- [ ] gaphor/diagram/diagramtoolbox.py
- [ ] gaphor/diagram/painter.py

Ultimately, the UI should not depend on the UML model as well:

- [x] gaphor/ui/filemanager.py -> Only used for generating the default model (#3632)
- [x] gaphor/ui/modelbrowser.py -> Mode event handlers and UML code blocks to a separate (UML specific) service? That service can provide the tree model. Can we manage multiple services, so we can have a separate one for SysMLv2?
- [x] gaphor/ui/treemodel.py -> move to `gaphor.UML`?
- [x] gaphor/ui/modelmerge/organize.py

amolenaar:
The last bit is solved by #3649.

PR #3606, opened 2024-11-27
Fix undo order

amolenaar:
This PR builds on #3582

### PR Type

- [x] Bug fix
- [ ] Feature

The undo order is not always right for presentation elements.

Issue Number: #3577

### What is the new behavior?

Diagrams will always be removed after all presentations are removed. This way the undo order is maintained.

### Other information

This opens the opportunity to have different types of diagrams. It could be that we create separate `ClassDiagram`, `SequenceDiagram`, etc. types, instead of using the diagramType objects we currently use.

### DIFF
--- a/gaphor/core/modeling/base.py
+++ b/gaphor/core/modeling/base.py
@@ -78,7 +78,6 @@ class Base:
         # The model this element belongs to.
         # NOTE: Will be unset by ElementFactory once it's `unlink()`ed.
         self._model = model
-        self._unlink_lock = 0
 
     @property
     def id(self) -> Id:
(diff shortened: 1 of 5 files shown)


tompkins-ct:
I like this 
> This opens the opportunity to have different types of diagrams. It could be that we create separate ClassDiagram, SequenceDiagram, etc. types, instead of using the diagramType objects we currently use.

I think most digrams are fine, but sequence-like diagrams are difficult in Gaphor because they have a pretty defined layout and Gaphor provides too much flexibility in placing items for that particular diagram. A more structured foundation for them might be beneficial, maybe something structured like Eclipse/Capella in the future? This typing also enables tuning autolayout tools to the diagram type which can produce much better results.

danyeaw (approved):
Nice update @amolenaar!

Issue #3605, opened 2024-11-27
Many macOS 13 (intel) builds fail


amolenaar:
### Describe the bug

It's noticable that a lot of macOS 13 builds are failing.

I see mainly 2 kinds of errors:

1. Errors [during self-test](https://github.com/gaphor/gaphor/actions/runs/12004414436/job/33459398746)
1. `Illegal instruction` errors [during unit testing](https://github.com/gaphor/gaphor/actions/runs/11700255329/job/32583924141)
2. Errors in [creating the DMG](https://github.com/gaphor/gaphor/actions/runs/12000352110/job/33449370761)

```
 DIHLDiskImageCreate() returned 49168
hdiutil: create: returning 49168
hdiutil: create failed - Resource busy
```

Is there something we can do about this? It looks like the workaround (killing XProtect) doesn't help much.

What can we do to make the macOS 13 build more stable?

Test failures:

```
Current thread 0x00007ff852cf1940 (most recent call first):
gaphor/ui/tests/test_main.py .
  File "/Users/runner/work/gaphor/gaphor/.venv/lib/python3.13/site-packages/gi/overrides/Gio.py", line 42 in run
  File "/Users/runner/work/gaphor/gaphor/gaphor/ui/tests/test_main.py", line 27 in fake_run
  File "/Users/runner/work/gaphor/gaphor/gaphor/ui/__init__.py", line 107 in run
  File "/Users/runner/work/gaphor/gaphor/gaphor/main.py", line 117 in run
  File "/Users/runner/work/gaphor/gaphor/gaphor/main.py", line 26 in main
  File "/Users/runner/work/gaphor/gaphor/gaphor/ui/tests/test_main.py", line 46 in test_application_startup_with_model
```

```
Current thread 0x00007ff84a01e940 (most recent call first):
gaphor/ui/tests/test_selftest.py 
  File "/Users/runner/work/gaphor/gaphor/.venv/lib/python3.13/site-packages/gi/overrides/Gio.py", line 42 in run
  File "/Users/runner/work/gaphor/gaphor/gaphor/ui/__init__.py", line 107 in run
  File "/Users/runner/work/gaphor/gaphor/gaphor/ui/tests/test_selftest.py", line 10 in test_self_test
```

danyeaw:
Upstream issue for the problem creating DMGs: https://github.com/actions/runner-images/issues/7522

amolenaar:
[...]

PR #3612, opened 2024-11-28
Build macOS DMG with dmgbuild

amolenaar:
Let's see if this tool performs better.

### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: #3605

- [ ] Yes
- [x] No


danyeaw (approved):
Nice idea! The approach to build the DMG using Python instead of the create-dmg shell script should hopefully be much more reliable.

amolenaar:
Let's see how it works out. We've used dmgbuild in the past and swapped it out for create-dmg.

PR #3617, opened 2024-11-29
Dependency should reside in a package

### LINKED ISSUE #3467: Dependency Ownership Inconsistent with UML Specification
### Describe the bug
The Gaphor implementation of Dependency (and its sub-classes) makes the client the owner of the dependency. In the UML specification a Dependency is a subclass of PackageableElement, whose owner must be a Package.

### To Reproduce
Here's the Gaphor definition:
image

Here's the UML Specification:
image

Note that in the Gaphor implementation, the owner property is derived from the client property.

### Expected behavior
According to the UML specification, the Dependency must be owned by a Package. This makes some sense, as the Dependency (or its subclasses may be declared in other contexts and not necessarily be a part of the NamedElement.

In MagicDraw, the default behavior when creating a Dependency (or any of its subclasses) in a diagram is to make the package owning the diagram the owner of the Dependency.

### OS
- [ ] macOS
- [x] Windows

### Version
Version of Gaphor: 2.26.0

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: Fixes #3467

### What is the new behavior?

Dependencies are owned by a package.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/UML/classes/classconnect.py
+++ b/gaphor/UML/classes/classconnect.py
@@ -31,6 +31,11 @@ class DependencyConnect(DirectionalRelationshipConnect):
         )
         self.line.subject = relation
 
+        if not relation.owningPackage:
+            assert isinstance(self.diagram, UML.Diagram)
+
+            relation.owningPackage = owner_package(self.diagram.owner)
+
     def update_dependency_type(self, handle):
         line = self.line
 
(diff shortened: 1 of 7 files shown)


danyeaw (approved):
Thanks @amolenaar!

PR #3620, opened 2024-12-01
Only show status window on save for big files

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

An issue arose with the Adwaita dialogs: if a dialog is closed to quickly, the close call is ignored and the dialog remains visible. 

Issue Number: N/A

### What is the new behavior?

This avoids flicker when smaller files are saved.

A workaround is in place for issue https://gitlab.gnome.org/GNOME/libadwaita/-/issues/970.

Small update for setting up a local Gaphor dev environment on Fedora.

- [ ] Yes
- [x] No


danyeaw (approved):
Thanks @amolenaar!

PR #3627, opened 2024-12-02
Move editing settings to global preferences window

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Settings for tool reset and automatic model maintenance used to be configured on a per-model basis.

Issue Number: N/A

### What is the new behavior?

Settings moved from a model-based to global settings. This should make our settings story more consistent. Now all we have

- [ ] Yes
- [x] No

### Other information

I think we discussed this idea a while back as part of Mareike's study on MBSE tooling for neurodivergent people.

NB. Do not forget to run `gaphor install-schemas`.


danyeaw (approved):
Hi @amolenaar, thanks! This is a nice simplification to move these to the main program settings.

PR #3668, opened 2024-12-16
Fixed bug in dropping Generalization

pbrown12303:
The logic for drop assumed that the relationship would have a multiplicity of many (>1) at both ends. Generalization has an upper bound of 1 on each end. Added logic to drop to handle the case when the upper bound is 1.

### PR Checklist

### PR Type

- [X] Bug fix
- [ ] Feature

Issue Number: N/A

- [ ] Yes
- [X] No


amolenaar:
Thanks @pbrown12303.

I ran into another issue: `drop()` is expected to return only a single presentation. I fixed the method to return on the first match. Also added (and fixed) some type annotations.

Issue #3549, opened 2024-10-28
Improve diagram layouting


amolenaar:
### Is your feature request related to a problem?

Currently we have an half-baked auto layout function based on GraphViz. Most diagram types produce okay results, but diagrams like sequence diagrams are not supported.

I picked Graphviz, since it's the most common one and the results are predictable.

### Describe the solution you'd like

I came across http://blockdiag.com by accident. It looks like a good fit for our problem. Even more since everything is in Python, so it should be a lot easier to integrate: no intermediate text files.

Downside is that the code has not seen any love in around 3 years.

### Describe alternatives you've considered

Stick with Graphviz?

---
This issue is a follow-up of #1653.

An incomplete of diagram layout libraries for Python is available at https://github.com/gaphor/gaphor/issues/615#issuecomment-1054236210.

tompkins-ct:
Improved diagram autolayout would be nice.

I think we'd want a couple capabilities to facilitate easier use
1. separate engines based on diagram type  (selectable or auto)
    1. Sequence / swimlanes probably need a different layout engine / settings.
    2. block diagrams showing functional flow vs block diagrams showing taxonomy (tree / hierarchical)
1. clean up connections only vs blocks as a whole
2. only layout selected items (limit scope of the layout engine)


I think we have a couple options here. 
1. Maybe take a look at blockdiag and re-implement its layout engine in gaphor (it's layout engine might be graphviz, but I didn't look).
    1. We may be able to improve graphviz for specific diagrams by adjusting some of the layout parameters.
3. Use the eclipse layout kernal (ELK). (https://github.com/kieler/elkjs / ). This may require some import / export work to interface with it but it's got layout engines suitable for UML/sysML type diagrams. It looks like it can import a graph via JSON or custom formats and support laying out ports and hierarchical nested systems. 
4. Implement a pure-python layout engine in gaphor (or a new library). Right now, it's doesn't look like there are a lot out there outside of graphviz. However, the algorithms on a lot of the java / js libraries are easy to access we can also take a look at journal articles for references.

amolenaar:
We can now easily separate the layout engine based on diagram type (#3649).

I do not feel like reaching out to Java/nodejs applications. They're very hard to package in the installers we provide for Windows and macOS.

Re-implementing is a huge effort, though.

tompkins-ct:
[...]

PR #3690, opened 2024-12-20
Elk layout engine test

tompkins-ct:
Demonstration of the ELK layout engine in gaphor.

### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Currently using graphviz DOT layout engine which doesn't seem to provide a lot of support for node definitions.

Issue Number: #3549

### What is the new behavior?
Uses [ELK layered layout engine](https://eclipse.dev/elk/reference/algorithms/org-eclipse-elk-layered.html) which may be better suited to heirarchical graphs with placement of labels, ports and edges.

ELK also provides an edge routing algorthm via `libavoid` which would be a nice addition to gaphor. I haven't tested that, but it may call on additional dependencies.

- [ ] Yes
- [x] No

### Other information
- Currently doesn't support edge-to-edge connections (e.g. comment lines referencing an edge)
- Not sure how to package node dependencies.


amolenaar:
@tompkins-ct Thanks for the update. I think we can still get more out of our current (graphvviz) layout engine, but do not really have time to dig into it. 

I'd like to add your project to the website, if you're okay with that.

tompkins-ct:
@amolenaar yeah, I think that would be good.

PR #3696, opened 2024-12-28
macOS menubar

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

We have the GNOME hamburger menu. Although this looks nice, on macOS, we can put those items in the menu bar too.

While working on https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7906, I wanted to see how it would look if we'd integrate the menu bar in Gaphor.

Builds on #3780.

### What is the new behavior?

Gaphor has a menu bar on macOS.

It's behind a feature flag, since it needs GTK 4.17.2, with my MR merged.

```sh
GAPHOR_FEATURE_FLAG=menubar gaphor models/Core.gaphor
```

image

### Other information

* Both `app.` and `win.` actions have their shortcuts (accelerators) set on app level. This way the shortcuts show up in the menu.
* Changed when `DiagramClosed` event is emitted.
* Disable layout and export actions if no diagram is open.
* The hamburger menu is generated from the ui file.
* Add maximize window action.
* Some cleanup around action groups and shortcuts.
* Standardized action names a bit: (un)select diagram items with `selection.select-all`, `selection.unselect-all`, `text.undo`, `text.redo`.
* Some items need an accelerator defined at app level, so it shows in the menu.
* Special handling for `text.undo|redo`: This action is defined in the menu and should be used for our Undo manager, as well as text entry undo/redo. A new service is introduced that checks if the action is applicable on the focused widget, if not, delegate to window level. This is similar to how macOS menu items are handled.


danyeaw (approved):
This looks like a much more native macOS experience. Nice job upstream to implement this!

amolenaar:
With the menu bar in place, undo/redo in text fields is broken: instead of using the widget's native undo/redo, it now relies on the UndoManager service. Therefore it would be good to have a look at how we can make our undo manager service work with text undo/redo.

danyeaw:
Hi @amolenaar, I noticed that Ctrl+Z stopped working on Linux with this update.

amolenaar:
@danyeaw Undo/redo shortcuts work again They had to be registered in `gaphor/ui/__init__.py`.

danyeaw (approved):
Thanks @amolenaar! This is a really nice capability you added.

rodrigopedra:
~Is exporting the menu to other Linux's DE which supports global menu (e.g. Plasma) planned?~

~If not, would a PR adding such support be desirable?~

~Thanks in advance~

---

EDIT: looking into the companion MR to GTK which resulted in this PR (https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7906/diffs), it seems support was added in GTK only to macOS.

So it seems exporting the menu on Linux would not be an easy task without upstream support.

Never mind then. Thanks anyway.

PR #3709, opened 2025-01-06
Update Poetry to version 2.0.0

danyeaw:
Updates to Poetry and upgrades to the pyproject.toml to using the project metadata.

I also learned that Poetry is not quite ready to use the optional-dependencies as part of the project metadata. I was seeing some inconsistent behavior with what was being installed and removed, and the `--only` and `--with` keywords no longer work.

I also fixed an issue that happened to be uncovered by this update. With the Wheel version of our Linux workflows, meson started to find the wrong Python version (version 3.13 instead of the matrix version), so I added a step to set the matrix version as the main alternative.

Finally I upgraded the readthedocs runner we were using to the latest LTS and Python 3.13.

### PR Checklist

### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: N/A

- [ ] Yes
- [X] No


danyeaw:
@amolenaar, Snyk is telling me that I am not a member of the organization - even though I know you previously added me. Could you please see what's going on?

amolenaar:
This looks really nice. It's good to see we can use the standard PEP layout now.

Do you want to merge it now, or wait for the next round of development (there may be some issues in the .0 release ;) ).

amolenaar (approved):

danyeaw:
My usage of it seems like it is robust with this configuration, I would be good with upgrading it now.

What do you think we should do about Shyk, should we try to export to a requirement.txt for it?

amolenaar:
We can disable Snyk for the time being. We also have dependabot running, which fixes most security issues quickly.

Since other builds are failing due to Poetry 2.0, it would be good to merge this I think. And announce it in the matrix chat, since developers need to upgrade their poetry version.

The error message Poetry 1.8 produces is not very informative.

danyeaw:
Hi @amolenaar, that sounds good. I set the Snyk pyproject.toml integration as inactive.

Issue #3720, opened 2025-01-15
Rethinking properties.association handling of subset changes


pbrown12303:
The current implementation of the properties.association handling of subset changes appears to be inconsistent with the logic of the UML specification. The current Gaphor implementation of the editing of relationships between Property and Association seems to support this idea, but since the current UML implementation does not fully express the subset relations in the UML specification, the problem has not yet manifested itself.

In bringing the Gaphor UML model into alignment with the UML specification, the problem has surfaced. 

### Discussion

Here is the current Gaphor model of the associations linking UML.Property and UML.Association:
![image](https://github.com/user-attachments/assets/007774be-5a50-452c-9dc9-e1beee9f55be)

And here is the UML specification model of these associations:
![image](https://github.com/user-attachments/assets/c8fbce8f-fe8e-413a-a42e-8010bbd7d384)

Take a look at the UML specification of the UML.Association properties related to UML.Property. Shown in set-theoretic terms, these properties are related to each other as follows:
![image](https://github.com/user-attachments/assets/04f768c7-e089-4fa4-b7dc-b5059d263c1d)

After implementing the full UML specification model in Gaphor, some of the built-in tests failed. One, in particular, illustrates the problem. We have to dig through three layers of code for the issue to become apparent.

At the top level, we have the ```test_copy_paste_items_with_connections```, the first few lines of which are as follows:
```def test_copy_paste_items_with_connections(diagram, element_factory):
    gen_cls_item, spc_cls_item, assoc_item = two_classes_and_an_association(
        diagram, element_factory
    )

```
The invocation of ```two_classes_and_an_association``` executes the following code:
``` 
def two_classes_and_an_association(diagram, element_factory):
    gen_cls = element_factory.create(UML.Class)
    spc_cls = element_factory.create(UML.Class)
    gen_cls_item = diagram.create(ClassItem, subject=gen_cls)
    spc_cls_item = diagram.create(ClassItem, subject=spc_cls)
    assoc_item = diagram.create(AssociationItem)
    gen_cls.name = "Gen"
    spc_cls.name = "Spc"

    connect(assoc_item, assoc_item.handles()[0], spc_cls_item)
    connect(assoc_item, assoc_item.handles()[1], gen_cls_item)
    UML.recipes.set_navigability(
        assoc_item.subject, assoc_item.subject.memberEnd[0], True
    )

    assert (
        assoc_item.subject.memberEnd[0]
[...]

Issue #3517, opened 2024-10-01
Unable to establish the owningPackage/packagedElement relationship between Event and Package


pbrown12303:
### Describe the bug
The Package.owningPackage and Package.packagedElement are both derived attributes. The setting of these attribute values is supposed to occur as a side-effect of setting some other attribute value. I have been unable to identify the attribute whose value needs to be set in order to update owningPackage and packagedElement. I have tried setting the following attribute values, all of which produce an error.

```
event.memberNamespace = package
event.namespace = package
event.owner = package
event.owningPackage = package
event.package = package
package.member = event
package.ownedElement = event
package.ownedMember = event
package.ownedType = event
package.packagedElement = event
```

### To Reproduce
1. Programmatically create a Package and an Event (or ChangeEvent)
2. Try setting any of the attributes above. An error message indicating that the attribute is a derived construct will occur.

### Expected behavior
In the UML standard, neither Package.packagedElement nor PackagedElement.owningPackage are derived. They can be set directly. this is the expected behavior.
![image](https://github.com/user-attachments/assets/1bf585fb-48dc-4724-b7fe-e4f841b33693)

### OS
- [ ] Linux (Please put in notes the specific distro)
- [ ] macOS
- [x] Windows

### Version
Version of Gaphor: 2.26.0

pbrown12303:
The same issue exists for Trigger.

amolenaar:
I did a quick check and it looks like Event and Trigger do not have a property that links them to a package.

In our model we do not have the ability to subset from something other than a derived union. Alternatively, the relation `package: Type` should redefine `owningPackage`, instead of being a subset.

PR #3749, opened 2025-02-03
Align with UML Specification - all tests pass

pbrown12303:
This change brings the core structure of Gaphor UML into alignment with the UML 2.5 specification. It does not implement all of the UML specification. There are two significant areas of modification:

- The ```<<simpleAttribute>>``` stereotype was removed from ValueSpecification. This required implementing the Literal subclasses of ValueSpecification, with the exception of LiteralReal. Attributes that previously referenced ValueSpecifications as simple values have been converted to references to ValueSpecifications. This required updating all code that accessed these references. Recipes have been added to facilitate accessing and updating these references. The scope of this change was fairly broad.
- All relationships have been aligned with the UML specification. This particularly impacted parent-child relationships. An update to the property subset logic was required. It should be noted that in a couple of places the UML specification has two different relationships between the same two classes that use the same role name. ```StructuredClassifier <--> Property``` and ```Property <--> Association``` are two examples. In both cases, one of the relationships had to be renamed as Gaphor treats the role names as class property names, which led to ambiguity. 

The storage.py file has been updated to make these changes backwards compatible.

### PR Checklist

- [ ] I have read, and I understand the GNOME [Code of Conduct](https://conduct.gnome.org/)

### PR Type

- [ ] Bug fix
- [X] Feature

Issue Number: #3720, #3517, 

- [ ] Yes
- [X] No


amolenaar:
I merge the main branch into this PR. In the process it looks like I introduced an issue: I cannot set a default value on an attribute anymore.

pbrown12303:
Thanks @amolenaar - I couldn't figure out the git incantation to do this merge on my machine. I'll take a look and fix the problem.

amolenaar:
In also want to make another attempt. 

It's a bit of a pity though that all changes are in one commit. If it were multiple smaller commits, merge conflicts are easier to resolve.

gaphor/C4Model/c4model.py
```diff
@@ -4,6 +4,8 @@
 
 from __future__ import annotations
 
+from decimal import Decimal as UnlimitedNatural
```
amolenaar, line 7: Why use Decimal as UnlimitedNatural?

A natural number is whole number right? If so, we can use `int`: integers in Python grow and do not overflow, similar to `BigInteger` in Java.

gaphor/codegen/coder.py
```diff
@@ -314,17 +323,43 @@ def operations(c: UML.Class, overrides: Overrides | None = None):
...
+                | UML.LiteralUnlimitedNatural
+                | UML.LiteralBoolean,
+            ):
+                defaultValue = UML.recipes.get_literal_value_as_string(a.defaultValue)
+            elif (a.defaultValue[0] == a.defaultValue[-1]) and (
                 a.defaultValue[0] in ['"', "'"]
             ):
                 defaultValue = a.defaultValue
             else:
                 defaultValue = f'"{a.defaultValue}"'
```
amolenaar, line 353: This should be `defaultValue`, not `a.defaultValue`.

gaphor/codegen/coder.py
```diff
@@ -335,11 +370,25 @@ def default_value(a):
 
 
 def lower(a):
-    return "" if a.lowerValue in (None, "0") else f", lower={a.lowerValue}"
+    lowerValue = ""
+    if isinstance(a.lowerValue, UML.LiteralInteger):
+        if a.lowerValue.value is not None and a.lowerValue.value != 0:
+            lowerValue = str(a.lowerValue.value)
+    else:
+        if a.lowerValue is not None and a.lowerValue != "0":
```
amolenaar, line 378:
```suggestion
        if a.lowerValue and a.lowerValue.value is not None and a.lowerValue.value != "0":
```

gaphor/codegen/coder.py
```diff
@@ -335,11 +370,25 @@ def default_value(a):
...
 
 
 def upper(a):
-    return "" if a.upperValue in (None, "*") else f", upper={a.upperValue}"
+    upperValue = ""
+    if isinstance(a.upperValue, UML.LiteralUnlimitedNatural):
+        if a.upperValue.value is not None and not math.isinf(a.upperValue.value):
+            upperValue = str(int(a.upperValue.value))
+    else:
+        if a.upperValue is not None and a.upperValue != "*":
```
amolenaar, line 389:
```suggestion
        if a.upperValue and a.upperValue.value is not None and a.upperValue.value != "*":
```

gaphor/core/modeling/coremodel.py
```diff
@@ -42,6 +44,7 @@ class ElementChange(PendingChange):
 
 class ValueChange(PendingChange):
     property_name: _attribute[str] = _attribute("property_name", str)
+    property_type: _attribute[str] = _attribute("property_type", str)
```
amolenaar, line 47: I'm not sure we need this: the type we can take from the model.

gaphor/core/modeling/properties.py
```diff
@@ -235,8 +237,21 @@ def set(self, obj, value):
                 or self.type
             )
 
-        if self.type is int and isinstance(value, str | bool):
-            value = 0 if value == "False" else 1 if value == "True" else int(value)
+        if self.type is decimal.Decimal and isinstance(value, str):
+            if value == "*" or value == "inf" or value == "Infinity":
+                value = decimal.Decimal(math.inf)
+            else:
+                value = decimal.Decimal(int(value))
```
amolenaar, line 244: Not sure why we need `decimal` here. We deal with natural (whole, positive) numbers.

amolenaar (changes requested):

gaphor/C4Model/c4model.py
pbrown12303: The int type can't handle "*". I used Decimal with conversions in and out. Can't see any other way.

gaphor/codegen/coder.py
pbrown12303: done

gaphor/codegen/coder.py
pbrown12303: Done - made the code cover the math.inf case as well.

gaphor/core/modeling/coremodel.py
pbrown12303: I thought so, too, but In some of the contexts in which the ValueChange is used you don't have the model element.

gaphor/core/modeling/properties.py
pbrown12303: See earlier comment abut UnlimitedNatural. The issue is handling infinite values.

pbrown12303:
This is bizarre. The test that failed on windows (in the teardown) passes when I run it in the debugger in my development environment (which is where and how I usually run the tests), but it fails when I run it without the debugger! I'm looking at it now.

gaphor/C4Model/c4model.py
amolenaar: We can use a union type `int | typing.Literal["*"]` instead.

gaphor/core/modeling/properties.py
```diff
@@ -912,6 +950,8 @@ def __init__(
         self.original: association | derived = original
         self.upper = original.upper
         self.lower = original.lower
+        if opposite is not None:
+            self.original.opposite = opposite
```
amolenaar, line 954: We should probably not change the original opposite relation when a relation is redefined.

I assume the opposite relation should wither point to the original opposite end, or a redefined one.

gaphor/C4Model/c4model.py
pbrown12303: Interesting. Let me try that.

gaphor/UML/recipes.py
```diff
@@ -241,8 +269,8 @@ def create_association(type_a: Type, type_b: Type):
     assoc = model.create(Association)
     end_a = model.create(Property)
     end_b = model.create(Property)
-    assoc.memberEnd = end_a
-    assoc.memberEnd = end_b
+    assoc.ownedEnd = end_a
+    assoc.ownedEnd = end_b
```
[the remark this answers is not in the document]
amolenaar: Changed this: `Association.ownedEnd` is set by the `set_navigability` function. To my understanding, `Property.association` should referenjce the association.

See 3dc474776bbd66c36bc73c6d3be286b5d34790c4

amolenaar (approved):
I think we can merge this.

The use of `decimal.Decimal` is internal, so we can factor that out in a later PR.

amolenaar:
@pbrown12303 Thanks again for your help on making Gaphor the best modeling tool out there :)

pbrown12303:
@amolenaar You are quite welcome. I hope to add more. Starting my research now...

PR #3767, opened 2025-02-21
Ellipsize long class names in association property editor

### LINKED ISSUE #3745: association property page overtakes screen when member end name is too large
### Describe the bug
See title. Member end element names are listed on the property page. If the names are large it causes the property page to expand horizontally and reduce available model editing space

### To Reproduce
1. make 2 elements with long names
2. connect with an association
3. see lack of available area.

### Expected behavior
Member end names on the property page should not introduce and oversized property page
Possible options
- cut off length with max-width character limit (e.g. Member End(:this is an oversi...) )
   - mouse over could show the full name in an a wrapped box overlay?
- add a scroll bar
- wrap the line in the property page when it exceeds max allowed width. 

### OS
- [X] macOS
- [ ] Windows

### Version
Version of Gaphor: 3.0.0

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: Fixes #3745

### What is the new behavior?

Association member end names are ellipsized, so long names do not make the property editor unnecessary wide.

- [ ] Yes
- [x] No

### Other information

image

@tompkins-ct This is the fix, right?

### DIFF
--- a/gaphor/UML/classes/propertypages.ui
+++ b/gaphor/UML/classes/propertypages.ui
@@ -52,6 +52,7 @@
         <child>
           <object class="GtkLabel" id="head-title">
             <property name="label" translatable="yes">Head:</property>
+            <property name="ellipsize">end</property>
             <property name="xalign">0</property>
             <property name="hexpand">1</property>
             <style>
@@ -164,6 +165,7 @@ Use &lt;b&gt;-&lt;/b&gt;/&lt;b&gt;=&lt;/b&gt; to move items up or down.</propert
         <child>
           <object class="GtkLabel" id="tail-title">
             <property name="label" translatable="yes">Tail:</property>
+            <property name="ellipsize">end</property>
             <property name="xalign">0</property>
             <property name="hexpand">1</property>
             <style>


tompkins-ct:
Based on your figure, that's consistent with what I was thinking. The Ellipses look nice on the test label. 

Another potential issue is the editable text fields.
It might be good to develop a consistent approach to these editable text fields to prevent the property panel from expanding uncontrollably. My preference would be to grow vertically by the textline height and wrap because that's easier to visually read while editing. Elipses there would make it difficult for larger item names. 

It might be a good idea to add a test (probably a test model because it's UI related) with long item names on all text fields so these can be checked in the future with UI changes.

amolenaar:
@tompkins-ct Thanks for checking.

Text fields are not a problem: the text "scrolls" out of the text field if it's to long. The text entry widget has no feature to wrap, and I think it's not a good idea to use text area's everywhere.

danyeaw (approved):

danyeaw:
Thanks @amolenaar!

PR #3770, opened 2025-02-22
Update Windows development instructions

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

### What is the new behavior?

I checked the windows instructions. Made a few minor changes (e.g. python versions).

One thing I noticed is that I only need to install pycairo from gvsbuild. PyGObject works fine, at least on my setup.

It also works if I disallow installing the precompiled binaries for pycairo:

```
poetry config --local installer.no-binary pycairo
poetry install
```

Maybe those are alternatives to force-installing the builds made with GVSBuild. We can also add this bit of configuration in a `poetry.toml` file. Is that file supposed to be checked in?

BTW. I do not see any instructions on installing compilers. Is that done automatically nowadays?

### DIFF
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -24,7 +24,7 @@ To set up a development environment in Windows first install
 [Git](https://gitforwindows.org) by executing as an administrator:
 
 ```PowerShell
-winget install git
+winget install git.git
 ```
 
 ### GTK and Python with Gvsbuild
@@ -47,7 +47,7 @@ The default installation options should be fine for use with Gaphor.
 2. Open a PowerShell terminal as a normal user and check the python version:
 
    ```PowerShell
-   py -3.12 --version
+   py -3.13 --version
    ```
 
 #### Install Graphviz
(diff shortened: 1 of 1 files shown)


danyeaw:
Hi @amolenaar, I think you are right, we probably don't need to use the Gvsbuild generated wheels at all, because Gaphor has this shim in place:
https://github.com/gaphor/gaphor/blob/main/gaphor/windowsshim.py

I would like to get rid of this shim by building Windows Wheels since asking every user of PyGObject to create a load DLL shim isn't a great experience, but that shouldn't prevent us from simplifying the dev environment instructions. We'll probably need to test what the minimum things needed to pip install the sdist for PyGObject and pycairo.

The no-binary is needed for pycairo because they are publishing statically linked wheels of pycairo with cairo only and no GTK.

danyeaw:
Hi @amolenaar, I wonder if there is a hybrid approach which may be what you are suggesting:

1. Download GTK compiled with Gvsbuild zip and extract it
2. Set up some environment variables
3. Install minimum dependencies like MSVC so that you can build PyGObject and pycairo from source

amolenaar:
That should work, I suppose.

That's simpler than having to install 2 local wheels every time you did a `poetry install`.

We used to have instructions for setting up MSVC in our docs. Should we put it back?

amolenaar:
> @amolenaar, yes, I think so. Then I'm going to make it my mission to finish this build wheels for PyGObject project :)

It won't be easy, due to the dependency on GLib, GObject and GIRepository. But there must be a way. But then, having a statically linked pycairo doesn't help either.

danyeaw (approved):
Thanks @amolenaar!

Issue #3664, opened 2024-12-15
Copying selected model elements to the clipboard in a text, vector and bitmap format


ghillebrand:
### Is your feature request related to a problem?
When creating documentation based on models, it is nice to be able to quickly select the section of a model needed, and paste it in. 

### Describe the solution you'd like
When using `Edit > copy`, in addition to the gaphor format blob, plain text, vector and bitmap elements are written to the clipboard, to be accessible to other programs - wordprocessors, presentations, etc.
The text should be some human readable version of the elements and their attributes, with edges at the end. Ideally, the format is also moderately easily machine parsable. Possibly `PlantUML` may be a good target, even if it is not fully implemented for this. This may be related to [this issue about textual descriptions](https://github.com/gaphor/gaphor/issues/615)

### Describe alternatives you've considered
Currently, one has to export the section to a file, and then import the file to the working document. This means that model changes are less likely to be kept in sync, as this is an awkward workflow.
Exporting textual attributes seems to require writing some form of CLI export script. This is a significant barrier to more widespread usage of the models
One can use also screenshots, but these get big, and may not scale well, and don't include any attribute text.

### Additional context
It seems that between cairo surfaces tkInter, and/ or one of the Python clipboard libraries, this should be possible in an OS agnostic way.

amolenaar:
This feature is definitely on my wish list 👍

danyeaw:
Hi @ghillebrand, the PNG and SVG support is now merged. Do you think we should keep this open or include the text portion with #615?

ghillebrand:
Hi,
PNG & SVG are a very useful step forward - thank you.

I think #615 is the right goal, but it seems like it's going to take a while? 
How hard would it be to create a simple `<Object>\t<Attribute>\t<Value>` list of the textual elements by iterating through the selection set? For edges, possibly `<startItem>\t<endItem>\t<Attribute>\t<Value>`?
This makes the text easily accessible, even if it is a bit messy. It can be pulled into a spreadsheet/ table in a doc, and people can pull out the bits they want/ need? I don't know how other people's use-cases work, but for me, the notes in the model are quite important, particularly as I use the models in analysis, where there are still a goodly number of open issues to think about.

amolenaar:
I'm not sure how far `object-attribute-value` will get you: a lot of information is put in related objects. Like how attributes and operations are constructed.

ghillebrand:
I assume then that traversing all of those relationships is non-trivial? 
[...]

PR #3771, opened 2025-02-22
Copy PNG and SVG to the clipboard

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Diagram data can't be copied to other apps.

Issue Number: #3664

### What is the new behavior?

Apart from the (internal) copy buffer, also provide PNG and SVG images, so diagrams can be copied in other apps.

### Other information

* Paste function is now async/await instead of callbacks.

### DIFF
--- a/gaphor/ui/clipboard.py
+++ b/gaphor/ui/clipboard.py
@@ -2,13 +2,18 @@
 
 from __future__ import annotations
 
+import contextlib
+import io
 from collections.abc import Collection
 
+import cairo
 from gi.repository import Gdk, GLib, GObject
 
+from gaphor.asyncio import TaskOwner
 from gaphor.core import Transaction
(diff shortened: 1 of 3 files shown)


gaphor/ui/clipboard.py
```diff
@@ -19,69 +24,101 @@ def __init__(self, buffer):
...
-        self._copy(items)
+        self._copy(view.model, items)
 
         with Transaction(self.event_manager):
             for i in list(items):
                 i.unlink()
 
     def paste_link(self, view):
-        self._paste(view, paste_link)
+        self.create_background_task(self._paste(view, paste_link))
```
danyeaw, line 53: Nice idea to use asyncio for this!

danyeaw (approved):
Thanks @amolenaar, this looks really good!

PR #3772, opened 2025-02-23
Async actions and event handlers

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

We need to schedule special "background tasks" in order to perform async code.

### What is the new behavior?

For EventManager events, allow event handlers to be async functions. This makes it more ideomatic to work with async code.

Actions can also be async. However, an action can only be executed one at a time.

- [ ] Yes
- [x] No

### Other information

Cleaned out the `gaphor.asyncio` module. Now it only has 2 functions that can be removed once PyGObject 3.52 has been released.

### DIFF
--- a/gaphor/UML/general/imagepropertypage.py
+++ b/gaphor/UML/general/imagepropertypage.py
@@ -1,10 +1,10 @@
+import asyncio
 import base64
 import io
 from pathlib import Path
 
 from PIL import Image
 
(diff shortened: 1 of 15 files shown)


danyeaw (approved):
Thanks @amolenaar, this is a really nice improvement for working with the async code 🤩

PR #3794, opened 2025-03-16
Highlight drop zone during DnD operation

### LINKED ISSUE #3750: SysML frames change element containment
### Describe the bug
Not sure if this is intended behavior. Adding an element from a different package to the frame inside of a new diagram initially doesn't affect the element. If, however, I then move the element inside the frame, the element is moved to the diagram's containing package and I lose any "From X" info. The diagram still shows "From X", but the qualified name says otherwise.

### Expected behavior
Frames shouldn't affect the containment of items in the model, unless this is part of the standard? (I'm a complete newbie to SysML)

#### 1. Initial placement:

Image

Image

#### 2. After moving the block:

Image

Image

### OS
- [x] macOS
- [ ] Windows

### Version
Version of Gaphor:
```
Gaphor version:         3.0.0
Operating System:       Darwin (23.6.0, arm64)
Display:                GdkMacosDisplay
Python version:         3.13.1
GTK version:            4.16.12
Adwaita version:        1.6.3
GtkSourceView version:  5.14.2
Cairo version:          1.18.2
Pango version:          1.56.1
PyGObject version:      3.50.0
Pycairo version:        1.27.0
```

### LINKED ISSUE #3769: Not grouping on diagram
### Describe the bug
Dropping an element on a diagram block is not consistently grouping. Sometimes group functions as normal and other times the items do not group. No errors in log.

### To Reproduce

### Expected behavior
An element dropped inside the bounds of another should be grouped.

### OS
- [X] macOS
- [ ] Windows

### Version
Version of Gaphor: 3.0.0

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Grouping didn't work when dragging elements from Model Browser or Toolbox.

Issue Number: fixes #3769, fixes #3750

### What is the new behavior?

Grouping works, with drop zone support.

To test, drop an element from the model browser on another element in a diagram (e.g. drop one class on either of 2 packages).

Diagram frames will no longer change ownership of elements dragged onto them.

- [ ] Yes
- [x] No

### Other information

Maybe iterate items at a point, instead of only looking at the topmost one?

### DIFF
--- a/gaphor/UML/drop.py
+++ b/gaphor/UML/drop.py
@@ -2,8 +2,14 @@ from __future__ import annotations
 
 from gaphor import UML
 from gaphor.core.modeling import Base, Diagram
-from gaphor.diagram.drop import diagram_has_presentation, drop, drop_relationship
-from gaphor.diagram.presentation import Presentation, connect
+from gaphor.diagram.drop import (
+    diagram_has_presentation,
+    drop,
+    drop_relationship,
+    grow_parent,
+)
+from gaphor.diagram.group import change_owner
+from gaphor.diagram.presentation import ElementPresentation, Presentation, connect
 from gaphor.diagram.support import get_diagram_item, get_diagram_item_metadata
 
 
@@ -104,3 +110,25 @@ def drop_relationship_on_diagram(
                     if new_item:
                         return new_item
     return None
+
+
+@drop.register(UML.Element, ElementPresentation)
+def drop_element_on_element_presentation(
+    element: UML.Element, parent_item: ElementPresentation, x: float, y: float
+) -> Presentation | None:
+    item_class = get_diagram_item(type(element))
+    if not item_class:
+        return None
+
+    diagram = parent_item.diagram
+    item = diagram.create(item_class, subject=element)
+    assert item
+
+    if change_owner(parent_item.subject, element):
+        item.parent = parent_item
+        item.matrix.translate(x, y)
+        grow_parent(parent_item, item)
+
+    parent_item.request_update()
+
+    return item
--- a/gaphor/SysML/drop.py
+++ b/gaphor/SysML/drop.py
@@ -1,9 +1,35 @@
-from gaphor.core.modeling import Diagram
+from gaphor.core.modeling import Diagram, Presentation
 from gaphor.diagram.drop import drop
 from gaphor.diagram.presentation import connect
 from gaphor.diagram.support import get_diagram_item
+from gaphor.SysML.diagramframe import DiagramFrameItem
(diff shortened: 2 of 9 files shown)


tompkins-ct:
@amolenaar  Do you have a test to demonstrate programmatically dropping an element into an element presentation on a diagram? Looks like this is fixing some grouping issues and wondered if you had an example or test demonstrating that functionality. (mostly because I'm looking to accomplish that via some auto-generated diagrams ;) )

amolenaar:
I mostly tested this by starting the app and doing the actual drawing. It's probably better to add some tests as well.

amolenaar:
@tompkins-ct I added some tests.

gaphor/UML/drop.py
```diff
@@ -104,3 +110,25 @@ def drop_relationship_on_diagram(
...
+) -> Presentation | None:
+    item_class = get_diagram_item(type(element))
+    if not item_class:
+        return None
+
+    diagram = parent_item.diagram
+    item = diagram.create(item_class, subject=element)
+    assert item
+
+    if change_owner(parent_item.subject, element):
```
tompkins-ct, line 127: I don't know if this `if` statement is correct.

When you drop an existing element onto a new presentation of its current parent (e.g., on a new diagram), then the parent doesn't change but the target presentation of the parent still needs to be grown and the item placed within it's bounds.

tompkins-ct (commented):
comment on dropping on presentation.

danyeaw:
I like how this works with Blocks, that looks really good. For the use case of dropping a Block on a diagram, I find it a bit confusing to highlight a diagram border, though. It makes it feel like the block is going to appear on the border of the diagram, or that it is owned by the diagram.

gaphor/UML/drop.py
amolenaar: Yup. I changed the implementation of `change_owner()` to return `True` if the `parent_item.subject` is already the owner. This way calling `change_owner()` multiple times will always return `True` (idempotency).

amolenaar:
@danyeaw I updated the CSS to not highlight diagram frames when dragging over it. Is this what you mean?

danyeaw:
Hi @amolenaar, I'm still seeing diagram frame highlighting.

amolenaar:
@danyeaw When I drag an item on the frame, the frame doesn't highlight. Am I missing something?

https://github.com/user-attachments/assets/e3a6546e-8489-4bc3-b5b2-ba902bc38b81

danyeaw (approved):

danyeaw:
Hi @amolenaar, thanks, my environment must have been stale - I tried again, and it is working great now. This is a really nice improvement!

PR #3797, opened 2025-03-19
Update PyGObject

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Issue Number: N/A

### What is the new behavior?

Small round of dependency update. Removed upper bounds where possible. See https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/418

* Updating pygobject (3.50.0 -> 3.52.3)
* Use latest (git) version for PyInstaller

- [ ] Yes
- [x] No


danyeaw (commented):
Hi @amolenaar, these changes look great. I did notice we are still overriding PyGObject back to 3.50.x on Windows. Do you think we should update that as well or make it part of a separate PR, like the dependabot Gvsbuild updates?

amolenaar:
What would it entail to use 3.52 on Windows? Just not installing the prebuilt wheel?

amolenaar:
It's not fully working. There's an issue with finding typelibs.

danyeaw:
Hi @amolenaar, I isolated this issue to needing to set the `GI_TYPELIB_PATH` environment variable. This seems like a regression to me in gobject-introspection. It should automatically search for the typelibs in the lib\girepository-1.0 directory, since that is the normal search path.

amolenaar:
That's [this line in gobject-introspection](https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/girepository/girepository.c?ref_type=heads#L125)? This is [also in girepository](https://gitlab.gnome.org/GNOME/glib/-/blob/main/girepository/girepository.c?ref_type=heads#L250-251).

```python
>>> from gi.repository import GIRepository
>>> GIRepository.Repository().get_search_path()
['/opt/homebrew/Cellar/glib/2.84.0_1/lib/girepository-1.0', '/opt/homebrew/lib/girepository-1.0']
```

amolenaar:
@danyeaw I [found the issue](https://gitlab.gnome.org/GNOME/glib/-/issues/3657). Quick fix is to add a `-DDLL_EXPORT` CFlag when compiling glib. The real fix is, I think, to remove the check for the `DLL_EXPORT` variable.

danyeaw:
@amolenaar Nice find!

PR #3810, opened 2025-03-29
Present existing window on activation

City-busz:
This raises the window when the user launches the application and the main window is already open.

### PR Checklist

### PR Type

- [ ] Bug fix
- [x] Feature

Nothing happens when the main window is already open and the application is launched second time, the window is left in the background.

Issue Number: N/A

### What is the new behavior?
Existing main window will be raised on second launch.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/ui/__init__.py
+++ b/gaphor/ui/__init__.py
@@ -72,6 +72,8 @@ def run(argv: list[str], *, launch_service="greeter", recover=False) -> int:
         assert application
         if not application.sessions:
             application.get_service(launch_service).open()
+        elif gtk_app.get_active_window():
+            gtk_app.get_active_window().present()
 
     def app_open(gtk_app, files, n_files, hint):
         # appfilemanager should take care of this:


gaphor/ui/__init__.py
```diff
@@ -72,6 +72,12 @@ def app_activate(gtk_app):
         assert application
         if not application.sessions:
             application.get_service(launch_service).open()
+        else:
+            for session in application.sessions:
+                main_window = session.get_service("main_window")
+                if main_window.window:
+                    main_window.window.present()
+                    break
```
amolenaar, line 80: This is a really nice idea. However, instead of showing a random session, we can also show the last window that has been active.

```suggestion
        elif gtk_app.get_active_window():
            gtk_app.get_active_window().present()
```

City-busz: Yes, I was thinking about that. I updated the code.

amolenaar:
Thanks @City-busz

PR #3838, opened 2025-04-10
Replace floating toolbar

### LINKED ISSUE #3831: Floating toolbar layout
I just noticed that in the latest version of Gaphor there's a floating toolbar on the canvas. I'm not sure this is a good fit in this case layout wise, especially at the top of the view near the headerbar it makes the layout feel unbalanced.

I could maybe see a case for having floating zoom buttons in the bottom right like in Loupe, but undo/redo and the alignment options would definitely be a better fit in the headerbar. My intuition would be to just move all of these controls into the headerbar on the left edge of the right panel.

Image

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Replace the toolbar in the center of the diagram editor with something more sophisticated.

Issue Number: Fixes #3831 

### What is the new behavior?

A few functions end up in the header bar. The zoom buttons move to the lower-right corner.

Image

### Other information

By having the zoom buttons in the lower right corner, we can extend it and have a mini-map.


danyeaw (approved):
Thanks @amolenaar and @bertob, this looks nice! :star_struck:

PR #3842, opened 2025-04-11
Fix creating main window

City-busz:
The `window` property is used by conditions in multiple places, but it was always available due to the getter function, which is not the expected behavior. Turn it into an empty property, and assign the window only in the `open()` function, which is the only place where the window is supposed to be created.

This fixes the problem that `set_application()` has no effect, because it's called after the window is presented by `_on_file_manager_state_changed()`.

### PR Checklist

### PR Type

- [x] Bug fix
- [ ] Feature

The `app_id` property of the main window is "gaphor", because the window is not associated with the application before shown.

Issue Number: N/A

### What is the new behavior?
The `app_id` property of the main window is "org.gaphor.Graphor", the window properly initialized and associated with the application before shown.

- [ ] Yes
- [x] No


amolenaar:
Thanks @City-busz

PR #3846, opened 2025-04-13
Use view switcher in element editor

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Improve the element editor.

Issue Number: #3831

### What is the new behavior?

View switchers are in the element editor.

- [ ] Yes
- [x] No

### Other information

The style editor can also be added to the element editor/sidebar now.

Also tested with merge conflict.

### DIFF
--- a/gaphor/diagram/propertypages.py
+++ b/gaphor/diagram/propertypages.py
@@ -298,7 +298,7 @@ class InternalsPropertyPage(PropertyPageBase):
     This info may come in handy if you want to code.
     """
 
-    order = 400
+    order = 500
 
     def __init__(self, subject):
         self.subject = subject
--- a/gaphor/diagram/styleeditor.py
+++ b/gaphor/diagram/styleeditor.py
@@ -19,7 +19,7 @@ from gaphor.transaction import Transaction
 class StylePropertyPage(PropertyPageBase):
     """A button to open a easy-to-use CSS editor."""
 
-    order = 300
+    order = 450
     style_editor = None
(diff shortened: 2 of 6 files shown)


bertob:
The flat view switcher at the top is not the best fit here style-wise. I'd either move it to the bottom (like in the sidebar in e.g. Papers) or go with an [inline view switcher](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.InlineViewSwitcher.html) at the top.

amolenaar:
@bertob I played around a bit. I think the bottom bar looks best:

image

This is the inline view switcher:

image

bertob:
The bottom bar looks good, but I'd go for the wide policy, since there's enough space to fit icon next to label here and it'd avoid some of the whitespace on the sides.

amolenaar:
The view-switcher-bar does not have that option. I swapped it out for a AdwViewSwitcher:

image

danyeaw (approved):
Wow this looks amazing 🤩!

amolenaar:
It does, doesn't it?

danyeaw:
It doesn't need to be part of this PR, but what if we also moved the Open Style Editor button to the Styling view?

amolenaar:
I'm definitely in favor of integrating that dialog in the main ui. For a next PR, probably ;)

amolenaar:
I expect it to be flaky. The macOS 13 build also fails regularly.

PR #3888, opened 2025-05-17
Relation between CallBehaviorAction and Behavior should not be composite

### LINKED ISSUE #3862: Delete of Call Behavior Action Deletes Linked Activity
### Describe the bug
When deleting a call behavior action, the corresponding activity is also deleted from the model.

### To Reproduce
1. Create an Activity
2. Create an Activity Diagram
3. Add a Call Behavior Action
4. Set the Behavior of the Call Behavior Action to the Activity Created in Step 1

Note that I also have the setting for auto deleting from the model `off`, but in this case I wouldn't expect deletion in any case.

Also note, that if you have 2 call behavior actions pointing to the Activity, when deleting one call behavior action, the other call behavior action and the activity are deleted.

### Expected behavior
Only the Call Behavior Action is deleted from the model

### OS
- [x] macOS
- [ ] Windows

### Version
Version of Gaphor: 3.0.0

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Linked Behavior (e.g. Activity) is removed when CallBehaviorAction is removed.

Issue Number: Fixes #3862

### What is the new behavior?

Linked behavior is not removed.

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/UML/uml.py
+++ b/gaphor/UML/uml.py
@@ -1558,7 +1558,7 @@ DirectedRelationship.source.add(InformationFlow.informationSource)  # type: igno
 Element.owner.add(InformationFlow.realizingConnector)  # type: ignore[attr-defined]
 DirectedRelationship.target.add(InformationFlow.informationTarget)  # type: ignore[attr-defined]
 CallAction.result = association("result", OutputPin, composite=True)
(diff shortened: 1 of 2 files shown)


danyeaw (approved):

danyeaw:
Thanks @amolenaar!

PR #3923, opened 2025-06-04
Use media features during style sheet compilation

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

A `dark-mode` option has to be provided to every painter and node.

Issue Number: N/A

### What is the new behavior?

The `prefers-color-scheme` option is now provided to the style sheet compiler, so all bits that do not apply are not compiled.

- [x] Yes
- [ ] No

### Other information

Currently this PR has several fixes and improvements:

* I like that media queries are resolved at CSS compile time. Conceptually that makes sense, I think.
* I like that the preferred color scheme is now defined in the `Painter`, and not propagated via `StyleNode`s.
* The style sheet can directly be obtained from the `ElementFactory`. This makes more sense than the diagram
* Style sheets are now in charge of computing styles. Computed styles are cached. The cache is cleared when diagrams need to update themselves.
* Parent styles are no longer calculated based on their psuedo style (selected, hovered, drop zone).
* Adds support for empty media query (`@media { ... }`). This should [evaluate to true](https://drafts.csswg.org/mediaqueries-5/#mq-list).


gaphor/core/styling/declarations.py
```diff
@@ -56,6 +56,12 @@ class Var(NamedTuple):
     name: str
 
 
+class PrefersColorScheme(Enum):
```
danyeaw, line 59: This is a nice improvement to replace `dark_mode` :+1:

danyeaw (approved):
This looks great, I like that you centralized the management to `style_sheet.compute_style()` and the caching should result in some speed improvements.

gaphor/core/styling/declarations.py
amolenaar: It does make it easier to understand, doesn't it :).

PR #3924, opened 2025-06-04
Support CSS classes

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Items in a diagram can only be identified by tag.

Issue Number: N/A

### What is the new behavior?

With classes, we have more ways to define styling.

The initial idea is to have two styles: `.element`, and `.relation`. Those should make it easier to, for example, move the name of an element to the top if an element contains nested elements.

- [ ] Yes
- [ ] No


danyeaw (approved):

danyeaw:
Thanks @amolenaar!

Issue #3906, opened 2025-05-27
Package Diagram formatting with nested packages


tompkins-ct:
### Describe the bug
On package diagrams, nested packages don't result in package label moving to accommodate the nesting. Blocks reformat to visually depict nested items. Package diagrams should do this as well. 

Note: nesting blocks / classes works just fine.

### To Reproduce
1. Create a package diagram
2. Place packages on the diagram
3. Nest packages in the diagram (these attach as expected but visually the package block isn't adjusted.

If possible, add the error report from `≡` (top-right menu) → `Tools` → `Error Reports`.

### Screenshots

![Image](https://github.com/user-attachments/assets/feb34622-c72e-4a4f-b0a2-5d42f27024d2)

### OS
- [ ] Linux (Please put in notes the specific distro)
- [X] macOS
- [ ] Windows

### Version
Version of Gaphor: 3.1.0

amolenaar:
It works fine for classes, but not for a package:

![Image](https://github.com/user-attachments/assets/26b80837-776b-40fd-8690-1a5d915e27cc)

The CSS bit at work for this is:

```css
:not([children=""]):has(compartment),
:not([children=""]) > compartment {
  justify-content: start;
}
```

PR #3925, opened 2025-06-05
Align text vertically if package is contained

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: issue #3906 

### What is the new behavior?

Parent package name is vertically aligned at the top:

image

- [ ] Yes
- [ ] No

### Other information

This is a quick fix in the CSS. I'm considering to give all elements an "element" class, so we can easily find out if an element is contained: the diagram structure for CSS also includes name and stereotype elements, which should be excluded if we want to see if an element contains another element.

### DIFF
--- a/gaphor/diagram.css
+++ b/gaphor/diagram.css
@@ -249,6 +249,7 @@ state compartment:first-child {
 
 :has(compartment + compartment),
 :has(regions),
+:has(package),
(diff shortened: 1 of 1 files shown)


amolenaar:
Try #3924 instead.

Issue #3915, opened 2025-06-02
Unable to box select inside diagram frames


danyeaw:
I accidentally moved the previous issue https://github.com/gaphor/gaphor/issues/3905 to a discussion. Reopening a new issue.

<div type='discussions-op-text'>

<sup>Originally posted by **tompkins-ct** May 27, 2025</sup>
### Describe the bug
When a diagram frame is present, you can't box select. The diagram frame is selected and dragged instead of creating a selection box.

### To Reproduce
1. Create a SysML diagram
2. Add blocks to the diagram
3. Try to box select with your cursor inside the diagram frame

If possible, add the error report from `≡` (top-right menu) → `Tools` → `Error Reports`.

### Expected behavior
Diagram frame should not be selected unless the frame edge is specifically targeted.

### OS
- [ ] Linux (Please put in notes the specific distro)
- [X] macOS
- [ ] Windows

### Version
Version of Gaphor: 3.1.0

### Additional information
</div>

amolenaar:
This is an interesting one. We use `Item.point` to determine the distance from a point to an item (negative distance means your inside the item). I suppose:

* We only want to select the frame by clicking on the border
* Items should still be placed in the frame (be children of the frame)

For both `Item.point()` is used.

amolenaar:
@danyeaw Is it valid to assume that, if you have a diagram frame on a diagram, all items are always placed inside the diagram frame? For example: if you would place a block outside a BDD, the BDD frame will grow so it also contains that block?

danyeaw:
Hi @amolenaar - yes, that's a good assumption

amolenaar:
I was wondering: DiagramFrame is currently a diagram item, not a decorative thing, such as the diagram type for UML diagrams.

Is there a reason for DiagramFrame to be a diagram item?

- Should you be able to attach items to the border?
- Should you be able to place items outside the diagram frame?

If the answer to both is no, then we do not need frames as a diagram item.

danyeaw:
Thanks @tomkins-ct for the original issue report!

PR #3966, opened 2025-07-11
Allow Rubberband selection in diagram frame

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: #3915

### What is the new behavior?

Rubberband selection is possible in diagram frame. Frames are not selected by default.

- [ ] Yes
- [x] No

### Other information

To implement it nicely, I had to upgrade Gaphas and Generic.


danyeaw:
Hi @amolenaar, for example with an activity diagram, how do I resize the diagram frame with this change?

image

amolenaar:
You can still use the handles in the corner.

Maybe that's not obvious enough and the handles should still be visible on hover?

danyeaw:
Hi @amolenaar, I didn't see it at first, but I think that is the right amount of discoverability without getting in the way all the time.

amolenaar:
I can change it to show the handles on hover, if you think that works better.

danyeaw (approved):
Thanks @amolenaar!

Issue #1399, opened 2022-03-12
Full Support for SysML v2


danyeaw:
### Is your feature request related to a problem?
The SysML v2 Request for Proposal was released in December 2017. The final submission is expected in 2022Q2, with the final release probably in 2023.

This feature request states our intent to upgrade to SysML v2, with work starting once the final submission is made. Ideally, we can beat the bigger commercial tools and be the first out of the gate.

API Services are also being tracked in #1188
Text entry of diagrams are also being tracked in #615

sz332:
Hello guys, we just had some discussion on the systems engineering discord server, and actually this would be a great feature, because SysML V2 really lacks an editor. What I found that everything they made is based on EMF, and actually there are some libraries capable of handling EMF. What do you think, should we make some steps towards SysML V2 support?

danyeaw:
@sz332 Ready to help us figure out how to do that?

sz332:
https://github.com/Systems-Modeling/SysML-v2-Release this is the release page for SysML-v2. 
https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation is the location for a reference implementation.

SysML V2 is based on KerML instead of just being an UML profile. KerML is implemented on top of EMF metamodel and there is a Jupiter support so theoretically we could import KerML / SysML v2 from Python, that part is expected to be implemented. 

So the model can be read and probably extended, but the part on how to actually display a diagram is totally missing. That would be a huge, HUGE advantage of Gaphor.

amolenaar:
All this SysML code is Java. They support a Jupyter kernel (in Java), but that only allows for interacting with it through the notebooks.

amolenaar:
I think the relevant documentation is [in the sysml github repo](https://github.com/Systems-Modeling/SysML-v2-Release/tree/master/doc), in particular [2-OMG_Systems_Modeling_Language.pdf](https://github.com/Systems-Modeling/SysML-v2-Release/blob/master/doc/2-OMG_Systems_Modeling_Language.pdf).

It looks like KerML has a lot of overlap with the UML core modules.

sz332:
I found that there are libraries that can work with ecore files. Might it help?

amolenaar:
We'll have to figure that out. Our data model has some features not present in (for example [PyEcore](https://pypi.org/project/pyecore/)), such as event emission. We may be able to use them to generate (part of) the data model, though, since they contain the same constructs.

danyeaw:
[...]

PR #3975, opened 2025-07-25
KerML and SysML 2.0 model generated from XMI

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

We have no support for SysMLv2.

Instead of rebuilding the model in Gaphor, let's create a small converter that converts "just enough" to have a proper Gaphor model that we can use for code generation.

Issue Number: #1399

### What is the new behavior?

Gaphor has models for KerML and SysMLv2. You can enable it by starting gaphor with:

```sh
GAPHOR_FEATURE_FLAG=sysml2 gaphor
```

This PR supports a minimal model. Two diagram items, and a tree model. 

The KerML and SysML2 model are generated directly from the (official) XMI file.

The tree model is swapped if you switch between UML and SysML2. I'm not sure if ownership works properly. That should be checked once we implement packages. The tree model is pretty much a copy of the UML one. I think it makes sense to keep it separate though, since relations are defined differently in SysML2.

The KerML and SysML2 models are quite minimal. Quite a lot of warnings are printed. There's no proper subsetting. I like to parse the OCL expressions for that, but that's an (epic) journey on itself.

No documentation is generated about the model. Not sure if we should.

- [ ] Yes
- [x] No


gaphor/KerML/group.py
```diff
@@ -0,0 +1,32 @@
+from gaphor.diagram.group import Root, group, owner, owns, ungroup
+from gaphor.KerML import kerml
+
+
+@owner.register
+def _(element: kerml.Element):
+    return element.owner or Root
```
amolenaar, line 7: Should be owner or `None`. `kerml.Namespace` should be at `Root`.

gaphor/KerML/tests/test_kermlfmt.py
```diff
@@ -0,0 +1,31 @@
...
+    definition.declaredName = "PartDef"
+
+    assert format(definition) == "PartDef"
+
+
+def test_format_sysml2_usage_uses_kerml_element_formatter(element_factory):
+    usage = element_factory.create(sysml2.PartUsage)
+    usage.declaredName = "PartUse"
+
+    assert format(usage) == "PartUse"
```
amolenaar: Remove redundant tests

gaphor/SysML2/tests/test_dependency_boundaries.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
```
amolenaar: This file can go away. Make a pytest-archon rule instead

gaphor/SysML2/toolbox.py
```diff
@@ -0,0 +1,124 @@
...
+
+def part_definition_config(item: PartDefinitionItem) -> None:
+    item.subject.declaredName = gettext("Part Definition")
+
+
+def requirement_definition_config(item: RequirementDefinitionItem) -> None:
+    item.subject.declaredName = gettext("Requirement Definition")
+
+
+general_tools = ToolSection(
```
amolenaar, line 33: Because of images (UML.Image) we have to copy the general tools for now.

gaphor/SysML2/treemodel.py
```diff
@@ -0,0 +1,283 @@
...
+        )
+
+    def tree_item_for_element(self, element: Base | RootType) -> TreeItem | None:
+        if element is Root:
+            return None
+        if (owner_branch := self.owner_branch_for_element(element)) is not None:
+            return next((ti for ti in owner_branch if ti.element is element), None)
+        return None
+
+    def include_element(self, element: Base) -> bool:
```
amolenaar: Move to a function

gaphor/UML/treemodel.py
```diff
@@ -259,8 +259,18 @@ def tree_item_for_element(self, element: Base | RootType) -> TreeItem | None:
             return next((ti for ti in owner_branch if ti.element is element), None)
         return None
 
+    def include_element(self, element: Base) -> bool:
```
amolenaar: Move to a function

gaphor/ui/modelbrowser.py
```diff
@@ -226,7 +226,10 @@ def tree_view_create_element(self, id: str):
 
         with Transaction(self.event_manager):
             element = self.element_factory.create(element_def.element_type)
-            element.name = element_def.name
+            if hasattr(element, "name"):
+                element.name = element_def.name
+            elif hasattr(element, "declaredName"):
+                element.declaredName = element_def.name
             if own:
```
amolenaar, line 253: Move element creation to the tree model?

[update] I don't think TreeModel is the place: tree model is a read-only model, and it should not be convoluted with element creation.

The next subject is ModelingLanguage. Although it's sort of a configuration/meta-service (returns classes and definitions)

gaphor/UML/group.py
```diff
@@ -137,6 +137,30 @@ def class_ungroup(parent, element) -> bool:
...
+    parent.ownedLiteral = element
+    return True
+
+
+@group.register(UML.Operation, UML.Parameter)
+def parameter_group(parent: UML.Operation, element: UML.Parameter):
+    if element.owner:
+        return False
+    parent.ownedParameter = element
+    return True
```
amolenaar, line 161: Those are needed in order to properly load the XMI model.

Does this have impact on UML models? For example: can parameters be moved from one operation to another?

test-models/simple-items.gaphor
```diff
@@ -30,12 +30,6 @@
 <diagram>
 <ref refid="DCE:658E851A-3A1D-11DC-8B61-000D93868322"/>
 </diagram>
-<horizontal>
```
amolenaar, line 33: Model has changed since properties are now saved under a different name. 

Test if this has effect on models: set/unset line properties

REUSE.toml
```diff
@@ -32,3 +32,10 @@ path = "gaphor/ui/language-specs/gaphorcss.lang"
 precedence = "aggregate"
 SPDX-FileCopyrightText = ["2004 Scott Martin <scott@coffeeblack.org>", "2010 Patryk Zawadzki <patrys@pld-linux.org>", "2016 Tobias Schönberg <tobias47n9e@gmail.com>", "2018 Jeffery To <jeffery.to@gmail.com>", "2022 Arjan Molenaar <gaphor@gmail.com>"]
 SPDX-License-Identifier = "LGPL-2.1-or-later"
+
+[[annotations]]
+path = ["models/KerML-25-04-04.xmi", "models/SysML2-25-02-15.xmi"]
+precedence = "aggregate"
+# Source: https://github.com/Systems-Modeling/SysML-v2-Release/
+SPDX-FileCopyrightText = ["2019 Object Management Group, Inc. (OMG)"]
+SPDX-License-Identifier = "LGPL-3.0-only"
```
amolenaar, line 41: I think (assume) this is the right license. Should double check, though.

gaphor/codegen/coder.py
```diff
@@ -282,11 +306,15 @@ def associations(
     c: UML.Class,
     overrides: Overrides | None = None,
 ):
-    redefinitions = []
+    redefinitions: list[tuple[str, str]] = []
+    subsets: list[tuple[str, str, set[str]]] = []
+    emitted: set[str] = set()
```
amolenaar: Use a more understandable name for "emitted".

Is it enough to keep track of the name, or should it be full name, or maybe more (object id?)?

gaphor/diagram/group.py
```diff
@@ -76,9 +79,25 @@ class RootType(Enum):
...
+
+    This distinction has been made to allow the GUI (model browser) to
+    distinguish between elements that can appear at the model root (e.g. Packages),
+    versus elements that should not be shown there (comments, properties).
+    """
     return None
 
 
+@owner.register
+def _(element: Diagram):
```
amolenaar, line 97: Does this break existing behavior? Are (UML/SysML) diagrams placed correctly?

amolenaar (commented):
It's getting there. There's some things to test, but the basics for SysML v2 model support are in place.

There's still important stuff missing: relationships, and many components.

gaphor/diagram/group.py
amolenaar: Works fine.

gaphor/UML/group.py
amolenaar: Parameters cannot be moved.

EnumerationLiteral do not show in the tree view

gaphor/KerML/group.py
amolenaar: Let's do this once grouping becomes a thing. We have no packages at this moment.

PR #3981, opened 2025-07-30
Only import classes from their authoritative model

amolenaar:
Builds on #3979 

### PR Type

- [ ] Bug fix
- [x] Feature

Models are a bit loose (e.g. you can import `Class` from the SysML model, although it's defined in UML).

This breaks as soon as we add another modeling language: KerML or SysMLv2.

Issue Number: N/A

### What is the new behavior?

This PR creates some prerequisites needed to make working with KerML and SysML2 models easier. Those models have similarly named types as our current UML/SysML model and should therefore be strictly namespaced.

The import mechanism has been simplified: imported packages should exist in a package with a name similar to the model name (entry point name). This makes importing types from super-models more predictable.

- [x] Yes
- [ ] No

You can only import classes from their authoritative model.

### DIFF
--- a/gaphor/C4Model/c4model.py
+++ b/gaphor/C4Model/c4model.py
@@ -18,14 +18,14 @@ from gaphor.core.modeling.properties import (
 )
 
 
-from gaphor.UML.uml import Actor
-class Person(Actor):
+from gaphor.UML.uml import Actor as _Actor
+class Person(_Actor):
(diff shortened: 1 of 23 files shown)


danyeaw (approved):
Thanks @amolenaar - this is a great enhancement towards improving how we namespace our modeling languages.

PR #4022, opened 2025-09-21
Fix macOS builds

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

macOS builds are constantly failing.

Issue Number: N/A

### What is the new behavior?

Stable macOS builds, hopefully.

### Other information

* Updated macOS intel image to `macos-15-intel`. Not sure what that does on minimal OS version for our build
* Build on macOS with Homebrew's Python version
* Simplified self-test
* Install Poetry with pipx
* Fix locale file generation
* Allow support for integrated GPU (See https://gitlab.gnome.org/GNOME/gtk/-/issues/7676)

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -203,7 +203,6 @@ jobs:
           persist-credentials: false
       - name: Install macOS Dependencies
         env:
-          HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
           HOMEBREW_NO_INSTALL_CLEANUP: 1
         run: |
           brew update
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -183,7 +183,7 @@ jobs:
     strategy:
       matrix:
         target:
-          - runner: macos-13
+          - runner: macos-15-intel
             arch: intel
           - runner: macos-15
             arch: arm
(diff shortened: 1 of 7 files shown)


amolenaar:
Build results are better, but it looks like locale files are not packaged by PyInstaller (anymore?).

amolenaar:
@danyeaw The intel build works on macOS 11, so we're good.

amolenaar:
> Build results are better, but it looks like locale files are not packaged by PyInstaller (anymore?).

Fixed

.github/workflows/full-build.yml
```diff
@@ -218,22 +217,16 @@ jobs:
           install_name_tool -change @rpath/librsvg-2.2.dylib $(brew --prefix librsvg)/lib/librsvg-2.2.dylib $(brew --prefix librsvg)/lib/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader_svg.so
           codesign --sign - --force --preserve-metadata=entitlements,requirements,flags,runtime $(brew --prefix librsvg)/lib/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader_svg.so
           GDK_PIXBUF_MODULEDIR=$(brew --prefix)/lib/gdk-pixbuf-2.0/2.10.0/loaders gdk-pixbuf-query-loaders --update-cache
-      - name: Ensure Homebrew Library Path (ARM only)
-        if: matrix.target.arch == 'arm'
-        run: echo "DYLD_LIBRARY_PATH=/opt/homebrew/lib" >> $GITHUB_ENV
+      - name: Ensure Homebrew Library Path
+        run: echo "DYLD_LIBRARY_PATH=$(brew --prefix)/lib" >> $GITHUB_ENV
         shell: bash
-      - name: Set up Python
```
danyeaw, line 225: Is relying on the homebrew installed Python going to consistently work?

danyeaw (commented):
Hi @amolenaar - nice one, this looks like a great cleanup! I just had one question.

.github/workflows/full-build.yml
amolenaar: It's the same Python version I use locally. I do not think the python installation used was the problem, so we can revert this change, if you think that's better.

danyeaw: Ok great. Ya, I think being consistent with using setup-python to get the right Python version would be better.

PR #4072, opened 2025-10-21
Move selected items with arrow keys

### LINKED ISSUE #3990: Move objects with directional keys
Moving diagram presentations with the mouse can be unreliable, and I'd like a more precise method to shift presentations around.

Enable moving presentation elements / jog points using the directional keys. This should increment the location of the presentation / job in the direction of the arrow at intervals defined by a spacing grid (possibly a coarse and fine resolution determined by a modifier key (e.g. shift or ctr/cmd)

Continue with the current shifting scheme.

amolenaar:
### PR Type

- [ ] Bug fix
- [x] Feature

Issue Number: Fixes #3990

### What is the new behavior?

Move selected items using the arrow keys.

- [ ] Yes
- [ ] No

### DIFF
--- a/gaphor/diagram/tools/__init__.py
+++ b/gaphor/diagram/tools/__init__.py
@@ -8,6 +8,7 @@ from gaphas.tool.scroll import pan_tool
 import gaphor.diagram.tools.handlemove
 from gaphor.diagram.tools.dnd import drop_target_tool
 from gaphor.diagram.tools.dropzone import drop_zone_tool
+from gaphor.diagram.tools.itemmovetool import item_move_tool
 from gaphor.diagram.tools.itemtool import find_item_and_handle_at_point, item_tool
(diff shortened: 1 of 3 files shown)


amolenaar:
Already implemented as (<kbd>Shift</kbd>)+<kbd>Alt</kbd>+<kbd>Up</kbd>.

tompkins-ct:
@amolenaar @danyeaw  In that case, can we get it documented somewhere? The shortcuts dialog does not show this key-combination and It doesn't show up with a quick search of the documentation. The current one is okay, but not intuitive due to the use of <kbd>Alt</kbd>. 

Pressing the raw arrow keys seem to move the highlighted item which is unintuitive. (it makes sense in the model tree but nowhere else, I would expect <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd> to perform that function).

PR #4263, opened 2026-04-14
Set html-report default zoom level to 1

frankvanbever:
This prevents small diagrams from being blown up. This change only affects the default zoom level.

Fixes #4228

### PR Checklist

### PR Type

- [x] Bug fix
- [ ] Feature

image

 
 Diagrams in the html report are always zoomed to the full canvas

Issue Number: #4228 

### What is the new behavior?

image

The zoom level is always set to 1 by default

- [ ] Yes
- [x] No

### DIFF
--- a/gaphor/plugins/htmlreport/report.js
+++ b/gaphor/plugins/htmlreport/report.js
@@ -176,10 +176,6 @@
     var svg = doc.documentElement;
     if (!svg || svg.nodeName !== "svg") return;
 
-    // Let the SVG scale to fill its container via CSS; the viewBox
-    // attribute (preserved) handles the internal coordinate mapping
-    svg.removeAttribute("width");
-    svg.removeAttribute("height");
     // adoptNode transfers ownership from the parsed document to ours,
     // avoiding a cross-document insertion error
     content.appendChild(document.adoptNode(svg));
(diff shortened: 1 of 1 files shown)


gaphor/plugins/htmlreport/report.js
```diff
@@ -205,9 +201,23 @@
           zoomScaleSensitivity: 0.3
         });
 
+        // Prevent small diagrams from being scaled up beyond 100%
+        var realZoom = currentPanZoom.getSizes().realZoom;
+        var cappedZoom = realZoom > 1 ? currentPanZoom.getZoom() / realZoom : null;
```
danyeaw, line 206: Hey @frankvanbever, thanks for fixing this!

If I understand this correctly, `cappedZoom` is captured at init time and reused in the reset handler. If the user resizes the browser window, `realZoom` will have changed but `cappedZoom` won't update. The reset behavior might then be wrong after a resize. Have you checked the behavior after a window resize?

amolenaar:
I added a small commit that places the diagrams in the top-left corner of the diagram content space. This looks nicer IMHO, especially for diagrams with a diagram type and name:

image

gaphor/plugins/htmlreport/report.js
frankvanbever: I forgot to reply but yes it does not properly center when you resize the window. I have an implementation that does that but I think it might conflict with @amolenaar's top left corner implementation. I think his approach will also solve this problem.

danyeaw (approved):

danyeaw:
Thanks @frankvanbever and @amolenaar!

PR #4300, opened 2026-05-13
Rename umlproperty

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

The base property for models is called `umlproperty`. This was an okay-ish name when we had only one model in use. Now with multiple models, and SysML v2 on the way, a more generic name would be better.

Also turn the "subsettable" feature into a mixin, allowing it to be applied to the appropriate types.

`umlproperty` is still there, for backwards compatibility.


danyeaw (approved):
Thanks @amolenaar!

PR #4301, opened 2026-05-13
SysML2: derived properties from OCL expressions

amolenaar:
This PR builds on #3975 

### PR Type

- [ ] Bug fix
- [x] Feature

The SysML2 model has no proper way to automatically generate 

Issue Number: #1399

### What is the new behavior?

Derived associations are derived from the `derive-` constraints defined in the KerML/SysMLv2 models.

The code generator has been updated to create `watch()` statements for derived properties. This is the same mechanism we already use for diagram items.

### Other information

* Events are emitted if the derived property changed. Only the values (`old_value`, `new_value`) are not updated correctly.
* The OCL parser uses pyparsing. This library is already used by one of our dependencies.
* The grouping rules have been updated to support the constructs we read from the SysML2 XMI file.
* Because we can now use the OCL constraints directly, we do not need the overrides file for KerML anymore.
* Parts of the code are generated (Copilot).


danyeaw:
Hey @amolenaar, I started looking at this. Thanks so much for pushing SysML v2 forward, derived properties is a huge step. Could you provide a little more detail to guide about what is in scope of this PR and what is out of scope? I think we want to incrementally improve this, but should the lambdas be returning 100% the right types for the properties at this point?

gaphor/SysML2/sysml2.py
```diff
@@ -87,6 +89,12 @@ class TriggerKind(enum.StrEnum):
 
 
 class Definition(_Classifier):
+
+    def __init__(self, id=None, model=None):
+        super().__init__(id, model)
+        w = self.watcher()
```
amolenaar, line 95: Should keep a reference to the watcher, so we can unsubscribe all on unlink.

amolenaar:
> Could you provide a little more detail to guide about what is in scope of this PR and what is out of scope? 

I updated the description.

> I think we want to incrementally improve this, but should the lambdas be returning 100% the right types for the properties at this point?

I haven't thoroughly checked that yet. My focus was on the `owner` relationship, since that's what we'll use for the Model Browser. This needs more testing for sure.

Issue #3998, opened 2025-08-15
ESC key should switch back to pointer tool


fweber3:
### Is your feature request related to a problem?

No.

### Describe the solution you'd like

For switching the current diagram tool back to the pointer tool, there currently is a) "Reset tool automatically" in the general preferences or b) pressing the ESC key.

I would like a right mouse click inside the diagram to work the same as pressing the ESC key.

### Describe alternatives you've considered

The current state is also usable, IMVHO.

### Additional context

As far as I could tell, the right mouse button is still unused in the given context.

My requested behaviour is similar to how FreeCAD handles its many sketcher tools. The general context of "drawing with many tools" seems very comparable to me.

amolenaar:
Hi @fweber3, The ESC key should bring you back to the pointer tool. I confirm that that's not happening anymore. This is a regression bug.

fweber3:
Ahhm ... this could be a misunderstanding ... the point of my feature request was not to bugreport a broken ESC key (which I'm all for fixing, but it was working when I wrote the feature request).

Instead, my request here was to make single-right-mouse-button-click behave identical to ESC - *in addition* to a working ESC key. I find this much more usable than "switching" between mouse/keyboard hand for entering/exiting a tool

amolenaar:
A right-click already shows the popup menu. This is pretty standard behavior IMHO.

fweber3:
Obviously, there's a huge lot of variability, personal opinion and also learned reflexes in any "well-usable UI". My main application is FreeCAD (~ 2-3 hours/day), so their UI behaviour is second nature for me - whether it is fitting for Gaphor, I dare not judge.

Their _sketcher_ tool uses a tool palette to edit 2D drawings and add (exact, numeric) geometric constraints - IMVHO very similar to drawing a Gaphor diagram and entering initial object names.

In _sketcher_:
- in "pointer mode", right-click opens a (long) context menu, ESC-key closes the current drawing (or the open context menu)
- when any other tool is active, the general context menu is unavailable (because it's contents are not relevant _while actively using a specific tool_). Instead right-click cancels the tool and reverts back to "pointer mode", as does ESC-key

This keeps the sequence "select tool" [leftclick] - "draw with tool" [leftclicks/-drags] - "back to pointer" [rightclick] - "rearrange/resize objects" [leftdrags] - "select tool" ... entirely in the mouse hand. The other hand is used for number/text entry, or to rest my head on (literally). Reaching for ESC is a distraction.

[...]

PR #4305, opened 2026-05-17
Make shortcuts work inside the toolbox pane

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

When you select a tool in the toolbox with your mouse, you can not reset it (press `Esc`) or select another tool via the keyboard.

Issue Number: #3998 

### What is the new behavior?

Pressing a key when the toolbox has "focus" works the same as when a diagram has "focus".


danyeaw:
Thanks Arjan!

danyeaw (approved):

Issue #4245, opened 2026-04-06
Gaphor rendering issue -> full window redraw?


tompkins-ct:
### Describe the bug
While running gaphor with windows behind it, gaphor will occasionally break rendering (See figure).

### To Reproduce
Unsure of normal input. Typically occurs after saving a model though using `cmd` + `s`
- Running python via `poetry run gaphor`

Debugging doesn't show any errors.

```
Gaphor version:         3.3.0
Operating System:       Darwin (25.3.0, arm64)
Display:                GdkMacosDisplay
Python version:         3.13.11
GTK version:            4.22.2
Adwaita version:        1.9.0
GtkSourceView version:  5.20.0
Cairo version:          1.18.4
Pango version:          1.57.1
PyGObject version:      3.56.2
Pycairo version:        1.29.0


No errors have been intercepted. We’re good.
```

### Expected behavior
Rendering should be stable.

### Screenshots

<img width="611" height="464" alt="Image" src="https://github.com/user-attachments/assets/dbef3383-06f1-457c-8194-47c8e1cefbd8" />

### OS
- [ ] Linux (Please put in notes the specific distro)
- [X] macOS
- [ ] Windows

### Version
Version of Gaphor: current version development version https://github.com/gaphor/gaphor/commit/1507297aa88b77ddcf4f8c95986bb3a1f415dd4c

danyeaw:
Hey @tompkins-ct, thanks for submitting this. I can reproduce this with version 3.3.0 on macOS. It appears to be some type of upstream macOS video rendering issue with GTK 4.22.2.

mikekidner:
Yep, I get it on MacOSX also.

amolenaar:
I tried with the libadwaita and gtk demos, but I cannot reproduce the issue. With Gaphor the issue is easy to trigger.

I can also trigger it with high certainty after a save. Maybe it has something to do with the save action happening in the same thread? So a simple example with a sleep of a few 100ms will trigger it.

I get the issue with both the GL and Cairo renderer (haven't tested Vulkan).

amolenaar:
https://gitlab.gnome.org/GNOME/gtk/-/work_items/8214

amolenaar:
As a workaround, you can set `GSK_DEBUG=full-redraw` in your environment before starting Gaphor.

PR #4330, opened 2026-06-05
Fix window glitches on macOS

amolenaar:
### PR Type

- [x] Bug fix
- [ ] Feature

Issue Number: #4245

### What is the new behavior?

We did set the border radius on macOS to 0.
Normally, with border radius > 0, a "Radius Corner" node is added to GTK's frame graph. With border radius == 0, no such node is added, casing Transform and Opacity nodes to end up at the top level. Somehow this interferes with the rendering on macOS. The simple solution is to set a small radius value.

- [ ] Yes
- [ ] No

### DIFF
--- a/gaphor/ui/styling-macos.css
+++ b/gaphor/ui/styling-macos.css
@@ -15,7 +15,7 @@
 
 .csd {
   box-shadow: none;
-  border-radius: 0;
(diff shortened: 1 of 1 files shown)


mikekidner:
Thank you for this fix.

PR #4346, opened 2026-06-27
Update symbolic icons

amolenaar:
### PR Type

- [ ] Bug fix
- [ ] Feature

Our current icons need librsvg (now a fallback option in GTK) for rendering. This can be a limitation in the near future. We already had trouble with GTK 4.22.0.

### What is the new behavior?

The stencil file and build has been dropped. All icons are now simply created on a 16x16 grid. The ones updated by hand are simple enough that they can be edited with a text editor (I use VSCode with SVG extension).

Ideally all icons are redrawn (by hand), to make them simpler to work with and more consistent. Also small differences in icon styling can lead to inconsistency in the UI.

The SVG syntax used is quite straightforward. I used mostly `<path>` tags and some `<circle>`s. Paths have a `d` attribute that allows you to move (`m` or `M`) the cursor, draw a line (`l` for relative, `L` for absolute) and close a path (`z` or `Z`).
paths are assigned a class, normally `foreground-stroke`, so the color can be updated when we switch between light and dark mode.

- [ ] Yes
- [ ] No

### Other information

I wanted the icons fixed, because I worked on a little project (https://github.com/amolenaar/Adwaita-Framework). The idea is that this will provide the libraries we need to bundle Gaphor on macOS. Currently we use homebrew, which results in too many libraries (yes, X11 is bundled :/ ).

See https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-4-22/docs/symbolic-icons.md.


danyeaw (approved):
Thanks @amolenaar!

PR #4347, opened 2026-06-28
Render the placement icon without GdkPixbuf

amolenaar:
Should be combined with #4346.

### PR Type

- [ ] Bug fix
- [ ] Feature

We have some code that utilized `GdkPixbuf` to create a special "placement" cursor if a tool is selected from the toolbox.

Issue Number: N/A

### What is the new behavior?

Use GTK only to render the placement icon. This
basically makes our code no longer dependent on
GdkPixbuf, since all (svg) image rendering is done
by GTK itself. As an extra plus, this means we do not need librsvg anymore.

The icon base (top-left arrow) has been replaced with an SVG version.

the logic for creating the special placement cursor has been moved to `DiagramView`.

(we use Pillow for images on diagrams)

- [ ] Yes
- [x] No

### Other information

It works fine on macOS, but it may be good to test this on Linux and Windows too.


danyeaw:
Hey @amolenaar, I did some testing of this on Linux. It looks like for a non-Flatpak dev environment we need another dependency in order to prevent getting a libgles2.so not found error. I changed the Fedora and Ubuntu dev environment instructions to add this, I was only able to test this on Fedora though.

On Windows, it works, but the placement icon and preview are much lighter and more blurry - it might be just ok, but definitely looks worse.

amolenaar:
Thanks for testing. Maybe we should do the rendering with the fallback Cairo renderer. Let's see if that will work. 

It strikes me as odd that we need a separate GL library, while we already use the renderer from the underlying surface.

amolenaar:
Updated PR to use `Gsk.CairoRenderer` instead of the native one.

